PR #11, opened 2015-03-11
Bug363896 mixed tabs spaces in pluginxml

kthoms:
The patch will consequently replace space indentations by tabs. Most templates used 4 spaces, but some also 3 or 2. Most template code was already using tabs.

Further did some refactoring on the indentation.

Moved attribute 'point' to the same line as '<extension'. This was partially the case, but some extensions broke this to a new line.

Shortened usage of reference:
  <reference
     definitionId="«name».Editor.opened">
  </reference>

=>

  <reference definitionId="«name».Editor.opened"/>


oehme:
Please rebase this change on the current master =)

szarnekow:
If we apply this change, all languages have to be regenerated as part of the PR, too. Otherwise it's the next one's burden to figure out the relevant changes to generated code from the changes caused by this cosmetic modification.

spoenemann:
A lot of whitespace issues have already been fixed while implementing the new generator infrastructure. I improved some more whitespace in #906. From now on we should focus on the new generator and change the old one only for bugfixes.

PR #12, opened 2015-03-11
[bug 443574] Added interface IFileSystemAccessExtension4

kthoms:
This patch introduces interface IFileSystemAccessExtension4, which defines method exists().
Implemented new exists() method for FSA implementations.
Adding test code.

Change-Id: I4380ebdabe023d5abeb0219fc8f6b27c73e01d4d

### DIFF
--- a/plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/ParallelFileSystemAccess.xtend
+++ b/plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/ParallelFileSystemAccess.xtend
@@ -10,12 +10,13 @@ import org.eclipse.xtext.generator.IFileSystemAccessExtension2
 import org.eclipse.xtext.generator.IFileSystemAccessExtension3
 import org.eclipse.xtext.resource.IResourceDescription.Delta
 import org.eclipse.xtext.util.RuntimeIOException
+import org.eclipse.xtext.generator.IFileSystemAccessExtension4
 
 /**
  * @author Anton Kosyakov
  * @since 2.7
  */
-class ParallelFileSystemAccess implements IFileSystemAccess, IFileSystemAccessExtension, IFileSystemAccessExtension2, IFileSystemAccessExtension3 {
+class ParallelFileSystemAccess implements IFileSystemAccess, IFileSystemAccessExtension, IFileSystemAccessExtension2, IFileSystemAccessExtension3, IFileSystemAccessExtension4 {
 
 	Delta delta
 	
@@ -125,6 +126,19 @@ class ParallelFileSystemAccess implements IFileSystemAccess, IFileSystemAccessEx
 			return delegate.readTextFile(fileName, new NullProgressMonitor)
 		}
 		return (delegate as IFileSystemAccessExtension3).readTextFile(fileName)
+	}
+
+	/**
+	 * @since 2.9
+	 */
+	override isFile(String path, String outputConfigurationName) throws RuntimeIOException {
+		if (delegate instanceof EclipseResourceFileSystemAccess2) {
+			return delegate.isFile(path, outputConfigurationName, new NullProgressMonitor)
+		}
+		if (delegate instanceof IFileSystemAccessExtension4) {
+			return (delegate as IFileSystemAccessExtension4).isFile(path, outputConfigurationName)
+		}
+		return false
 	}
 
 }
--- a/plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/IFileSystemAccessExtension4.java
+++ b/plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/IFileSystemAccessExtension4.java
@@ -1,0 +1,28 @@
+/*******************************************************************************
+ * Copyright (c) 2015 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.generator;
+
+import org.eclipse.xtext.util.RuntimeIOException;
+
+/**
+ * This interface extends {@link IFileSystemAccess} with the capability to check files for existence.
+ *
+ * @author Karsten Thoms - Initial contribution and API
+ * @since 2.9
+ */
+public interface IFileSystemAccessExtension4 {
+
+	/**
+	 * Tests whether the file exists at the location denoted by the output configuration.
+	 * @param path using '/' as path separator
+	 * @param outputConfigurationName the name of the output configuration
+	 * @return <code>true</code> when the file at the given path exists and is a normal file. Will return <code>false</code> when
+	 * the path belongs to a directory.
+	 */
+	boolean isFile (String path, String outputConfigurationName) throws RuntimeIOException;
+}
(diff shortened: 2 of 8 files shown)


oehme:
Jenkins test

szarnekow:
Could you please update the @since tags to 2.9?

kthoms:
Updated @since tags

plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/JavaIoFileSystemAccess.java
```diff
@@ -235,4 +235,10 @@ public CharSequence readTextFile(String fileName, String outputCfgName) throws R
 			throw new RuntimeIOException(e);
 		}
 	}
+
+	@Override
+	public boolean exists(String fileName, String outputConfigurationName) throws RuntimeIOException {
+		File file = getFile(fileName, outputConfigurationName);
+		return file!=null && file.exists();
```
szarnekow: This logic seems to be different from the logic based on InputStreams, e.g. file.exists returns true for directories.

kthoms: In other words,
    && !file.isDirectory()
should be appended to the condition, right?

szarnekow: Yes, that should be correct and semantically equivalent to IFile.exists. Could you add a test for that case, too and add a remark to the JavaDoc?

kthoms: Will do when I have some time again.

intellij/org.eclipse.xtext.core.idea.tests/xtend-gen/org/eclipse/xtext/index/idea/completion/IndexTestLanguageCompletionContributor.java
```diff
@@ -0,0 +1,16 @@
+package org.eclipse.xtext.index.idea.completion;
```
szarnekow: This file should not be part of the PR

intellij/org.eclipse.xtext.idea/xtend-gen/org/eclipse/xtext/idea/extensions/IdeaProjectExtensions.java
```diff
@@ -1,14 +1,12 @@
 package org.eclipse.xtext.idea.extensions;
```
szarnekow: This file should not be part of the PR

plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/ParallelFileSystemAccess.xtend
```diff
@@ -10,17 +10,18 @@ import org.eclipse.xtext.generator.IFileSystemAccessExtension2
...
 
 /**
  * @author Anton Kosyakov
  * @since 2.7
  */
-class ParallelFileSystemAccess implements IFileSystemAccess, IFileSystemAccessExtension, IFileSystemAccessExtension2, IFileSystemAccessExtension3 {
+class ParallelFileSystemAccess implements IFileSystemAccess, IFileSystemAccessExtension, IFileSystemAccessExtension2, IFileSystemAccessExtension3, IFileSystemAccessExtension4 {
 
 	Delta delta
-
```
szarnekow: Please avoid reformatting entire files in the future as it makes reviews more time consuming than necessary.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/IFileSystemAccessExtension4.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * This interface extends {@link IFileSystemAccess} with the capability to check files for existence.
+ *
+ * @author Karsten Thoms - Initial contribution and API
+ * @since 2.9
+ */
+public interface IFileSystemAccessExtension4 {
+
+	/**
+	 * Tests whether the file exists at the location denoted by the output configuration.
```
szarnekow: Please add a note to the javadoc that directories will return false

tests/org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/builderState/EclipseResourceFileSystemAccess2Test.java
```diff
@@ -28,23 +29,28 @@
...
+		fsa = injector.getInstance(EclipseResourceFileSystemAccess2.class);
 		fsa.setProject(project);
 		fsa.setOutputPath("src-gen");
 		fsa.getOutputConfigurations().get(IFileSystemAccess.DEFAULT_OUTPUT).setCreateOutputDirectory(true);
 		fsa.setMonitor(new NullProgressMonitor());
+	}
 
+	@After
+	public void tearDown() throws Exception {
+		IResourcesSetupUtil.cleanWorkspace();
```
szarnekow: Minor: Some test classes share a common test project to speed up test execution (not in this particular test suite, though). Recently we tried to avoid a full clean of the workspace and rather remove the projects that a test class created.

svenefftinge:
Returning false for an existing folder is very counterintuitive. We should change the semantics such that it is the same as in 'java.io.File#exists()', i.e. 

     * @return  <code>true</code> if and only if the file or directory denoted
     *          by this abstract pathname exists; <code>false</code> otherwise

That said, it might be the right time to also add 'isDirectory'.

Please squash the commits to a single one, if they all belong together. That makes reviewing much easier.

szarnekow:
> Returning false for an existing folder is very counterintuitive. 

There is one impl of the interface that doesn't know about folders thus could not implement the contract. What's wrong with returning false for folders? IFileSystemAccess cannot deal with folders anyway as far as I can tell, e.g. there is no mkdir or similar.

svenefftinge:
> There is one impl of the interface that doesn't know about folders thus could not implement the contract. 
Which one is that?
> What's wrong with returning false for folders?
As I said, it is against my expectations as I'm familiar with File#exists.

szarnekow:
It's org.eclipse.xtext.generator.InMemoryFileSystemAccess

If we want to handle folders, too with the IFileSystemAccess API, we should revise it holistically, e.g. add something to access the members of a folder, mkdir, isDirectory, isFile etc. At least isDirectory would be necessary since exists() followed by read() would still throw an IOException if exists() returns true for folders.

svenefftinge:
Yes, isDirectory should be added as well.

szarnekow:
Alternatively we only add #isFile and don't add #exists or #isDirectory

svenefftinge:
Yes, that would be ok for me as well.

svenefftinge:
It really is 'isFileAndExists' which somehow tells me it should be two methods.

szarnekow:
After an offline discussion and a look at the semantics of java.io.File.isFile, we decided to add isFile rather than exists() to the service. Karsten, do you want to update your PR?

kthoms:
I agree that isFile does what the user wants to do, but it is not intuitive to use the method isFile if the user actually wants to prove the existence of a file. When a user searches the API for such a method he would have to dig the javadocs to find that isFile actually does what is intended, and in the user code this would need some comment that isFile is used to check for existence. I would like to avoid this.

kthoms:
Is it now as requested?

szarnekow:
Please remove the two files that don't belong to this PR.

szarnekow:
Jenkins test

PR #13, opened 2015-03-11
Bug398157 antlr token def provider encoding

kthoms:
[bug 398157] Use IEncodingProvider to create InputStream with language encoding


szarnekow:
I'm not convinced by this change. The token def provider reads a properties file that was written at compile time whereas the IEncodingProvider is supposed to provide the encoding for the DSL files. Instead of using the IEncodingProvider, we should make sure that the token def file is always written and read with UTF-8.

szarnekow:
This would unfortunately include a .settings config for the token def file to allow to open it with the correct encoding within the dev eclipse.

PR #14, opened 2015-03-11
[bug 336323] Add Xtext version to generated file header

kthoms:
Change-Id: I3bafd69653b3e1f47fa141c8204b9a13e5057c91


oehme:
Jenkins test

oehme:

szarnekow:
I'm not really convinced by this change since it will produce new output in our source gen folders by default. We would at least need to remove that version information from all the languages in the Xtext repo.

kthoms:
This could be done by configuring the file header for these languages to
NOT use the ${version} property. The requirement is that from the generated
sources it is derivable with which Xtext version they were produced, and
this by default, since most users would be just too lazy to change it. It
might not be necessary to do this in all files, or even in a separate file.

szarnekow:
I understand the req. but I also want to avoid the noise in our test sources.
The four options that I see:
1) update all our mwe2 configs
2) change the default to not include the version number
3) change the default to not include the version number but adjust the new project wizard
4) dont' implement this feature at all

kthoms:
Option 3) sounds reasonable. Should I change it?

szarnekow:
Yes, please update the pull request accordingly.

kthoms:
Pull request updated. The new patch will set the fileHeader property with use of the ${version} variable for projects created by the wizard.

plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/Generator.java
```diff
@@ -313,10 +318,41 @@ public String getFileHeader() {
...
+			}
 		}
 		return result;
 	}
 
+	private String getVersion() {
+		// Read the exact version from the Manifest of the plugin.
+		try {
+			URL url = new URL(Plugin.INSTANCE.getBaseURL() + "META-INF/MANIFEST.MF");
+			final Manifest manifest = new Manifest(url.openStream());
```
szarnekow: The stream should be closed after it was read

plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/Generator.java
```diff
@@ -313,10 +318,41 @@ public String getFileHeader() {
...
+			return manifest.getMainAttributes().getValue("Bundle-Version");
+		}
+		catch (Exception e) {
+			return null;
+		}
+	}
+
+	// only needed to determine the Manifest file and its version of this plugin in standalone mode
+	private static class Plugin extends EMFPlugin {
```
szarnekow: Would it be reasonable to extends DelegatingResourceLocator directly? I don't see any code that uses anything of the EMFPlugin

szarnekow:
Jenkins test

PR #15, opened 2015-03-11
[xtend] Bug 365910: Validation for primitive dispatch method arguments

spoenemann:
- Added validation for different primitive types as dispatch method arguments
- Fixed type computation so a non-primitive type is given to the dispatcher method


spoenemann:
The change has already been reviewed on Gerrit:
https://git.eclipse.org/r/#/c/43312/
Waiting for auto-build to verify...

oehme:
Jenkins test

PR #19, opened 2015-03-11
[xtend] Bug 460454: enhanced inheritance cycle detection such that containing types are considered

spoenemann:
Transferred from Gerrit:
https://git.eclipse.org/r/#/c/43500/
Waiting for auto-build to verify...


oehme:
Jenkins test

PR #25, opened 2015-03-13
[bug 378057] type filter

kthoms:
This patch will add the methods isFilteredByPreference and isFilteredByDefault to TypeMatchFilters. isFilteredByPreference considers the JDT type filter preference.

### DIFF
--- a/plugins/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/contentassist/XtendProposalProvider.java
+++ b/plugins/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/contentassist/XtendProposalProvider.java
@@ -133,33 +133,42 @@ public class XtendProposalProvider extends AbstractXtendProposalProvider {
 		}
 	}
 	
-	@Override
-	protected ITypesProposalProvider.Filter createVisibilityFilter(ContentAssistContext context, int searchFor) {
-		XtendFile file = (XtendFile) context.getRootModel();
-		final char[] contextPackageName = Strings.emptyIfNull(file.getPackage()).toCharArray(); 
-		return new TypeMatchFilters.All(searchFor) {
-			@Override
-			public boolean accept(int modifiers, char[] packageName, char[] simpleTypeName,
-					char[][] enclosingTypeNames, String path) {
-				if (super.accept(modifiers, packageName, simpleTypeName, enclosingTypeNames, path)) {
-					if ("org.eclipse.xtend.lib".equals(String.valueOf(packageName))) {
-						if ("Property".equals(String.valueOf(simpleTypeName))||"Data".equals(String.valueOf(simpleTypeName))) {
-							return false;
-						}
-					}
-					if (Flags.isPublic(modifiers)) {
-						return true;
-					}
-					if (Flags.isPrivate(modifiers)) {
+	private static class VisibilityFilter extends TypeMatchFilters.AbstractFilter {
+		private ContentAssistContext context;
+		public VisibilityFilter(ContentAssistContext context, int searchFor) {
+			super(searchFor);
+			this.context = context;
+		}
+		@Override
+		public boolean accept(int modifiers, char[] packageName, char[] simpleTypeName, char[][] enclosingTypeNames,
+				String path) {
+			if (TypeMatchFilters.isNotInternal(getSearchFor()).accept(modifiers, packageName, simpleTypeName, enclosingTypeNames, path)
+				&& TypeMatchFilters.isAcceptableByPreference().accept(modifiers, packageName, simpleTypeName, enclosingTypeNames, path)) {
+				XtendFile file = (XtendFile) context.getRootModel();
+				final char[] contextPackageName = Strings.emptyIfNull(file.getPackage()).toCharArray();
+
+				if ("org.eclipse.xtend.lib".equals(String.valueOf(packageName))) {
+					if ("Property".equals(String.valueOf(simpleTypeName))||"Data".equals(String.valueOf(simpleTypeName))) {
 						return false;
 					}
-					if (Arrays.equals(contextPackageName, packageName)) {
-						return true;
-					}
 				}
-				return false;
-			}
-		};
+				if (Flags.isPublic(modifiers)) {
+					return true;
+				}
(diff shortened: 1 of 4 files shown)


eclipsewebmaster:
:white_check_mark: 
#### Eclipse validation results:
All committers passed Eclipse CLA and Signed-off-by validation.

oehme:

oehme:
Jenkins test

szarnekow:
Should we use `isFilteredByDefault` instead of isInternalClass at the other locations, too? I think that would be reasonable.

kthoms:
I think this is too much. isInternalClass is used in CanInstantiate and IsPublic, they will return false for internal classes. The JDT type filter is only used for type proposals, this does not change that these types could be referred to. For example, you can use java.awt.List of course even if it is a filtered type. It won't be proposed as a type only.

An alternative is just to leave the isFilteredByDefault method away and use isFilteredByPreference and isInternalClass together where isFilteredByDefault is used now.

svenefftinge:
It's somewhat strange if something called 'all' has fewer results than the other filters. Isn't all this stuff intended to be used for content assist? If not, I suggest we introduce a new filter with a name that tells the user that the JDT preference filters apply.

kthoms:
OK, I'll come up with a new proposal

2015-03-18 8:10 GMT+01:00 Sven Efftinge <notifications@github.com>:

> It's somewhat strange if something called 'all' has fewer results than the
> other filters. Isn't all this stuff intended to be used for content assist?
> If not, I suggest we introduce a new filter with a name that tells the user
> that the JDT preference filters apply.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/eclipse/xtext/pull/25#issuecomment-82787391>.
>



-- 
Karsten Thoms
Softwarearchitekt

Telefon: +49 (0) 231 / 98 60-269
Telefax: +49 (0) 231 / 98 60-211
Mobil: +49 (0) 170 / 6369841

http://www.itemis.de

Karsten.Thoms@itemis.de

https://www.xing.com/profile/Karsten_Thoms
http://twitter.com/kthoms
http://twitter.com/itemis
http://kthoms.wordpress.com/

itemis AG
Am Brambusch 15-24
44536 Lünen

Rechtlicher Hinweis:

Amtsgericht Dortmund, HRB 20621

Vorstand: Jens Wagener (Vors.), Wolfgang Neuhaus, Dr. Georg Pietrek, Jens
Trompeter, Sebastian Neus
Aufsichtsrat: Dr. Burkhard Igel (Vors.), Stephan Grollmann, Michael Neuhaus

kthoms:
Updated proposal: I removed the method "isFilteredByDefault" and used isFilteredByPreference where appropriate.

szarnekow:
I think Svens proposal was to have something like TypeMatchFilters.filteredByUserPreference that can be used along with TypeMatchFilters.and(..)

kthoms:
Now created a filter "AllVisible" that extends All.
Please advice more specific when this does not meet your expectations.

svenefftinge:
Thanks, the name "AllVisible" is still kind of irritating to me. It doesn't sound different to "All". What about the name Sebastian suggested? Could you squash the commits to one?

kthoms:
I implemented that and it felt quite ugly in use. If this filter should be used together with "and", it usually must be negated.
For me it felt best when implemented in the All filter, since "All" means in the context "All types available for content assist". Actually "All" does already restrict types to non-inner types, thus it is not really an "All types" filter.
Could you please take over and implement this like you want? I think the intention and the basic implementation is already clear.

kthoms:
In an offline discussion with Sebastian we decided the following:
- the additional filter will be called "AcceptableByPreference". This avoids the necessarity to use a negation("... and is not filtered by preference")
- the All filter gets an alternative "IsNonInternal". This addresses the issue that an All filter combined with an and() is difficult to understand.

As a result, locations that used the All filter so far will use "and(isNonInternal(),isAcceptableByPreference()).

Maybe we should deprecate All then?

plugins/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/contentassist/XtendProposalProvider.java
```diff
@@ -133,33 +133,42 @@ public void completeTypeReferenceNoTypeArgs_Type(EObject model, Assignment assig
...
+			super(searchFor);
+			this.context = context;
+		}
+		@Override
+		public boolean accept(int modifiers, char[] packageName, char[] simpleTypeName, char[][] enclosingTypeNames,
+				String path) {
+			if (TypeMatchFilters.isNonInternal(getSearchFor()).accept(modifiers, packageName, simpleTypeName, enclosingTypeNames, path)
+				&& TypeMatchFilters.isAcceptableByPreference().accept(modifiers, packageName, simpleTypeName, enclosingTypeNames, path)) {
+				XtendFile file = (XtendFile) context.getRootModel();
+				final char[] contextPackageName = Strings.emptyIfNull(file.getPackage()).toCharArray();
```
szarnekow: Please make 'contextPackageName' a field in the inner class as we'd need to convert the package name to an array for all candidates otherwise.

plugins/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/TypeMatchFilters.java
```diff
@@ -20,17 +21,37 @@ private TypeMatchFilters() {
...
+	 * @since 2.9
+	 */
+	public static ITypesProposalProvider.Filter isNonInternal() {
+		return new IsNonInternal();
+	}
+	
+	/**
+	 * @since 2.9
+	 */
+	public static ITypesProposalProvider.Filter isNonInternal(int searchFor) {
```
szarnekow: Since there was some confusion about `all`, we should have javadoc on `isNonInternal`. Can we call the method `isNotInternal`?

kthoms: Will update the PR now

PR #27, opened 2015-03-14
[bug 406066] Improve usage of simple names for java.lang classes

kthoms:
When the import manager is created with a type that collides with java.lang, the flag thisCollidesWithJavaLang is set. But in method allowsSimpleName it is not considered that the actual type name that should be shortened is different from the one with that the ImportManager was created.

### DIFF
--- a/tests/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug406066Test.xtend
+++ b/tests/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug406066Test.xtend
@@ -1,0 +1,34 @@
+/*******************************************************************************
+ * Copyright (c) 2015 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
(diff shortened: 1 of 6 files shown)


oehme:

tests/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug406066Test.xtend
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @author Karsten Thoms - Initial contribution and API
+ */
+class CompilerBug406066Test extends AbstractXtendCompilerTest {
+	
+	@Test
+	def testBug406066_01() {
+		assertCompilesTo('''
+			class Error {
+				def static err(Integer i) {
```
szarnekow: Could you add a usage of java.lang.Error in this method, too?

kthoms: Added 2 more test cases.

szarnekow:
Jenkins test

szarnekow:
Jenkins test

szarnekow:
@kthoms unfortunately the patch breaks existing tests (which may have a bogus expectation, though). Could you double check these?

kthoms:
Fixed test expectation

szarnekow:
Thank you for the patch, Karsten.

PR #28, opened 2015-03-16
[serializer] keywords don't have precedence over rule calls

meysholdt:
Because rule calls can refer to DataType rules which again may
contain keywords.

Change-Id: If128a1fa89330df970fe5a076ec7db0dc8e01e64


meysholdt:
Jenkins test

meysholdt:
Jenkins rebase

szarnekow:
Are you going to add tests for that, too?

meysholdt:
I came across this when serializing something like "extension FooClass foo" with Xtend.xtext#Member. Running the Xtend formatter tests with the serializer enabled tests it.

PR #34, opened 2015-03-16
[461923]: Fixed type inference issue with deferred type arguments

szarnekow:
In cases like ImmutableList.builder.addAll(arg).build with
arg: Iterable<? extends Something>, the type argument for builder
was infered to ? extends Something which produced an expectation
of Iterable<? extends ? extends Something> for the argument to addAll.

see https://bugs.eclipse.org/bugs/show_bug.cgi?id=461923

Signed-off-by: Sebastian Zarnekow <Sebastian.Zarnekow@itemis.de>


szarnekow:
Jenkins test

PR #36, opened 2015-03-17
[xtend] Bug 462313: Problem with synthetic _this variable for anonymous classes

spoenemann:
protected members may not be visible if accessed through a synthetic _this variable.


szarnekow:
Could please add a test that combines the scenario of this bug with the scenario that required to _this field? Do we see a proper validation for the seemingly not compilable code?

spoenemann:
Currently the `_this` is generated even if it's not absolutely necessary. The proposed change suppresses it in cases where it would reduce the visibility (protected members). Cases where the `_this` is really necessary *and* we have a reference to a protected member from another package would be compiled incorrectly:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=458429
A respective validation is still missing, but that could be done for bug 458429 (hence not for v2.8.1)?

szarnekow:
Agreed.

szarnekow:
Jenkins test

PR #41, opened 2015-03-18
[bug 461761] fixes for storage loaded resources: 1) don't collect imported constants 2) don't use concreteSyntaxName for operators

svenefftinge:


plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/ConstantExpressionsInterpreter.xtend
```diff
@@ -94,6 +94,10 @@ class ConstantExpressionsInterpreter extends AbstractConstantExpressionsInterpre
 	 * That essentially includes static imports and the fields declared in the containing types
 	 */
 	protected def Map<String, JvmIdentifiableElement> findVisibleFeatures(XExpression expression) {
+		// don't build a scope for a storage loaded resource, since everything is computed and we will resolve the proxies
+		switch res : expression.eResource {
```
szarnekow: This will break the interpreter for statically imported constant values, e.g. they can no longer be used as annotation values.

svenefftinge: Why is that? This only affects storage loaded resources, in which case we use the linked feature calls.

szarnekow: I only see one caller where this should be relevant, but I agree, it's ok as it is.

PR #44, opened 2015-03-18
[462260]: Fixed compilation problem with collection literals

szarnekow:
If the expected type was a type parameter with a constraint
of the form T extends Map<..> or T extends Iterable<..>, XSetLiterals
did not correctly coerce to a map or set respectively. Also
unresolved expectations caused trouble, e.g. Map&lt;E, List&lt;E>>
could not be fulfilled by #{ a->#[] }.

see https://bugs.eclipse.org/bugs/show_bug.cgi?id=462260


szarnekow:
jenkins test

szarnekow:
Jenkins test

PR #45, opened 2015-03-19
[462246] Don't set enum and class references as compile time constants in JvmField (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=462246)

svenefftinge:


szarnekow:
+1
It's way better when the serialized JVM model looks the same as the one produced from the class file.

PR #47, opened 2015-03-19
[serializer] excluded tests from old GenericSemanticSequencer

meysholdt:
...because BacktrackingSemanticSequencer can handle them

Change-Id: I83ac829a43a5623dbb38e8086ab462c89e7e0a2d


szarnekow:
It would be interesting to know why these tests suddenly started to fail.

meysholdt:
I'm explaining in 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=462548
why there is an ambiguity during serialization that wasn't there before. 
The BacktrackingSemanticSequencer deals with this ambiguity by choosing the version it finds in the node model. However, the old GenericSemanticSequencer can't handle the ambiguity.

meysholdt:
merged to unblock fellow committers

PR #49, opened 2015-03-19
[461761] use constant values in constant expression interpreter

svenefftinge:


plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/ConstantExpressionsInterpreter.xtend
```diff
@@ -365,9 +365,11 @@ class ConstantExpressionsInterpreter extends AbstractConstantExpressionsInterpre
 	}
 	
 	protected def evaluateField(XAbstractFeatureCall call, JvmField field, Context context) {
+		if (field.isSetConstant) {
```
szarnekow: isSetConstant only indicates that the boolean flag 'isConstant' was set. You'll have to check field.isConstant, too.

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/ConstantExpressionsInterpreter.xtend
```diff
@@ -365,9 +365,11 @@ class ConstantExpressionsInterpreter extends AbstractConstantExpressionsInterpre
...
+			return field.constantValue
+		}
 		if (field.eResource instanceof TypeResource) {
-			val clazz = getJavaType(field.declaringType, context.classFinder)
-			return clazz.getField(field.simpleName)?.get(null)
+			throw new ConstantExpressionEvaluationException("Field "+field.declaringType.simpleName+"."+field.simpleName+" is not a constant")
```
szarnekow: field.isSetConstant is also true, if field.isConstant was explicitly set to false. The ConstantExpressionEvaluationException should be raised in these cases, too

PR #50, opened 2015-03-19
[461018] don't log JavaModelExceptions in the build as errors

JanKoehnlein:
Change-Id: Ib55253074550456a6d526ee9f0a490d7dcf15f2e


JanKoehnlein:
Analysis: A workspace change triggered the initialization of the Storage2UriMapperJavaImpl's cache which found that the jar of a PFR does no longer exist.

We should not log any JavaModelExceptions as errors. It is most likely that JDT has its way of reporting them to the user when it takes its build turn. Switching logging level to debug.

JanKoehnlein:
Unrelated tests are failing. Moritz has been notified. Merging anyway.

PR #52, opened 2015-03-19
[461019] detect missing jar files before trying to index their content

JanKoehnlein:
Change-Id: I8362cf065ca5c7af2cd0d5749729e6ceed6adc22


JanKoehnlein:
The stack trace is actually logged by JDT, so we have to try to avoid it upfront if we really want to get rid of this.

JanKoehnlein:
Unrelated tests are failing. Moritz fixed them. Merging.

PR #54, opened 2015-03-19
[462121] don't log broken JARs during indexing as errors

JanKoehnlein:
Change-Id: I

A PFR with a broken jar. This is likely detected and reported by JDT as well, so we should not log this as an error.


JanKoehnlein:
Unrelated test failure only. Moritz is informed. Meging.

PR #61, opened 2015-03-19
Bug 462554 - [xbase] add support for context assist of referenced

Flamenco:
enumerations in explicit operation calls

Change-Id: Ifdca705bedb9d8ba700dd69e6c3a8924a06399e4
Signed-off-by: Steven Spungin <steven@spungin.tv>

I'm sure there are better ways to do this, but it works form me :)

This would have been much simpler if there was a proposal stub that gave the call expression argument index, along with the target reference type;  The XText grammar seems to not be explicit enough to handle it automatically,  but this patch might pave the way for that.

### DIFF
--- a/plugins/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseProposalProvider.java
+++ b/plugins/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseProposalProvider.java
@@ -24,10 +24,13 @@ import org.eclipse.xtext.GrammarUtil;
 import org.eclipse.xtext.Group;
 import org.eclipse.xtext.Keyword;
 import org.eclipse.xtext.RuleCall;
+import org.eclipse.xtext.common.types.JvmEnumerationLiteral;
+import org.eclipse.xtext.common.types.JvmEnumerationType;
 import org.eclipse.xtext.common.types.JvmExecutable;
 import org.eclipse.xtext.common.types.JvmFeature;
 import org.eclipse.xtext.common.types.JvmFormalParameter;
 import org.eclipse.xtext.common.types.JvmIdentifiableElement;
+import org.eclipse.xtext.common.types.JvmOperation;
 import org.eclipse.xtext.common.types.JvmTypeReference;
 import org.eclipse.xtext.common.types.TypesPackage;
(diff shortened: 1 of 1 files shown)


oehme:
[Jenkins Bot] Can a team member please verify that this is good for testing?

dhuebner:
Jenkins test

szarnekow:
Thanks for the proposed patch, Steven. At a first glance, it lacks some important things, e.g it doesn't consider extension methods and their shifted arguments lists, assignment expression or the fact that the model is incomplete quite often during CA.
We should discuss if we want to have that feature and see how it can be implemented best in Xbase.

Flamenco:
@szarnekow Thanks for quick response.  Seeing that Xbase and Xtend are really into inferring things, it seems pretty reasonable that if one refers to a type that maps to an enumeration, the content assist would suggest the options.  I think this is the default behavior in Eclipse vanilla Java.

I would suggest to simply not add suggestions in the cases you mentioned, and then implement them on a case-by-case basis.  It seems that most users of Xtend and Xbase would be very happy with at least basic support for this.

szarnekow:
We are considering to allow unqualified enum literals where an enum is expected. This will render the content assist customizing obsolete thus I'd like to close this pull request, ok?

Flamenco:
> We are considering to allow unqualified enum literals where an enum is expected

Unqualified enums are already allowed in switch statements.  For the other places, I don't know if I would prefer them to be qualified or not. IMHO this is a completely different issue concerning syntax;  I don't see why it has anything to do with content assist.  Are you suggesting that by allowing the unqualified name the context assist will then just work?  If so, will it show the enums when no prefix is present? 

Consider that the current proposed patch will:
- suggest names based on the unqualified name
- search for the enum values without needing to use capital letters (and enums by convention are). 
- suggest the enums first in the list.  
- show all the enums when no prefix entered

szarnekow:
> Are you suggesting that by allowing the unqualified name the context assist will then just work? If so, will it show the enums when no prefix is present? 

Yes and yes.

> search for the enum values without needing to use capital letters (and enums by convention are). 

Content assist is generally case insensitive and also camel case aware. No need to adopt smth there.

> suggest the enums first in the list.

That's generally speaking an orthogonal problem, where proposals with a matching type should be ranked higher.

> show all the enums when no prefix entered

That will work transparently because all enums match the empty prefix.

Flamenco:
That works for me :)  I am looking forward to seeing this implemented!

PR #70, opened 2015-03-23
[461764]: Fixed wrongly propagated exception

szarnekow:
The IResourceValidator is not supposed to throw a platform
specific cancellation exception but should wrap it in an
OperationCanceledError instead.

see https://bugs.eclipse.org/bugs/show_bug.cgi?id=461764


svenefftinge:
+1

PR #75, opened 2015-03-24
[460191]: [xtend][linking] Expected type had too high prio when linking

szarnekow:
The expected type contributed to unresolved type parameters as if
it was directly inferred which caused trouble, e.g. with the overload
of IterableExtensions.filter

see https://bugs.eclipse.org/bugs/show_bug.cgi?id=460191


szarnekow:
This is expected to fail right now.

PR #80, opened 2015-03-25
 [462914]: Fixed exception during compilation

szarnekow:
Broken create extensions are now properly guarded
by the error-safe extension mechanism

see https://bugs.eclipse.org/bugs/show_bug.cgi?id=462914


szarnekow:
Jenkins test

PR #88, opened 2015-03-26
[serializer] use node model even for rules without assignment

meysholdt:
This fixes a bug that caused the SyntacticSequencer to not consult the
node model if a serialized rule doesn't have any assignments but only an
action.


meysholdt:
Jenkins test

PR #95, opened 2015-03-26
[grammar][contentAssist] Fixed two cosmetic issues

szarnekow:
- Prefix matching didn't work properly for the super grammar reference.
- The comma after the first super grammar should not be proposed since only one super grammar is supported.


szarnekow:
Jenkins xtext-ide-tests-maintenance-latest

PR #106, opened 2015-03-30
Added integration test for xtend.maven.archetype

vorburger:
Hi guys, perhaps this is useful? It adds an integration test to the xtend.maven.archetype, as in with this the build actually "instantiates" the archetype template, and runs a 'mvn package' on it, to make sure it cleanly builds.

PS: I built this because I suspect the archetype was broken, and meant to contribute a fix, but then later found it was another problem on my end, as I had already built this test, I though I'd contribute it anyway, as you may find this useful for future non-regression?

PPS: This doesn't actually pass on current master head for me, but the failure I'm getting is probably because you're working on something new on 2.9 and probably already know that it seems to be currently broken?

### DIFF
--- a/maven/org.eclipse.xtend.maven.archetype/pom.xml
+++ b/maven/org.eclipse.xtend.maven.archetype/pom.xml
@@ -1,16 +1,23 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
 	<parent>
 		<groupId>org.eclipse.xtext</groupId>
 		<artifactId>org.eclipse.xtext.parent</artifactId>
 		<version>2.9.0-SNAPSHOT</version>
 		<relativePath>../org.eclipse.xtext.parent</relativePath>
 	</parent>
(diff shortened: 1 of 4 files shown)


dhuebner:
Thanks Michael!
Unfortunately build fails using m2e with embedded maven:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-archetype-plugin:2.3:integration-test (default-integration-test) on project xtend-archetype: Execution default-integration-test of goal org.apache.maven.plugins:maven-archetype-plugin:2.3:integration-test failed: ${maven.home} is not specified as a directory: '/Users/dhuebner/git/org.eclipse.xtext-master/maven/org.eclipse.xtext.parent/EMBEDDED'. -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :xtend-archetype

I will try to find a solution for building with m2e.

vorburger:
Actually I also had that, and know how to get around this (if you don't want to open a CLI): In the M2E Launch Configuration, at the bottom of the Main tab where it says Maven Runtime you can use Configure.. and easily Add.. a standard external standalone Maven home directory, then it works like from a CLI even if launched from within Eclipse via M2E.

PR #115, opened 2015-04-02
derived state error handling

svenefftinge:


plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DerivedStateAwareResource.java
```diff
@@ -233,10 +238,19 @@ public void installDerivedState(boolean preIndexingPhase) {
 						throw Throwables.propagate(e);
 					}
 				}
-				fullyInitialized = true;
+			} catch (RuntimeException e) {
```
szarnekow: We shouldn't log OperationCancelExceptions

PR #140, opened 2015-04-09
[formatter/xbase] format 'instanceof' expressions

meysholdt:


szarnekow:
Trivial fixes like this should probably be done in maintenance such that they will be available in a potential 2.8.3. Could you please cherry-pick that after 2.8.2 was build?

meysholdt:
done, see #150.

PR #141, opened 2015-04-09
Make @ToString faster

oehme:
Signed-off-by: Stefan Oehme <stefan.oehme@itemis.de>


szarnekow:
Given that we have tests that fail with the changed default, we should rather use a flag on the annotation to use a plain StringBuilder or similar. Regarding the performance expectation, I'll try to do some experiments with a handwritten impl.

oehme:
The first failing test is a compiler test that expected the old output.
The second failure is a @Data class that has a mutable reference in it. I don't think that is a good representation of what @Data is supposed to be. So I'm still in favor of making it the default.

oehme:
I reran the JMH benchmark and this seems to have almost no effect.

PR #153, opened 2015-04-14
[formatter] improve indent api

meysholdt:
The API useed to have the methods increaseIndentation() and
decreaseIndentation(). While this is quite flexible, it is also quite
likely that the two methods are not called symmetrically.

This can happen, if
- the developer forgets to call one of the methods
- due to a bug in the code's logic that calls one of the methods
- due to an exception that's being thrown after inc() has been called
  and before dec() would be called. Exceptions are likely to be thrown
  because formatter may be executed on syntactically broken documents.
  
If inc() and dec() aren't called symmetrically, the indenteation for the
remaining part of the document will be wrong which makes any formatter
look broken. Therefore, we can consider it to be the better solution to
apply one level of intention properly or not at all. 

The new API makes use of methods from the IFormattableDocument that are
resposnsible for two HiddenRegions at the same time. Namely, the methods
are:
- surround(ISemanticRegion)
- surround(EObject)
- interior(ISemanticRegion, ISemanticRegion)
- set(IHiddenRegion, IHiddenRegion)

When calling one of these methods, IHiddenRegionFormatting#indent() can
be used to increase the level of indentation form the first HiddenRegion
to the second one.


meysholdt:
Jenkins test

PR #158, opened 2015-04-16
[380232] getColumn() should return 0 for "for no information available"

kthoms:


szarnekow:
Jenkins test

PR #163, opened 2015-04-17
[xtext] Added two methods to EmbeddedEditorModelAccess

spoenemann:
- Update the model with no prefix and suffix
- Update the model with a given EObject and use the serializer to get the text

The latter is very useful for live synchronization with another EMF-based editor, e.g. a GMF editor.

### DIFF
--- a/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/embedded/EmbeddedEditorModelAccess.java
+++ b/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/embedded/EmbeddedEditorModelAccess.java
@@ -9,6 +9,8 @@ package org.eclipse.xtext.ui.editor.embedded;
 
 import java.util.Collections;
 
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.IDocumentExtension4;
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/embedded/EmbeddedEditorModelAccess.java
```diff
@@ -92,6 +108,41 @@ public void updateModel(String prefix, String editablePart, String suffix) {
 		this.viewer.getUndoManager().connect(this.viewer);
 		this.viewer.setRedraw(true);
 	}
+	
+	public void updateModel(EObject model, final EObject editablePart) {
```
meysholdt: instead of passing in a editablePart as EObject, it may make sense to pass in a more generic selector instead, i.e. 

Function<XtextResource, ITextRegion> editablePartSelector

Otherwise use cases such as 
- including/excluding leading comments
- displaying the region spanning two or more EObjects
are not possible

The method updateModel(EObject model, final EObject editablePart) could be a convenience method delegating to the more general method.

plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/embedded/EmbeddedEditorModelAccess.java
```diff
@@ -92,6 +110,52 @@ public void updateModel(String prefix, String editablePart, String suffix) {
 		this.viewer.getUndoManager().connect(this.viewer);
 		this.viewer.setRedraw(true);
 	}
+	
+	public void updateModel(final EObject model, final EObject editablePart) {
```
meysholdt: maybe just make this updateModel(final String model, final URI editablePartURIFragment)

I'm not quite sure what the benefit is of invoking the serializer inside this methods, because it's just one more call for the caller:

editor.updateModel(serializer.serialize(eObject), uri)

Passing EObjects into this methods may be misleading because the editor doesn't actually reuse them, but creates it's own ResourceSet with a freshly parsed model.

szarnekow: > Passing EObjects into this methods may be misleading because the editor doesn't actually reuse them, but creates it's own ResourceSet with a freshly parsed model.

Very good point. Having EObject in the arg list will likely cause confusion.

meysholdt:
awesome! not having EmbeddedEditorModelAccess call the serializer also ensures users can pass custom parameters (SaveOptions) to the serializer.

spoenemann:
The code using the new methods is here:
https://github.com/spoenemann/xtext-gef/blob/master/org.xtext.example.statemachine.gmf/src/org/xtext/example/statemachine/gmf/ui/TextPropertiesViewPart.xtend
(work in progress)

PR #172, opened 2015-04-21
Fix automatic indentation for RichStrings for some situations

holgerschill:
Test cases for that :
org.eclipse.xtend.ide.tests.autoedit.AutoEditTest.testBug434717_04()
org.eclipse.xtend.ide.tests.autoedit.AutoEditTest.testBug434717_05()

### DIFF
--- a/plugins/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/autoedit/RichStringPartionIndentationStrategy.java
+++ b/plugins/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/autoedit/RichStringPartionIndentationStrategy.java
@@ -13,9 +13,13 @@ import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.DefaultIndentLineAutoEditStrategy;
 import org.eclipse.jface.text.DocumentCommand;
 import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.IDocumentExtension4;
 import org.eclipse.jface.text.IRegion;
 
 /**
+ * This strategy handles a paste of multiline text into a RichtString.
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/autoedit/RichStringPartionIndentationStrategy.java
```diff
@@ -33,14 +36,12 @@ public void customizeDocumentCommand(IDocument d, DocumentCommand c) {
 					regex+= Pattern.quote(s);
 				}
 				regex += "]";
-				String[] lines = pastedText.split(regex);
+				String[] lines = c.text.split(regex);
 				String convertedText = lines[0];
 				for(int i = 1; i < lines.length ; i++){
-					String oneLine = lines[i];
-					convertedText += d.getLineDelimiter(line) +  lineIndentation + oneLine;
+					convertedText += ((IDocumentExtension4) d).getDefaultLineDelimiter() +  lineIndentation + lines[i];
```
szarnekow: We should better use a stringbuilder here instead of applying text in a loop
Also we should read the lineDelimiter only once

plugins/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/autoedit/RichStringPartionIndentationStrategy.java
```diff
@@ -33,14 +36,12 @@ public void customizeDocumentCommand(IDocument d, DocumentCommand c) {
 					regex+= Pattern.quote(s);
```
szarnekow: s may have more than one char, so the pattern should better use (a|b)

PR #181, opened 2015-04-22
[464591] [editor] Coloring not refreshed after a preference change

dhuebner:
Changed to instance scope to get events when syntax coloring changes.

Signed-off-by: Dennis Huebner <dennis.huebner@itemis.de>


svenefftinge:
jenkins test

PR #184, opened 2015-04-22
Reduce log level of JDT startup message from warning to info

heeckhau:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=465225


svenefftinge:
I think this should be closed and replaced by another pull request against maintenance.

PR #188, opened 2015-04-23
[formatter] made mechanism for finding text regions more powerful

meysholdt:


plugins/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/regionaccess/internal/AbstractSemanticRegionsFinder.java
```diff
@@ -0,0 +1,414 @@
...
+		Set<Predicate<ISemanticRegion>> predicates = Sets.newHashSet();
+		for (int i = 0; i < features.length; i++) {
+			EStructuralFeature feat = features[i];
+			assertNoContainment(feat);
+			predicates.add(new FeaturePredicate(feat));
+		}
+		return findAll(Predicates.or(predicates));
+	}
+
+	protected ImmutableList<ISemanticRegion> findAll(Predicate<ISemanticRegion> predicate) {
```
szarnekow: findAll will never return more than the first element, is that intentional?

plugins/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/regionaccess/ITextRegionExtensions.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.xtext.Action;
+import org.eclipse.xtext.ParserRule;
+import org.eclipse.xtext.RuleCall;
+
+/**
+ * @author Moritz Eysholdt - Initial contribution and API
+ */
+public interface ITextRegionExtensions {
```
szarnekow: Why was this moved to a new interface? ITextRegionAccess now has an accessor to obtain the extensions but in fact the methods here would be easier to discover if they were still on the *access interface - after all, they provide access to regions by means of finders.

meysholdt: Discussed offline to keep it this way; Methods from the *Extensions interface are intended to be on Xtend's extension scope while most methods from *Access should be on the extensions scope and some not even on the implicit scope.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/regionaccess/internal/AbstractSemanticRegionsFinder.java
meysholdt: refactored this in 1d942e3 to be an abstract method. Now subclasses get to decide on an implementation explicitly.

meysholdt:
Jenkins test

PR #199, opened 2015-04-27
RE in MarkerUtilities.handleCoreException see https://bugs.eclipse.org/b...

holgerschill:
...ugs/show_bug.cgi?id=465490

### DIFF
--- a/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/IssueUtil.java
+++ b/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/IssueUtil.java
@@ -7,7 +7,10 @@
  *******************************************************************************/
 package org.eclipse.xtext.ui.util;
 
+import java.util.Map;
+
 import org.eclipse.core.resources.IMarker;
+import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/IssueUtil.java
```diff
@@ -24,28 +27,56 @@
...
+			String markerType = marker.getType();
+			Object message = attributes.get(IMarker.MESSAGE);
+			issue.setMessage(message  instanceof String ? (String) message : null);
+			Object lineNumber = attributes.get(IMarker.LINE_NUMBER);
+			issue.setLineNumber(lineNumber instanceof Integer ? (Integer) lineNumber -1 : null);
+			Object offset = attributes.get(IMarker.CHAR_START);
+			Object endOffset = attributes.get(IMarker.CHAR_END);
+			if(offset instanceof Integer && endOffset instanceof Integer) {
+				issue.setOffset((Integer) offset);
+				issue.setLength((Integer) endOffset - (Integer) offset);
```
szarnekow: This will cause an NPE if offset is null

PR #201, opened 2015-04-27
[465080] Fixed classpath URI resolution

szarnekow:
Resources from the project root may also be considered
to be on the classpath thus the JDT based resolver
finds those, too.
This is a shortcut since we’d need to introspect the
build.properties, too. OTOH different build technologies
may chose different description formats so the shortcut
seems to be a valid assumption.


szarnekow:

PR #207, opened 2015-04-28
[web] handle operation canceled in async background process, use cancel-manager to check whether an exception is an operation-canceled-exception.

svenefftinge:


spoenemann:
Reviewed, +1

PR #216, opened 2015-05-04
Fix possible NPE found by FindBugs

holgerschill:
Set -1 instead of null as a default value for offset and length of an issue

### DIFF
--- a/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/IssueUtil.java
+++ b/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/IssueUtil.java
@@ -56,8 +56,8 @@ public class IssueUtil {
 				issue.setOffset((Integer) offset);
 				issue.setLength((Integer) endOffset - (Integer) offset); 
 			} else {
-				issue.setOffset(null);
-				issue.setLength(null);
+				issue.setOffset(-1);
+				issue.setLength(-1);
 			}
 			Object code = attributes.get(Issue.CODE_KEY);
 			issue.setCode(code instanceof String ? (String) code:null);


plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/IssueUtil.java
```diff
@@ -56,8 +56,8 @@ public Issue createIssue(IMarker marker) {
 				issue.setOffset((Integer) offset);
 				issue.setLength((Integer) endOffset - (Integer) offset); 
 			} else {
-				issue.setOffset(null);
-				issue.setLength(null);
+				issue.setOffset(-1);
```
szarnekow: -1 will probably cause other follow-up errors. Which problem did you see with null for Integer?

holgerschill:
https://xtext-builds.itemis.de/jenkins/view/All/job/xtext.code.analysis/3405/findbugsResult/new/package.1857478882/type.-1863013579/source.47/#50

holgerschill:
This has been used before so I just wanted to make sure that the code behaves like before.

public static int getCharEnd(IMarker marker) {
		return getIntAttribute(marker, IMarker.CHAR_END, -1);
	}

	/**
	 * Returns the starting character offset of the given marker.
	 *
	 * @param marker the marker
	 * @return the starting character offset, or <code>-1</code> if not set
	 * @see IMarker#CHAR_START
	 * @see IMarker#getAttribute(java.lang.String,int)
	 */
	public static int getCharStart(IMarker marker) {
		return getIntAttribute(marker, IMarker.CHAR_START, -1);
	}

szarnekow:
I see. Makes sense. Length was probably zero before, since charEnd -
charStart would not return -1.

PR #222, opened 2015-05-05
[incremental standalone builder] tests, proper IAllContainerStateAdapter, remove deleted resources

svenefftinge:
- Added tests for the incremental builder
- some refactorings to make the state less shared
- ‘FlatResourceSetBasedAllContainersState’ uses ResourceDescriptionsData if existent.
- some minor other things.


svenefftinge:
@JanKoehnlein could you have a look?

plugins/org.eclipse.xtext.builder.standalone/src/org/eclipse/xtext/builder/standalone/incremental/BuildRequest.xtend
```diff
@@ -25,6 +35,12 @@ class BuildRequest {
 	List<URI> dirtyFiles = newArrayList
 	List<URI> deletedFiles = newArrayList
 	
+	IIssueHandler issueHandler = new IIssueHandler.DefaultIssueHandler()
```
JanKoehnlein: Maybe we should rename this class. It used to be the request (data provided by the client only) before, now it gets additional data on setup (like the BuildContext) and collects even more during the build (used to be in the BuildResultCollector). Not sure whether this change is for the better.

intellij/org.eclipse.xtext.idea.builder/src/org/eclipse/xtext/idea/build/daemon/XtextBuildResultCollector.xtend
```diff
@@ -11,14 +11,14 @@ import com.google.common.collect.HashMultimap
...
 import org.eclipse.emf.common.util.URI
-import org.eclipse.xtext.builder.standalone.incremental.IncrementalStandaloneBuilder.FileListener
+import org.eclipse.xtend.lib.annotations.Accessors
 import org.eclipse.xtext.idea.build.net.Protocol.BuildResultMessage
 import org.eclipse.xtext.idea.build.net.Protocol.GeneratedFile
 
 /**
  * @author Jan Koehnlein - Initial contribution and API
  */
-class XtextBuildResultCollector implements FileListener {
```
JanKoehnlein: Is that class obsolete now?

PR #224, opened 2015-05-07
[xtend] don’t index xtend sources from jars (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=466667)

svenefftinge:


szarnekow:
I wonder if we can still safely navigate to the Xtend sources if we link against the java class. JDT will only give us a simple name to find the trace file et al.

svenefftinge:
Yes, we can. We used to link against them already, they are just no longer indexed.

szarnekow:
Looks like this is save to merge into master. The test failures seem to be unrelated.

PR #238, opened 2015-05-18
Non final tasks

andreasmuelder:
removed 'final' from class Task to allow addition of custom data


andreasmuelder:
closeing and recreate pull request with valid CLA

PR #248, opened 2015-05-22
[467477] Guard against the internal NoValue: java.lang.Object

szarnekow:
JDT may expose an internal represenation of a missing annotation
value wrapped in an array.

see https://bugs.eclipse.org/bugs/show_bug.cgi?id=467477


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=467477

PR #252, opened 2015-05-26
ant target for executing tests

LorenzoBettini:
It would be useful to have an ant target that executes also the tests (I
took the launch configuration from
https://xtext-builds.itemis.de/jenkins/job/xtext-pull-request-verifier,
so that this should mimic what is executed for testing pull requests).


LorenzoBettini:
@dhuebner could you please have a look at this?

dhuebner:
I'm okay with the new ant target, but I'm not sure we can keep it in sync in the long term. At least I can't maintain the list of launch configs to be executed.

LorenzoBettini:
There's already a similar target, but it was only using a launch configuration for some graph tests.
In any case, I can try to keep it in sync in the future.

LorenzoBettini:
can then this be merged or do you think it's not worthwhile?

PR #256, opened 2015-05-28
Bug 468641 [xbase] issues with JvmTypesBuilder toGetter/toSetter

LorenzoBettini:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=468641

### DIFF
--- a/plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/jvmmodel/JvmTypesBuilder.java
+++ b/plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/jvmmodel/JvmTypesBuilder.java
@@ -49,6 +49,8 @@ import org.eclipse.xtext.common.types.util.AnnotationLookup;
 import org.eclipse.xtext.common.types.util.TypeReferences;
 import org.eclipse.xtext.documentation.IEObjectDocumentationProvider;
 import org.eclipse.xtext.naming.QualifiedName;
+import org.eclipse.xtext.nodemodel.ICompositeNode;
+import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
 import org.eclipse.xtext.util.Pair;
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=468641

svenefftinge:
Actually it is perfectly ok, to resolve the proxy here. 
The only thing that we must avoid is to resolve an XComputedTypeReference.

LorenzoBettini:
If the proxy is resolved at that point (during toGetter), then https://bugs.eclipse.org/bugs/show_bug.cgi?id=468174 won't work, but I saw that you left a comment in that bug as well, about the solution being wrong.

However, from what I understand, toGetter is the only place where proxy resolution is not delayed, and I seemed to understand that proxies should not be resolved during the model inferrer, that's why I thought that the current implementation was somehow wrong anyway.

The additional cloning of the parameter's type in toSetter looks wrong anyway.

svenefftinge:
The caution you see in the code is a relict of times when we hadn't fully understood what could be done and what not. :-| 
During (post indexing) model inference it is ok to type proxies. It is only not ok to depend on JvmType's members (since we are in the process of creating them).

LorenzoBettini:
So should I modify the patch so that only toSetter implementation is changed? I think that the additional cloning in toSetter is wrong anyway :)

And we keep on discussing the other bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=468174 there?

svenefftinge:
Extracting the method isBooleanType is also a good improvement. But it should just check for isInferred and otherwise directly check getType.

LorenzoBettini:
So now getType is used to check whether it's a primitive boolean (if not inferred); I also made it protected API (if I understand correctly what you said).

LorenzoBettini:
Hi, did you see the updated commits?

svenefftinge:
Thanks!

PR #259, opened 2015-06-01
[467565] Improved performance for case insensitive languages

szarnekow:
- Use QualifiedName.equalsIgnoreCase if appropriate
- Deserialize lowercase QN as lowercase instances from index

see https://bugs.eclipse.org/bugs/show_bug.cgi?id=467565


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=467565

PR #280, opened 2015-06-22
[idea auto builder] only build sources that belong to the currently built module, part 2

svenefftinge:


plugins/org.eclipse.xtext.builder.standalone/src/org/eclipse/xtext/builder/standalone/incremental/IncrementalBuilder.xtend
```diff
@@ -71,9 +77,10 @@ import org.eclipse.xtext.validation.CheckMode
 					if (resource.validate) {
 						resource.generate(request, newSource2GeneratedMapping)
 					}
-					return true
+					val old = request.previousState.resourceDescriptions.getResourceDescription(resource.URI)
+					return new DefaultResourceDescriptionDelta(old, copiedDescription)
```
szarnekow: this should probably be IResourceDescription.Manager.createDelta(IResourceDescription, IResourceDescription)

svenefftinge: +1 have that in an upcoming commit. :-)

PR #284, opened 2015-06-23
[xbase] reassignments in blockexpressions refine the type of a local variable.

svenefftinge:


spoenemann:
If the change is still relevant, please reapply it to xtext-extras.

PR #292, opened 2015-06-24
Fixed navigation to derived elements

akosyakov:


akosyakov:

PR #295, opened 2015-06-25
add createFileUri(File) helper method

oehme:
makes sure that folder URIs end with a trailing slash, so that they can
be used with isPrefix(), resolve() and deresolve()


oehme:
Jenkins test

PR #297, opened 2015-06-25
Clean up of common types and xbase

akosyakov:


akosyakov:

PR #298, opened 2015-06-25
Introduced IShouldGenerate as a platform independent hook, to suppress code generation.

svenefftinge:


plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/EclipseBasedShouldGenerate.xtend
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * @author Sven Efftinge - Initial contribution and API
+ */
+class EclipseBasedShouldGenerate implements IShouldGenerate {
+
+	@Inject IWorkspaceRoot workspaceRoot
+	@Inject IStorage2UriMapper storage2UriMapper
+
+	override shouldGenerate(Resource resource, CancelIndicator cancelIndicator) {
```
szarnekow: This will build other resources from the resource set, too since the project context is not passed from the builder to this method? I'd think IShouldGenerate should also take the current project configuration?

svenefftinge:
Jenkins test

svenefftinge:
Jenkins test

PR #319, opened 2015-06-30
UriBasedFileSystemAccess does postProcessing

svenefftinge:


plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/URIBasedFileSystemAccess.xtend
```diff
@@ -73,9 +77,10 @@ class URIBasedFileSystemAccess extends AbstractFileSystemAccess2 {
 	
 	override generateFile(String fileName, String outputCfgName, CharSequence contents) {
 		val uri = getURI(fileName, outputCfgName)
-		generateTrace(fileName, outputCfgName, contents)
 		val encoding = getEncoding(uri)
```
szarnekow: Here we need to take encoding from the language of the generated file rather than the language of the injector that instantiated the class. Right now it seems like all files will be produced with the runtime encoding.

svenefftinge: It should be the implementation of the source language not the generated one that detects the encoding. We don't have injectors for target languages.

szarnekow: Hmm I wonder how this could work, e.g. if your source lang is XML you have the XMLEncodingProvider. That one would not be able to provide encoding for other files in general (and vice versa). Also other langs shouldn't care that some file type has a fixed encoding. For example, properties files should usually be ISO-8859-1 whereas most other languages are likely to be UTF-8.
Also we do this already - JavaIoFileSystemAccess.getEncoding(URI)) will only fall back to the source languages encoding provider if the target language is not known the Xtext.

szarnekow: Giving this another though, I think your right - the generating language should control the encoding of the files that it writes. An XMLEncodingProvider won't be able to determine an encoding for a file that was not yet written to disk anyway.
Since there are a few langs with a fixed encoding (or explicit encoding in the header), I think an overloaded IFileSystemAccess2.generateFile(.., CharSequence) may be the best we could do. It should be possible to pass an explicit Charset whenever a text is written to disk. By default the encoding provider of the generating language should be used to determine the default Charset.

PR #323, opened 2015-07-01
[idea] type completion inserts import

svenefftinge:


intellij/org.eclipse.xtext.idea/src/org/eclipse/xtext/xbase/idea/completion/XbaseCompletionContributor.xtend
```diff
@@ -73,11 +87,50 @@ class XbaseCompletionContributor extends XtypeCompletionContributor {
...
+					}
 					completionResultSet.addElement(it)
 				}
 			}
 		]
 	}
+	
+	
+	@Inject RewritableImportSection.Factory factory
+	def void updateInsertHandler(JavaPsiClassReferenceElement element) {
```
szarnekow: Is this method obsolete?

PR #341, opened 2015-07-07
New version of Xtext generator

spoenemann:
Old generator fragments can be used in the new XtextGenerator by wrapping them into a FragmentAdapter. A workflow then looks like this:

```
var grammarURI = "classpath:/org/xtext/example/mydsl/MyDsl.xtext"
var projectName = "org.xtext.example.mydsl"
var runtimeProject = "../${projectName}"
var fileHeader = "/*\n * generated by Xtext \${version}\n */"
Workflow {
	component = XtextGenerator auto-inject {
		configuration = {
			project = model.WizardConfig {
				runtimeBase = runtimeProject
				eclipseEditor = true
				ideaEditor = true
				orionEditor = true
				genericIdeSupport = true
				testingSupport = true
			}
			code = auto-inject {
			}
		}
		language = auto-inject {
			uri = grammarURI

			fragment = adapter.FragmentAdapter {
				fragment = grammarAccess.GrammarAccessFragment auto-inject {}
			}

			fragment = adapter.FragmentAdapter {
				fragment = ecore.EMFGeneratorFragment auto-inject {}
			}

			...
		}
	}
}
```


szarnekow:
Nice.

Would it make sense to have a language adapter, too? With an additional addFragment for the old stuff such that the fragment adapter is instantiated internally and not repeated so often?

svenefftinge:
I think we should write dedicated adapters for those fragments that we haven't converted and have the generic one for externally defined fragments. That also allows for removing or replacing some of the old fragments (by not providing an adapted version).

szarnekow:
Sounds good to me, too.

PR #351, opened 2015-07-10
[472236] Added null-guard in XtextEditor

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472236

PR #357, opened 2015-07-13
[472142] Fixed compiler problem with instanceof, casts and nullsafe member access

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472142

PR #358, opened 2015-07-13
[469679] Guard against missing adapter

szarnekow:
Don’t write streams with bogus content to the stored resource. [1]
Also revert incomplete state that was read from storage. See bug 471535 [2]

[1] https://bugs.eclipse.org/bugs/show_bug.cgi?id=469679
[2] https://bugs.eclipse.org/bugs/show_bug.cgi?id=471535


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=469679

PR #359, opened 2015-07-13
Dh bug472478

dhuebner:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472478

plugins/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/actions/ImportsAwareClipboardAction.java
```diff
@@ -181,47 +180,65 @@ public XbaseClipboardData apply(Clipboard input) {
...
-		EObject targetElement = xtextEditor.getDocument().priorityReadOnly(new IUnitOfWork<EObject, XtextResource>() {
-
-			@Override
-			public EObject exec(XtextResource state) throws Exception {
-				EObject elementAtOffset = helper.resolveContainedElementAt(state, 10);
-				return elementAtOffset;
-			}
-		});
+		if (COMMENT_PARTITION.equals(typeRight) || STRING_LITERAL_PARTITION.equals(typeRight)
+				|| SL_COMMENT_PARTITION.equals(typeRight) || "__rich_string".equals(typeRight)) {
```
szarnekow: Should we have a service / an extension interface for the partition type mapper that returns the list of partitions where imports shouldn't be added? Here we add a dep on Xtends partition type in xbase

dhuebner: That would be nice to have

PR #363, opened 2015-07-13
[472013] Added NPE guard on breakpoint condition page

szarnekow:
Also made hyperlink presenters more failsafe


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472013

PR #367, opened 2015-07-14
Consider '/* ... EOF' as a multiline comment as well

akosyakov:
Change-Id: Id948e71fef750102b9f54f2bff589f4b811ec70d
Signed-off-by: akosyakov anton.kosyakov@itemis.de


spoenemann:
If the change is still relevant, please reapply it to xtext-xtend.

PR #368, opened 2015-07-14
[470767] Don’t bail out on missing class files for nested classes

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=470767

PR #374, opened 2015-07-14
[idea] improved framework detection improved framework detection to add xtend-gen relative to source (i.e. work for maven layout out-of-the-box. Also the source folders are now marked as generated.

svenefftinge:


intellij/org.eclipse.xtend.idea/src/org/eclipse/xtend/core/idea/config/XtendSupportConfigurable.xtend
```diff
@@ -33,18 +35,46 @@ class XtendSupportConfigurable extends FrameworkSupportInModuleConfigurable {
...
+			state.testOutputDirectory = xtendGenTest.canonicalPath
 		val libraryTable = LibraryTablesRegistrar.getInstance().getLibraryTable()
 		var library = libraryTable.getLibraryByName(XtendLibraryDescription.XTEND_LIBRARY_NAME)
 		if (library != null && !rootModel.moduleLibraryTable.libraries.contains(library)) {
 			rootModel.addLibraryEntry(library)
 		}
 	}
+	
+	private def VirtualFile getOrCreateDir(VirtualFile parent, String name) {
+		val existing = parent.children.findFirst[it.name==name]
```
szarnekow: parent.findOrCreateChildData(..) - the requestor should also not be null according to the docs

szarnekow: My bad - createChildData creates a file apparently, though com.intellij.openapi.vfs.VfsUtil.createDirectoryIfMissing(VirtualFile, String) may do the trick

PR #375, opened 2015-07-14
[471384] Yet another test that needs a real auto build

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=471384

PR #379, opened 2015-07-15
[469704] Newly created classpath entry inherits configuration from source

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=469704

PR #380, opened 2015-07-15
[470451] Fixed XtendReferenceFinder

szarnekow:
IResourceAccess may be null thus we have to check
it before we access the resource


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=470451

PR #381, opened 2015-07-15
[469805] Handle bogus value converter configuration

szarnekow:
If an AbstractLexerBasedConverter is configured
with a datatype rule, it will be disabled and an
error will be logged.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=469805

PR #396, opened 2015-07-17
[471384] Fixed issue in dirty state tests and editor support

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=471384

PR #399, opened 2015-07-17
[idea] prevent (follow-up) NPEs

svenefftinge:


svenefftinge:

intellij/org.eclipse.xtext.idea/src/org/eclipse/xtext/idea/build/BuildProgressReporter.xtend
```diff
@@ -57,7 +57,7 @@ class BuildProgressReporter implements BuildRequest.IPostValidationCallback {
 
 	override afterValidate(URI validated, Iterable<Issue> issues) {
 		markAsAffected(validated)
-		for (issue : issues) {
+		for (issue : issues.filterNull) {
```
szarnekow: Would it be better if we prevent null-issues in the iterable in the first place?

svenefftinge: The iterable comes from a language implementation, so we cannot guarantee it. The default impl, shouldn't do that, though.

szarnekow: I see - I got the impression that you had an NPE and added the guard afterwards here.

svenefftinge: Yes, I had one. I wanted to say that we cannot generally make sure that IResourceValidators are implemented correctly. So this null guard should be there.

PR #403, opened 2015-07-18
[472602] Problem with lambda expression in default generic method

szarnekow:
Improved the situation but a holistic fix is not possible
/ reasonable in the maintenance branch.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472602

PR #405, opened 2015-07-19
[469678] Check if an array with component type ‘void’ is to be created

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=469678

PR #409, opened 2015-07-20
[466702] Don’t highlight postfix operators as extensions

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=466702

PR #418, opened 2015-07-23
[idea] improved completion context matching

svenefftinge:
The pattern-based approach had way too many false positive matches, so I replaced 
them by our parser-based follow element computation. Instead of passing a pattern, you now directly pass abstract elements or eStructuralFeatures.

Advantages over the ContentAssistContextFactory-hook, is that the PsiFile’s XtextResource is not resolved and that we have the full IntelliJ API available.


intellij/org.eclipse.xtend.idea/src/org/eclipse/xtend/core/idea/completion/XtendCompletionContributor.xtend
```diff
@@ -58,7 +56,7 @@ class XtendCompletionContributor extends AbstractXtendCompletionContributor {
 	protected def completeAbstractSuperMethods() {
 		extend(
 			CompletionType.BASIC,
-			psiElement.withEReference(XtendPackage.Literals.XTEND_TYPE_DECLARATION__MEMBERS)
+			grammarAccess.memberAccess.newKeyword_2_2_2
```
szarnekow: A little hacky - would it work to pass the EStructuralFeature XtendType.members here, instead?

svenefftinge: I changed it to use the feature. As it is just computing the elements once on startup, I don't think it is a problem performance-wise. 
Changing it revealed a serious bug, which is fixed in the updated commit. thx.

svenefftinge:

svenefftinge:

PR #429, opened 2015-07-28
[idea] added ‘idea.config.path’ variable to point to project-local directory

svenefftinge:


svenefftinge:

PR #444, opened 2015-08-01
[473079] Fixed mvn test expectations

szarnekow:
The tests were still based on the assumption that all
trace regions are relevant for debugging. Since this
information is now persisted, the expectations had to
be adjusted too.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=473079

PR #453, opened 2015-08-05
[trace] removed unused Type parameters and unused inverseView methods

svenefftinge:


svenefftinge:
I removed some parts that were not really used in order to simplify the code a bit. 
I think we could go a step further and remove the PlatformSpecific variations by working passing strategies to convert SourceRelativeUri to an AbsoluteUri and another to load contents for an AbsoluteUri.

The downside would be that clients than no langer can directly obtain IStorage / VirtualFile but have to go through their platform-specific URI-to-Platform service (e.g. Storage2URIConverter). The upside is that we can have simpler type hierarchy for traces, trace provider and location.

svenefftinge:

PR #454, opened 2015-08-05
[xtext.xtext] use new Serializer instead of ParseTreeConstructor

meysholdt:
The 'SerializerFragment' is intentionally not used because it generates
Java files which would become public API. 

The Serializer is fully functional without the SerializerFragment.


meysholdt:
the failed tests seem unrelated; they don't fail locally on my machine.

PR #457, opened 2015-08-06
[xtext.xtext] use IResourceDescription.Manager for scoping

meysholdt:
...instead of GlobalResourceDescriptionProvider. 

Because when the grammar is loaded from an .xtextbin file, 
the GlobalResourceDescriptionProvider returns null, because there
is no IResourceServiceProvider registered for *.xtextbin.


meysholdt:

meysholdt:
this fixes org.eclipse.xtext.parsetree.reconstr.XtextGrammarReconcilationTest.testSelf()

szarnekow:
Looks like the only client of the GlobalResourceDescriptionProvider is now gone. Deprecate and schedule for removal?

meysholdt:
It’s also used in 
org.eclipse.xtext.scoping.impl.ResourceSetGlobalScopeProvider.resourceDecriptionProvider

szarnekow:
ResourceSetGlobalScopeProvider doesn't seem to be used either in our code... anyway, people may still use it, that's true.

meysholdt:
regarding "Jenkins/xtext-runtime-tests-master-latest", the remaining failing tests have been fixed by https://github.com/eclipse/xtext/commit/99f3524658b9da7542b4080a62e1d732b2155856 already.

PR #461, opened 2015-08-07
[idea] fixed problem with deletion of folders not triggering events for its children, registered IndexedJvmTypeAccess for auto builder

svenefftinge:


svenefftinge:

PR #467, opened 2015-08-11
[xbase][bug 466974] Create a variable for an argument if it has side effects and one of following arguments requires a variable declaration also

akosyakov:
Change-Id: Ibe68a5f012267acf96f55e69300aa0761bb5f42e
Signed-off-by: akosyakov <anton.kosyakov@itemis.de>


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=466974

PR #470, opened 2015-08-12
[idea] unify how we obtain virtual files for psi elements plus some other minor changes

svenefftinge:


svenefftinge:

PR #472, opened 2015-08-12
Bug 390789 - No marker when class name does not match file name

dhuebner:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=390789

PR #481, opened 2015-08-14
[263773] Reduce size of generated parsers

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=263773

PR #493, opened 2015-08-20
[410140] Extract Local Variable refactoring suggests invalid names fo…

JanKoehnlein:
…r arrays and generic types

Change-Id: I5a03a2f914bfa628d61042ad499753e405dbe91b

### DIFF
--- a/plugins/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/refactoring/NewFeatureNameUtil.java
+++ b/plugins/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/refactoring/NewFeatureNameUtil.java
@@ -7,14 +7,22 @@
  *******************************************************************************/
 package org.eclipse.xtext.xbase.ui.refactoring;
 
+import static com.google.common.collect.Lists.*;
 import static org.eclipse.xtext.util.Strings.*;
 
+import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.jface.viewers.StyledString;
 import org.eclipse.ltk.core.refactoring.RefactoringStatus;
+import org.eclipse.swt.graphics.Image;
 import org.eclipse.xtext.GrammarUtil;
 import org.eclipse.xtext.IGrammarAccess;
+import org.eclipse.xtext.common.types.xtext.ui.JdtVariableCompletions;
+import org.eclipse.xtext.common.types.xtext.ui.JdtVariableCompletions.CompletionDataAcceptor;
+import org.eclipse.xtext.common.types.xtext.ui.JdtVariableCompletions.VariableType;
 import org.eclipse.xtext.conversion.IValueConverterService;
 import org.eclipse.xtext.conversion.ValueConverterException;
 import org.eclipse.xtext.naming.QualifiedName;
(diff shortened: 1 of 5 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=410140

JanKoehnlein:

plugins/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/refactoring/NewFeatureNameUtil.java
```diff
@@ -121,10 +121,18 @@ protected String getFancyDefaultName(XExpression expression) {
...
+		return toVariableName(type);
+	}
+	
+	protected String toVariableName(LightweightTypeReference type) {
+		if(type != null) {
+			if(type.isPrimitive())
 				return type.getSimpleName().substring(0, 1);
-			else
+			else if(type.isArray()) 
+				return toVariableName(type.getComponentType()) + 's';
```
szarnekow: cosmetic: two dimensional arrays would get more than one 's' in the end, simple names with 's' and the end would be strange, too

szarnekow: Can we reuse smth from org.eclipse.xtext.common.types.xtext.ui.JdtVariableCompletions or extract common parts?

JanKoehnlein: IMHO, JdtVariableCompletions has a pretty cryptic API and involves StyledText. I'd rather stick with something simpler.

szarnekow: Yes, the API may not be optimal for this use case. On the other hand, the current Java project carries some user settings how variables are supposed to be named and we could (should) use these were possible to provide a consistent experience. Maybe you can factor the important parts into something easier to reuse?

JanKoehnlein: Note that the JdtVariableCompletions is about variables in Java code while NewFeatureNameUtil is about Xtend which has e.g. different keywords.

JanKoehnlein:
New version using JdtVariableCompletions.

Advantage:
- configured prefixes/suffixes for Java variables are used

Disadvantages
-The auto-pre/suffixing takes place inside NamingConventions, as well as adding a counter, so I had to add all feature names from the current feature call scope to the excluded names (as well as all Xtend keywords). May not be the most efficient thing to do.
- Our impl now extracts 'strings' for 'List<String>', while JDT extracts 'list'. So Xtend refactoring has different proposals form JDT
- Java keywords which are not Xtend keywords will not be valid, e.g. extracting "new Goto()" will result in the default variable name 'goto1' instead of 'goto'
- JDT does not convert UPPER_CASE_UNDERSCORE to camelCase as our former impl did.

JanKoehnlein:

dhuebner:
Jenkins xtend-ide-tests-master-oldest

JanKoehnlein:

PR #496, opened 2015-08-24
[osgi] xtext.m2e and xtend.m2e can not be started due to Package uses conflict - bug475706

dhuebner:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475706

PR #499, opened 2015-08-25
XtextGenerator improvements and standalone wizard

oehme:
This is a first working version of the improved XtextGenerator that will be used by the new standalone wizard. Unfortunately I had to copy quite a bit of code from the old Generator, as it was API.

There is still a lot of work to do: Translating more generator fragments, migrating more of our own workflows to the new generator, deleting the idea.generator and web.generator projects etc.

Also, the new wizard is not yet connected to the wizard page in Eclipse.

I'd like to merge this now, so others can keep working on fragments without too many merge conflicts.

I split the code into src and src-gen commits for easier review.


oehme:
Jenkins test

PR #500, opened 2015-08-25
[idea][bug 475771] when possible use document content as an input stream for the given URI

akosyakov:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475771

akosyakov:

intellij/org.eclipse.xtext.idea/src/org/eclipse/xtext/idea/resource/IdeaResourceSetProvider.xtend
```diff
@@ -117,6 +117,11 @@ class IdeaResourceSetProvider {
 				throw new FileNotFoundException("Couldn't find virtual file for "+uri)
 			}
 			
+			val cachedDocument = FileDocumentManager.instance.getCachedDocument(virtualFile)
+			if (cachedDocument != null) {
+				return new ByteArrayInputStream(cachedDocument.text.bytes)
```
szarnekow: This looks like a potential encoding issue to me. We need to use the same charset to produce the bytes as we'll use later to produce the content. Maybe we can use the LazyStringInputStream here?

PR #501, opened 2015-08-25
JdtBasedProcessorProvider isOutputFolderIncluded()

vorburger:
Hi guys, how about this? If you were to accept this, then I could simplify http://git.eclipse.org/c/emf/org.eclipse.emf.eson.git/tree/plugins/org.eclipse.emf.eson/src/org/eclipse/emf/eson/generators/JdtBasedClassLoaderProvider.java to just do @Override isOutputFolderIncluded() { return true; } for what I need, instead of copy/pasting the entire createClassLoaderForJavaProject() .. so then I would not have to "catch up" doing https://git.eclipse.org/r/#/c/54483/1/plugins/org.eclipse.emf.eson/src/org/eclipse/emf/eson/generators/JdtBasedClassLoaderProvider.java when you do https://github.com/eclipse/xtext/commit/5d3aa9798fc1b04eac08e645b0022821af433e14 ... ;-)


szarnekow:
Could you please update the PR and use tabs instead of spaces. The formatting changes make it really hard to get the gist of it.

vorburger:
@szarnekow yep; fixed now, and added description why this..

szarnekow:
Change looks good to me though I want to mention that the JdtBasedProcessorProvider is not API and will probably change in the future.

szarnekow:
Jenkins test

PR #505, opened 2015-08-27
Connect the new wizard to the existing UI

oehme:
Eclipse plugin projects work as before.

Gradle and Maven will need more polish. We should think about m2e/Buildship integration to make setup easier for our users.


oehme:

oehme:

PR #506, opened 2015-08-27
[idea][gradle] Better support for gradle and android projects

dhuebner:
Signed-off-by: Dennis Huebner <dennis.huebner@itemis.de>


dhuebner:
Jenkins test

PR #508, opened 2015-08-28
[wizard] generate correct build files for Xtext snapshot versions

oehme:


plugins/org.eclipse.xtext.xtext.wizard/xtend-gen/org/eclipse/xtext/xtext/wizard/XtextVersion.java
```diff
@@ -0,0 +1,56 @@
+package org.eclipse.xtext.xtext.wizard;
+
+import org.eclipse.xtend.lib.annotations.Data;
+import org.eclipse.xtext.xbase.lib.Pure;
+
+@Data
```
szarnekow: This cannot be used from MWE2

oehme: No, it's currently only used in the wizard and will never be directly exposed in the Generator API. 

We might want it in the Generator at some point (as well as BuildSystem, SourceLayout etc.), but then we will just offer String-valued setters on the Generator.

PR #510, opened 2015-08-28
[idea][bug 475906] Added GeneratedSourceAwareUsageSearcher to fix navigation from the find usages view

akosyakov:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475906

akosyakov:

akosyakov:
Rejected while we want to show usages in the generated code

PR #511, opened 2015-08-31
[idea] Rehighlight affected files after the auto build

akosyakov:


akosyakov:

PR #513, opened 2015-09-01
Proposed update about how to add TestmodelPackage

vorburger:
https://git.eclipse.org/r/#/c/54972/1/tests/org.eclipse.emf.eson.tests/src/org/eclipse/emf/eson/tests/util/ESONWithTestmodelInjectorProvider.java

http://blog2.vorburger.ch/2015/09/noclassdeffounderror-s-in-eclipse.html

### DIFF
--- a/xtext-website/documentation/303_runtime_concepts.md
+++ b/xtext-website/documentation/303_runtime_concepts.md
@@ -1226,10 +1226,19 @@ You may also need to initialize 'import'-ed ecore models that are not generated 
   class MyLanguageStandaloneSetup extends MyLanguageStandaloneSetupGenerated {
 
     def static void doSetup() {
-      MyPackageImpl.init
+      MyModelPackage.eINSTANCE.name // NOT MyPackageImpl.init, as that won't initialize MyModelPackage statics
       new MyLanguageStandaloneSetup().createInjectorAndDoEMFRegistration
     }
 
+  }
+```
+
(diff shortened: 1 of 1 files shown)


vorburger:
I've removed the "static" as suggested by @akosyakov (and tested in https://git.eclipse.org/r/#/c/55020/), and added this missing Signed-off-by ... good to merge?

svenefftinge:
Hi Michael,

I have replaced the call to MyPackageImpl.init() in the documentation, which was the main issue, if I'm not mistaken. Is it fine to close this PR now?

vorburger:
Hallo @svenefftinge, I cannot see your change in https://github.com/eclipse/xtext/commits/master/xtext-website/documentation/303_runtime_concepts.md, normal? There were two issues here really, 1. the wrong suggested use of MyPackageImpl.init instead of MyModelPackage.eINSTANCE.name (so that MyModelPackage statics get correctly initialized), and 2. the need to repeat a EPackage.Registry.INSTANCE.put. It didn't work without that. BTW: It doesn't have to be static, inside internalCreateInjector() as done in https://git.eclipse.org/r/#/c/55020/1/tests/org.eclipse.emf.eson.tests/src/org/eclipse/emf/eson/tests/util/ESONWithTestmodelInjectorProvider.java works too.

svenefftinge:
It is fixed on the maintenance branch, which gets published automatically . See https://www.eclipse.org/Xtext/documentation/303_runtime_concepts.html#how-to-write-tests-that-includes-multiple-different-languages

svenefftinge:
I think the suggestion fixes the problems you had, no?

vorburger:
Yup https://github.com/eclipse/xtext/commit/e4fbd29df559217472e7430a3dd412f5ba09f59e covers this; tx.

PR #515, opened 2015-09-01
[idea] java converter - Bug 474278 , Bug 474277

dhuebner:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=474278

PR #518, opened 2015-09-01
[tests] trying to fix a test that fails on the server only

svenefftinge:


dhuebner:
I noticed that the test fails only on older eclipse platform e.g. Helios, with Neon it works fine.
The same result is on the server side.

PR #521, opened 2015-09-03
[idea] Fixed failed IntentionsTest because of a different JDK version

akosyakov:


akosyakov:

PR #529, opened 2015-09-04
[474428] register current source folder before using FSA

svenefftinge:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=474428

PR #539, opened 2015-09-08
[j2x] PrefixOp as Block Bug 476813. Prefix op in for - Bug 476781

dhuebner:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=476813

PR #543, opened 2015-09-08
[475337] - Made JvmModelGenerator more robust against incomplete models. Handle null passed to TreeAppendable#append more gracefully.

svenefftinge:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475337

svenefftinge:
Jenkins test

szarnekow:
Shouldn't we also make sure that we do not produce enum literals without a name?

svenefftinge:
I think, it is better to pass on a literal without a name. Especially, since we cannot enforce such a contract on every Xbase client.

szarnekow:
We filter all other members that have a null name and do not produce JVM equivalents for them (for good reasons, since most of the code wouldn't expect JvmIdentifiable.getSimpleName to return null). It should at least be symmetric for enum literals. I would be in favor of filtering members with a null name.

svenefftinge:
So you mean filtering those for Xtend in addition to being null-tolerant in the JvmModelGenerator?

szarnekow:
Filtering in addition. See also JvmTypesBuilder.toXYZ(..) which are all guarded by if (name==null) return null

szarnekow:
Which makes me wonder why we do not call org.eclipse.xtext.xbase.jvmmodel.JvmTypesBuilder.toEnumerationLiteral(EObject, String) in the Xtend inferrer

szarnekow:
I see - because we do not really use the types builder in the xtend inferrer but only for a few special things.

svenefftinge:
AFAICS we also don't check XtendFunctions for getName, but it seems like there is no way the parser would create such an AST.

szarnekow:
It's checked here:
if (member instanceof XtendField
					|| (member instanceof XtendFunction && ((XtendFunction) member).getName() != null)
					|| member instanceof XtendConstructor) {
				transform(member, inferredJvmType, true);
			}

szarnekow:
Maybe we also want to add a test. I think the bug is reproduced by this enum:
```
enum MyEnum {
	First ,, Third	
}
```

PR #545, opened 2015-09-08
[idea] FrameworkSupport - use modifiable model instead of module

dhuebner:


dhuebner:

PR #549, opened 2015-09-09
[idea][xbase] Reference searcher based on jvm elements instead of traces

akosyakov:


akosyakov:

intellij/org.eclipse.xtext.idea/src/org/eclipse/xtext/idea/completion/AbstractCompletionContributor.xtend
```diff
@@ -55,8 +55,8 @@ abstract class AbstractCompletionContributor extends CompletionContributor {
 	@Inject protected IGrammarAccess grammarAccess
 	@Inject protected extension TokenSetProvider
 	
-	@Inject IContentAssistParser contentAssistParser
-	@Inject FollowElementComputer followElementComputer
+	@Inject(optional=true) IContentAssistParser contentAssistParser
```
szarnekow: Why marked as optional if it is not really optional?

PR #550, opened 2015-09-09
[idea][builder] Write to a cached document (if one exists) instead of directly to a virtual file to avoid conflicts

akosyakov:


akosyakov:

PR #552, opened 2015-09-10
[475906] fixed wrong length parameter for TextRange

svenefftinge:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475906

PR #553, opened 2015-09-10
[idea][navigation] Enabled jvm element based reference searcher

akosyakov:


akosyakov:

PR #556, opened 2015-09-10
[472489] Support class splitting in ExternalAntlrLexerFragment

jsebechlebsky:
Support for class splitting of lexer class in ExternalAntlrLexerFragment, 
requested by  [472489](https://bugs.eclipse.org/bugs/show_bug.cgi?id=472489).

Since Antlr predicates in custom lexer can reference non-static fields or methods,
DFA classes must stay non-static when class splitting is run from ExternalAntlrLexerFragment.

Signed-off-by: Jan Sebechlebsky <jan.sebechlebsky@cz.ibm.com>

### DIFF
--- a/plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/parser/antlr/splitting/AntlrLexerSplitter.java
+++ b/plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/parser/antlr/splitting/AntlrLexerSplitter.java
@@ -43,7 +43,9 @@ public class AntlrLexerSplitter {
 
 	private StringBuilder stringBuilder;
 	private final Scanner scanner;
-
+	
+	private boolean allowDFAStaticClasses;
+	
 	public AntlrLexerSplitter(String content) {
 		scanner = new Scanner(content);
 	}
@@ -70,7 +72,10 @@ public class AntlrLexerSplitter {
 		}
 		copyTail();
 		String result = stringBuilder.toString();
-		result = new LexerSpecialStateTransitionSplitter(false).transform(result);
+		LexerSpecialStateTransitionSplitter lexerSplitter;
+		lexerSplitter = new LexerSpecialStateTransitionSplitter(false);
+		lexerSplitter.setAllowDFAStaticClasses(allowDFAStaticClasses);
+		result = lexerSplitter.transform(result);
 		return result;
 	}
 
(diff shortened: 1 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472489

plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/parser/antlr/splitting/internal/LexerSpecialStateTransitionSplitter.java
```diff
@@ -27,10 +27,11 @@
 			
 	public static final Pattern CASE_PATTERN = Pattern.compile(
 			"(^\\s*case\\s+(\\d+)\\s*:(\\s*))" +// case # -> $1, $2, $3
-			"([^;]*;)" + // int .. = input.LA(..) -> $4
+			"([^;]*;" + // int .. = input.LA(..) -> $4
```
szarnekow: Here is an unclosed group in the regex. Could you please mention that in the comment after the line, e.g where this group is supposed to end.

szarnekow:
Hi Ján, I reviewed the patch and left a comment there where I wondered about a change in the regular expression. Could you please update the PR?

jsebechlebsky:
Hi Sebastian, the group is closed on the next line, but maybe it would be good to have the whole group on single line. I will fix it that way, if you're OK with that. The regular expression was modified to match also code generated when semantic predicates are used. It should match every Antlr3 output generated from (Java.stg)[https://github.com/antlr/antlr3/blob/master/tool/src/main/resources/org/antlr/codegen/templates/Java/Java.stg] - cyclicDFAState rule.

szarnekow:
Thanks for the updated PR. I'll give the change a try and regenerated all languages with it to see if it has any surprising side effect. Please ping me if you don't hear something until Friday :)

plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/parser/antlr/splitting/AntlrLexerSplitter.java
```diff
@@ -43,7 +43,9 @@
 
 	private StringBuilder stringBuilder;
 	private final Scanner scanner;
-
+	
+	private boolean allowDFAStaticClasses;
```
szarnekow: this should default to true

szarnekow:
Hi Jan, I regenerated some of our test languages with your patch applied and I think there's a problem with the default value of the allowDFAStaticClasses flag. Many lexer DFAs of the test langs are no longer static.

jsebechlebsky:
Sorry for not checking that twice... I've pushed the change.

szarnekow:
Thank you for the patch.

PR #558, opened 2015-09-10
[475368] Another approach to solve the deadlock issue

szarnekow:
Reduced the lock size and deferred code that may need
the ws lock to another thread such that deadlocks should
become less likely.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475368

szarnekow:
Jenkins test

svenefftinge:
I wonder if we should treat the storage2urimapper state like the rest of the builder state. I.e. persist it and only recompute / update within the build hook.
Anyway, I'll have a closer look at this PR tomorrow.

szarnekow:
> only recompute / update within the build hook.

That wouldn't work unfortunately since we need the element change event rather than the resource delta.

svenefftinge:
looks good to me

PR #562, opened 2015-09-11
[475368] Another approach to solve the deadlock issue

szarnekow:
Reduced the lock size and deferred code that may need
the ws lock to another thread such that deadlocks should
become less likely.

I figured that it's not the JobManager's scheduling rule that's causing trouble
(JDT doesn't even use a SR for the ops that we are calling) but the
workspace lock. Adjusted the change accordingly.


szarnekow:

PR #563, opened 2015-09-11
Handling cancellation of install derived state

akosyakov:


akosyakov:

PR #565, opened 2015-09-14
[idea] Highlight usages of simple and compound operators

akosyakov:


akosyakov:

PR #567, opened 2015-09-14
[477182] fixed NPE and IAE, on null entries and non existing projects.

svenefftinge:
On project init (when the state gets loaded), the modules are not populated, so I changed the map to use the module names.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=477182

svenefftinge:

PR #570, opened 2015-09-14
[wizard] make combination of all platforms possible

oehme:
Ideally we want to give the user a consistent build system. But currently we cannot build Eclipse plugins with Gradle and IntelliJ plugins with Maven. So instead the user can select a "Preferred Build System" which will be used for as many projects as possible (e.g. web can be built with either). No matter which he chooses, the following applies:

- If Eclipse UI is enabled, a Tycho build is generated, including runtime and ide
- If IntelliJ is enabled, a Gradle build is generated, including runtime and ide
- the tests for runtime/ide will be run by the "Preferred Build System"

One day we might fix this and provide full Gradle/Maven support for all platforms. Then we could rename the ComboBox to "Build System".


plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/WebProjectDescriptor.xtend
```diff
@@ -35,15 +39,14 @@ class WebProjectDescriptor extends ProjectDescriptor {
...
 		deps += createXtextDependency("org.eclipse.xtext.web.servlet")
 		deps += createMavenDependency("org.eclipse.xtend:org.eclipse.xtend.lib:${xtextVersion}")
-		deps += createMavenDependency("org.eclipse.xtext:org.eclipse.xtext.web.servlet:${xtextVersion}")
 		deps += createMavenDependency("org.webjars:requirejs:" + REQUIREJS_VERSION)
 		// RequireJS-Text is only needed for Orion
 //		deps += createMavenDependency("org.webjars:requirejs-text:" + REQUIREJS_TEXT_VERSION)
 		deps += createMavenDependency("org.webjars:jquery:" + JQUERY_VERSION)
 		deps += createMavenDependency("org.webjars:ace:" + ACE_VERSION)
 		deps += createMavenDependency("org.eclipse.jetty:jetty-annotations:" + JETTY_VERSION) => [maven.scope = Scope.PROVIDED]
-		deps += createMavenDependency("org.slf4j:slf4j-api:" + SLF4J_VERSION + "@jar") => [maven.scope = Scope.PROVIDED]
```
szarnekow: To me as a non-initiated: Why do we want to have transitive deps now but didn't need them in the first place?

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextVersion.xtend
```diff
@@ -0,0 +1,52 @@
...
+class XtextVersion {
+	
+	static def getCurrent() {
+		new XtextVersion(readVersionFromManifest ?: 'unknown')
+	}
+	
+	private static def readVersionFromManifest() {
+		var InputStream is
+		try {
+			val url = new URL(Plugin.INSTANCE.baseURL + 'META-INF/MANIFEST.MF')
```
svenefftinge: Seems like this doesn't work in a non-OSGi env, i.e. vanilla Java, does it?

szarnekow: It used to work in the past so expect this to work still :)

svenefftinge: Forget about it. I see now, that it works.

svenefftinge:
It's much better now, let's work on the outstanding things in additional commits/PRs.

plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/WebProjectDescriptor.xtend
oehme: You're right, this line should not be necessary.

PR #577, opened 2015-09-15
More generator improvements

oehme:
Split test project into runtime and ui tests
don't overwrite IDEA plugin.xml
don't create a Tycho build if the user asked for PDE

### DIFF
--- a/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/UiTestProjectDescriptor.xtend
+++ b/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/UiTestProjectDescriptor.xtend
@@ -1,11 +1,12 @@
 package org.eclipse.xtext.xtext.wizard
 
+import org.eclipse.xtext.xtext.wizard.TestProjectDescriptor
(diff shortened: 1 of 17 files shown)


plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/UiTestProjectDescriptor.xtend
```diff
@@ -0,0 +1,26 @@
...
+
+import org.eclipse.xtext.xtext.wizard.TestProjectDescriptor
+import org.eclipse.xtend.lib.annotations.FinalFieldsConstructor
+import static org.eclipse.xtext.xtext.wizard.ExternalDependency.*
+
+@FinalFieldsConstructor
+class UiTestProjectDescriptor extends TestProjectDescriptor {
+	
+	override getExternalDependencies() {
+		val deps = newHashSet
```
szarnekow: Maybe use LinkedHashSet to get a stable order?

oehme: Yep, I'll go over all such cases in another commit.

PR #581, opened 2015-09-16
[generator] some introduced GeneratedJavaFileAccess, to simplify the generated-once cases. Added overloaded methods for creating files.

svenefftinge:


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/GeneratedJavaFileAccess.xtend
```diff
@@ -0,0 +1,51 @@
...
+			�codeConfig.fileHeader�
+			package �javaType.packageName��IF appendSemicolons�;�ENDIF�
+			
+			�FOR importName : sortedImports�
+				import �importName��IF appendSemicolons�;�ENDIF�
+			�ENDFOR�
+			
+			�typeComment�
+			�FOR annot : classAnnotations�
+				�annot.generate()�
```
szarnekow: I wonder if the annotations would need information about the available, unambiguous imports to create proper code.

PR #595, opened 2015-09-17
Bug 477681 - [idea][builder] don't create annotations or semantically…

JanKoehnlein:
… highlight unopened files

After a build, we call WolfTheProblemSolver to update probelm markers.
This also triggers semantic highlighting and annotation creation for all affected files.
Both trigger a reparse/type resolve of the resource, which is quite expensive.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=477681

JanKoehnlein:

PR #598, opened 2015-09-17
[idea] Fixed NPE from the editor for an excluded file

akosyakov:


akosyakov:

PR #599, opened 2015-09-17
[idea][aa] Fixed cancellation of AA processors

akosyakov:


akosyakov:

PR #600, opened 2015-09-17
Domainmodel validator should extend generated validator

LorenzoBettini:
With this commit, the DomainmodelJavaValidator extends the generated
AbstractDomainmodelJavaValidator (which, in turn, extends
XbaseJavaValidator) instead of extending XbaseJavaValidator directly (as
it is usual in other DSLs, as far as I know). I think the only reason
why it wasn't so before is that in the workflow the
NamesAreUniqueValidator was enabled, making the generated validator use
also NamesAreUniqueValidator, which is known to cause problems with
Xbase model inferred elements. So I disabled the NamesAreUniqueValidator
in the workflow as well.

Tests are still green.


szarnekow:
Jenkins test

LorenzoBettini:
The Jenkins build failed but it does not seem to be related to this change: it fails at the very beginning :)

szarnekow:
Jenkins test

PR #601, opened 2015-09-17
Null guard in JvmTypesBuilder.isPrimitiveBoolean

LorenzoBettini:
The type reference could be null and a NPE should be avoided; this is
called by toGetter, and the problem is a regression due to my commit
4308750


szarnekow:
Jenkins test

LorenzoBettini:
The Jenkins build failed but it does not seem to be related to this change: it fails at the very beginning :)

szarnekow:
Jenkins test

szarnekow:
Yes, it was some network issue. Let's see if it works now.

PR #609, opened 2015-09-18
More Unit and Integration Tests for the Wizard

oehme:


plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ProjectDescriptor.xtend
```diff
@@ -77,7 +77,7 @@ abstract class ProjectDescriptor {
 		assignment + value.join(",\\\n" + indent)
 	}
 
-	def CharSequence manifest() '''
+	def String manifest() '''
```
szarnekow: Manifest should probably use windows line delimiter (see java.util.Manifest.write(..), even though https://docs.oracle.com/javase/7/docs/technotes/guides/jar/jar.html#JAR_Manifest doesn't impose that restriction, it seems that windows line demiliters are commonly used for Manifest files

PR #612, opened 2015-09-21
[idea builder] store installed languages in persisted index state and trigger clean-build if differs after restart.

svenefftinge:


svenefftinge:

PR #615, opened 2015-09-22
Fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=478024

cdietrich:
Makes organizing imports in the JvmModelGenerator better configurable


plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/JvmModelGenerator.xtend
```diff
@@ -122,9 +122,13 @@ class JvmModelGenerator implements IGenerator {
...
 			fsa.generateFile(type.qualifiedName.replace('.', '/') + '.java', type.generateType(generatorConfigProvider.get(type)))
 	}
+
+	protected def organizeImports() {
+		true
+	}
 	
 	def CharSequence generateType(JvmDeclaredType type, GeneratorConfig config) {
-		val importManager = new ImportManager(true, type)
+		val importManager = new ImportManager(organizeImports, type)
```
szarnekow: I think a method `createImportManager(JvmDeclaredType)` would be even better since clients could easily provide a custom instance if there is a bug in the default impl.

szarnekow:
Jenkins test

PR #627, opened 2015-09-23
[wizard] Make upgrading to Xbase seamless

oehme:


plugins/org.eclipse.xtext.xbase.ide/META-INF/MANIFEST.MF
```diff
@@ -8,7 +8,7 @@ Bundle-RequiredExecutionEnvironment: JavaSE-1.6
 Bundle-ActivationPolicy: lazy
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.xtext.ide,
- org.eclipse.xtext.xbase,
+ org.eclipse.xtext.xbase;visibility:=reexport,
```
szarnekow: I'm not convinced by this re-export. Why is that necessary?

oehme: I understand that concern with bundles outside our control, but why would anybody use xbase.ide without using xbase? I thought that was the same rationale why xbase.ui exports xbase too.

If we don't want this, then we'll have to add more require bundles to the XbaseGeneratorFragment2

szarnekow: Agreed. I was somehow concerned about re-exported re-exports, but those will be pulled into the mydsl.ide bundle via an explicit xbase dependency, too.

plugins/org.eclipse.xtext.xtext.wizard/xtend-gen/org/eclipse/xtext/xtext/wizard/WebProjectDescriptor.java
```diff
@@ -73,8 +73,10 @@ public boolean isPartOfMavenBuild() {
     final LinkedHashSet<ExternalDependency> deps = CollectionLiterals.<ExternalDependency>newLinkedHashSet();
     Set<ExternalDependency> _externalDependencies = super.getExternalDependencies();
     Iterables.<ExternalDependency>addAll(deps, _externalDependencies);
-    ExternalDependency _createXtextDependency = ExternalDependency.createXtextDependency("org.eclipse.xtext.web.servlet");
+    ExternalDependency _createXtextDependency = ExternalDependency.createXtextDependency("org.eclipse.xtext.xbase.web");
```
szarnekow: Should ExternalDependency have a meaningful equals/hashcode?

oehme: That would be helpful for testing. It is a mutable class though, that's why I left equals/hashCode out

szarnekow: Since we put them into a set to avoid duplicates, they have to have a proper equals semantics, don't they?

PR #634, opened 2015-09-28
Prioritze local classes over classes on the classpath

oehme:
To stay consistent with tools like javac, local classes should have
precedence over classes with the same name on the classpath.

See https://groups.google.com/forum/#!topic/xtend-lang/wBfRA64xxgo


oehme:
Jenkins test

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/batch/XtendBatchCompiler.java
```diff
@@ -733,6 +735,15 @@ public boolean apply(String input) {
 		// for annotation processing we need to have the compiler's classpath as a parent.
 		annotationProcessingClassLoader = createClassLoader(classpath, currentClassLoader);
 		resourceSet.eAdapters().add(new ProcessorInstanceForJvmTypeProvider.ProcessorClassloaderAdapter(annotationProcessingClassLoader));
+		removePossiblyShadowedTypeResources(resourceSet);
+	}
+
+	private void removePossiblyShadowedTypeResources(ResourceSet resourceSet) {
```
szarnekow: How can you be sure that no existing x-ref was resolved to one of these type resources?

oehme:
We decided to postpone this. It is pretty straightforward for the old batch compiler. But for the new builder infrastructure we would need some kind of "afterIndexing" hook that the JavaSourceLanguage could implement to throw TypeResources out of the ResourceSet. That might be something to consider for 2.10

PR #637, opened 2015-09-30
[IDEA/Xtend] Fixed broken template highlighting in IDEA.

svenefftinge:


plugins/org.eclipse.xtend.ide.common/src/org/eclipse/xtend/ide/common/highlighting/XtendHighlightingCalculator.java
```diff
@@ -384,6 +384,7 @@ public void acceptSemanticLineBreak(int charCount, RichStringLiteral origin, boo
 		@Override
 		public void acceptTemplateLineBreak(int charCount, RichStringLiteral origin) {
 			resetCurrentOffset(origin);
+			acceptor.addPosition(currentOffset, charCount, TEMPLATE_LINE_BREAK);
```
szarnekow: This breaks the Xtend highlighting tests and I think the linebreak will be highlighted twice in some cases, now (see acceptSemanticLineBreak)

svenefftinge: I don't think it will highlight twice, it's either one or the other case. 
Also 'acceptSemanticLineBreak' looks wrong to me, as it seem to mark semantic NLs as template NLs and POSSIBLE_LINE_BREAK for those that where controlStructureSeen is true (that method also doesn't look very accurate).

tests/org.eclipse.xtend.ide.tests/src/org/eclipse/xtend/ide/tests/highlighting/XtendHighlightingCalculatorTest.java
```diff
@@ -132,11 +140,13 @@ protected XtendClass clazz(String string) throws Exception {
 	}
 	
 	@Test public void testLiteral() {
+		strictMode = true;
 		expectInsignificant(0, 3);
 		expectInsignificant(3, 2);
+		expectTemplateLineBreak(5, 1);
 		expectInsignificant(6, 1);
+		expectTemplateLineBreak(14, 1); //TODO shouldn't this be POSSIBLE_LINE_BREAK
```
szarnekow: POTENTIAL_LINE_BREAK is used for lines that contain only expressions (no semantic ws). IF all expressions evaluate to null, there is no line emitted. TEMPLATE_LINE_BREAK seems about right to me here and for the other TODOs too.

plugins/org.eclipse.xtend.ide.common/src/org/eclipse/xtend/ide/common/highlighting/XtendHighlightingCalculator.java
szarnekow: I can't follow the last remark: which method do you refer to?

TEMPLATE_LINE_BREAK seems to refer to line breaks that will be emitted to the output thus a semantic LB is a template linebreak in the highlighting sense (rather unfortunate that the highlighting key has such gotten this name).
TEMPLATE_LINE_BREAKs are LB that are not visible in the output and thus are not rendered at all. Thinking about it, I wonder why they should get a dedicated highlighting if the do not appear in the output. My gut feeling right now is that the change in the highlighter is not correct - it depends on the semantics that we want to have, though. If we stick to: text that appears in the output will get the template coloring, than the call to acceptor.acceptPosition(..) is wrong.

svenefftinge: I was talking about 'acceptSemanticLineBreak' and 'announceSemanticLinebreak', which computes the 'controlStructureSeen' flag.

In IDEA the lexical highlighting is using IDEA's template background color for the whole template string. So we highlight the insignificant WS (paint it greyish again) rather than the significant (as we do in Eclipse). If don't paint the newline, the background color goes to the end of the editor.

It's probably better to do it as we do in eclipse in the first place but I still think this code is strange and I'd like to understand what template-, semantic- and possible line breaks are meant to be.

tests/org.eclipse.xtend.ide.tests/src/org/eclipse/xtend/ide/tests/highlighting/XtendHighlightingCalculatorTest.java
svenefftinge: I'm confused. In the comment above you said 
"TEMPLATE_LINE_BREAKs are LB that are not visible in the output and thus are not rendered at all. "
So the new line I'm talking about in this TODO is going into the output, no?

szarnekow: See org.eclipse.xtend.core.tests.richstring.AbstractRichStringEvaluationTest.testMultiLineTemplate():
This template emits 'foobar  \n' so [14,1] should be template line break, [5,1] is the wrong expectation, though.

plugins/org.eclipse.xtend.ide.common/src/org/eclipse/xtend/ide/common/highlighting/XtendHighlightingCalculator.java
szarnekow: I can only speak about the semantics of semantic_lb and template_lb as it is on the interface of the IRichStringPartAcceptor. Here, a semantic linebreak may occur in the output depending on the content in the line. A template_lb is never written to the output.
In the highlighter, the term template_lb appears to describe a linebreak that will definitely be written to the output. A potentail_lb is semantic but may be skipped if the line is empty.

szarnekow: Since we do the highlighting in IDEA the other way round as in eclipse, I don't see how we can reuse the entire implementation of the richstring highlighter.

svenefftinge: Well, we do. What's the problem?

tests/org.eclipse.xtend.ide.tests/src/org/eclipse/xtend/ide/tests/highlighting/XtendHighlightingCalculatorTest.java
svenefftinge: I'm confused. I guess it's better to discuss this offline. There must be some fundamental missunderstanding.

szarnekow: Ok

szarnekow:
Looks good, please merge.

PR #642, opened 2015-10-02
[idea] Lexer.getState is used to tell whether incremental lexing can be started at the current location.

svenefftinge:


intellij/org.eclipse.xtend.idea/src/org/eclipse/xtend/core/idea/highlighting/XtendHighlightingLexer.xtend
```diff
@@ -99,10 +100,7 @@ class XtendHighlightingLexer extends LexerBase {
 	}
 
 	override getState() {
-		if (currentRichTextToken != null)
-			currentRichTextToken.type
-		else
-			delegate.state
+		return if (_currentRichTextToken === null) 0 else 1
```
szarnekow: Do we need to override `start(.., initialState)`, too?

intellij/org.eclipse.xtext.idea/src/org/eclipse/xtext/idea/highlighting/SemanticHighlightVisitor.xtend
```diff
@@ -43,23 +44,24 @@ abstract class SemanticHighlightVisitor implements HighlightVisitor {
...
-			styles.forEach [
-				if(length > 0) {
-					val info = HighlightInfo
-						.newHighlightInfo(highlightInfoType)
-						.range(offset, offset + length)
-						.description(it)
-						.create
-					holder.add(info)			
+	    try {
+			acceptor = [
```
szarnekow: Cosmetics: Indentation looks inconsistent.

intellij/org.eclipse.xtend.idea/src/org/eclipse/xtend/core/idea/highlighting/XtendHighlightingLexer.xtend
svenefftinge: No, it's never called with something other than the base state (0) and we couldn't set the parser into that other state, as we don't know the rich string token.

PR #643, opened 2015-10-02
[xtext generator] Migration of the 'QuickOutlineFragment'

sailingKieler:
Migration of the 'QuickOutlineFragment'

### DIFF
--- a/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
+++ b/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
@@ -284,7 +284,7 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
 							}
 							
 							// outline API
-							fragment = ui.outline.OutlineTreeProviderFragment2 {}
+							fragment = ui.outline.OutlineTreeProviderFragment2 auto-inject {}
 							
 							// quick outline menu contribution
 							fragment = ui.outline.QuickOutlineFragment2 auto-inject {}
--- a/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
+++ b/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
@@ -286,9 +286,8 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
 							// outline API
 							fragment = ui.outline.OutlineTreeProviderFragment2 {}
 							
-							fragment = adapter.FragmentAdapter {
-								fragment = outline.QuickOutlineFragment auto-inject {}
-							}
+							// quick outline menu contribution
+							fragment = ui.outline.QuickOutlineFragment2 auto-inject {}
 							
 							// quickfix API
 							fragment = adapter.FragmentAdapter {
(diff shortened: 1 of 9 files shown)


plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
```diff
@@ -284,12 +284,10 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
 							}
 							
 							// outline API
-							fragment = adapter.FragmentAdapter {
-								fragment = outline.OutlineTreeProviderFragment auto-inject {}
-							}
-							fragment = adapter.FragmentAdapter {
-								fragment = outline.QuickOutlineFragment auto-inject {}
-							}
+							fragment = ui.outline.OutlineTreeProviderFragment2 {}
```
szarnekow: Should we use auto-inject here, too?

sailingKieler: Thanks for that hint ... got lost unintentionally. One more thing @spoenemann need to tell me about ;-)

sailingKieler: In the face of that: Why is that 'auto-inject' missing in 'GenerateXtend.mwe2'?

szarnekow: Currently we do not inject anything into that particular fragment, but if we create a project for clients, we want to to work as long and without surprises as possible. Thus if a client changes e.g. generateStubs on the level of the workflow, it should be applied to all fragments that have a property with that name. Thus auto-inject helps on new projects in general, but may be unnecessary for Xtend in particular.

sailingKieler: Got it (roughly)

plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
```diff
@@ -284,12 +284,10 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
...
+							
+							// quick outline menu contribution
```
oehme: Please also adjust the expectations in CLIWizardIntegrationTest. 

See https://xtext-builds.itemis.de/jenkins/job/xtext-runtime-tests-master-latest/640/#showFailuresLink

plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
sailingKieler: Re-added lost 'auto-inject'

plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
sailingKieler: What do I have to do?

oehme: Run the test, it will tell you which files differed from the expectation under testdata/wizard-expectations

oehme:
Jenkins test

sailingKieler:
expectations of ``CLIWizardIntegrationtest`` are up-to-date now. Please review merge!

PR #654, opened 2015-10-05
[old xtext generator] forward 'CodeConfig's flag 'preferXtendStubs' t…

sailingKieler:
…o instances of 'IStubGenerating.XtendOption'.

What do you think about that? Worked for me during testing. Only problem I saw yet:
The 'GenerateXbase' workflow implementation explicitly states
```
label.setGenerateXtendStub(true);
```
that gets overridden (the general preference is Java). Is that a serious blocker?


oehme:
I guess for xbase we could deactivate stub generation anyway. It's main use is to get new users an idea how to customize things.

sailingKieler:
Discussion gave that this would violate the compatibility, since the fragment-specific settings will be overwritten. Hence we drop this pull request.

PR #657, opened 2015-10-07
[xtext generator] Migration of LabelProviderFragment

sailingKieler:
Things I noticed:
* While the java package of the quickfixProvider stub is exported in manifest,
that of the labelProviders (and others) is not? Shall we consolidate that?
* If a grammar import another one/chain of grammars this hierarchy is reflected, e.g., in the validator or quickfixProvider classes; it is however not reflected in the labelProvider classes (and others), just the case of xbase is distinguished


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ui/labeling/LabelProviderFragment2.xtend
```diff
@@ -0,0 +1,235 @@
...
+	 * Extra getter facilitates customization by overriding. 
+	 */
+	def protected TypeReference getDefaultDescriptionLabelProviderSuperClass() {
+		new TypeReference("org.eclipse.xtext.ui.label.DefaultDescriptionLabelProvider")
+	}
+
+
+	override generate() {
+		if (!generateStub && !grammar.inheritsXbase) {
+			// the binding of the DefaultEObjectLabelProvider and the
```
oehme: Same argument against early exit here. It would be more readable in the positive condition

```
if (generateStub || inheritsXbase) {
  bindDescriptionLabelProvider
}

PR #658, opened 2015-10-07
[generator2] provide a simpler alternative to StandaloneSetup

oehme:


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGeneratorStandaloneSetup.xtend
```diff
@@ -0,0 +1,53 @@
...
+			delegate.addProjectMapping(new ProjectMapping => [
+				projectName = mapping.key
+				path = mapping.value.path
+			])
+		]
+	}
+
+	private def getProjectMappings() {
+		#[
+			projectConfig.runtimeProjectName -> projectConfig.runtimeRoot,
```
szarnekow: Maybe I don't get it, but how would a client extend this, e.g. to register more projects - let's say for netbeans integration?

oehme: I'll make those methods protected.

oehme: Also regarding 'more projects', we are thinking about giving more structure to XtextProjectConfig to make that easier. So then you could say something like project.allSubProjects.map[root]...

szarnekow:
StandaloneSetup#registerGeneratedEPackage was a nice and easy way to register a bunch of interdependent packages (they would be initialized transitively). How would I populate the registry with the simple API? Still fallback to the mwe StandaloneSetup I assume?

oehme:
Yes, the old StandaloneSetup will be around and we can add that method to the new one if there is demand. But I'd like to keep it simple for now.

PR #659, opened 2015-10-07
[serializer] re-implemented grammar constraint provider

meysholdt:
It's now based on the NFAs from SemanticSequencerNFAProvider.
Therefore, it will now benefit from all NFA/PDA processing
that's happening up the chain.


meysholdt:

szarnekow:
Sound's cool. Looking forward to it :)

plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/serializer/SemanticSequencerUtil.java
```diff
@@ -116,25 +117,24 @@ public Grammar getSuperGrammar(Grammar grammar) {
...
+				return null;
+			if (!feature.add(state.getFeature()))
+				return null;
+			if (state != nfa.getStart()) {
+				result.add(state);
+			}
+			state = state.getFollowers().get(0);
 		}
-		return false;
+		return null;
```
szarnekow: Is it important to distinguish between null and empty lists?

plugins/org.eclipse.xtext.util/src/org/eclipse/xtext/util/formallang/NfaToProduction.java
```diff
@@ -50,6 +51,12 @@ protected AbstractElementAlias(boolean optional, boolean many) {
 
 		protected abstract int getElementCount();
 
+		protected abstract void sort(Comparator<AbstractElementAlias<T>> comparator);
```
szarnekow: Should probably be Comparator<? super AbstractElementAlias<T>>

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/analysis/SemanticSequencerNfaProvider.java
```diff
@@ -63,6 +63,20 @@ public SemNfa(ISemState starts, ISemState stops) {
...
+			if (hashCode == -1)
+				hashCode = new NfaUtil().hashCode(this, GET_ASSIGNED_GRAMMAR_ELEMENT);
+			return hashCode;
+		}
+
+		@Override
+		public boolean equals(Object obj) {
+			if (obj == null || obj.getClass() != getClass())
+				return false;
+			return new NfaUtil().equalsIgnoreOrder(this, (SemNfa) obj, GET_ASSIGNED_GRAMMAR_ELEMENT);
```
szarnekow: Possible shortcut: obj == this

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/analysis/SemanticSequencerNfaProvider.java
```diff
@@ -63,6 +63,20 @@ public SemNfa(ISemState starts, ISemState stops) {
 		}
 
 		@Override
+		public int hashCode() {
+			if (hashCode == -1)
+				hashCode = new NfaUtil().hashCode(this, GET_ASSIGNED_GRAMMAR_ELEMENT);
```
szarnekow: It appears that the hashCode and equalsIgnoreCase are externalized but only used from here and used with the same keyfunction. Is that a temporary situation or would it make sense to inline the logic here and thereby reduce the complexity?

plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/serializer/SemanticSequencerUtil.java
meysholdt: yes. When null is returned, the SerializerFragment generates and invocation of the BacktrackingSemanticSequencers. When an empty list is returned, it generates code that consumes the values of zero features.

plugins/org.eclipse.xtext.util/src/org/eclipse/xtext/util/formallang/NfaToProduction.java
meysholdt: true, will fix.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/analysis/SemanticSequencerNfaProvider.java
meysholdt: good catch, will fix.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/analysis/SemanticSequencerNfaProvider.java
meysholdt: I like to keep them in NfaUtil because 
org.eclipse.xtext.util.formallang.Nfa<STATE>
org.eclipse.xtext.util.formallang.Pda<STATE, STACKITEM>
are two very lightweight, generic interfaces with may use-case-specific implementations. To allow this, the generic, reusable code goes into 
org.eclipse.xtext.util.formallang.NfaUtil
org.eclipse.xtext.util.formallang.PdaUtil

Separating use-case-specific-code from generic code to handle NFAs/PDAs has proven to be useful in the past to reduce complexity and enable reuse.

meysholdt:

plugins/org.eclipse.xtext.util/src/org/eclipse/xtext/util/formallang/NfaUtil.java
```diff
@@ -321,13 +318,22 @@ public Object apply(S from) {
 		});
 	}
 
-	public <S> boolean equalsIgnoreOrder(Nfa<S> nfa1, Nfa<S> nfa2, Function<S, ? super Object> keyFunc) {
+	public <S> int hashCode(Nfa<S> nfa, Function<S, ? extends Object> keyFunc) {
+		int result = 1;
+		for (S element : collect(nfa)) {
+			Object s = keyFunc.apply(element);
+			result = 31 * result + (s == null ? 0 : s.hashCode());
```
szarnekow: This hashCode computation is not symmetric to equalsIgnoreOrder, e.g. two equal NFAs will produce different hashCodes except if collect(..) is somehow used in equalsIgnoreOrder, too.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/analysis/SemanticSequencerNfaProvider.java
szarnekow: Makes sense!

plugins/org.eclipse.xtext.util/src/org/eclipse/xtext/util/formallang/NfaUtil.java
szarnekow: Could be fixed by xoring the hashcodes of all the nfas rather than multiplying with a prime.

szarnekow: Or just use the logic from java.util.AbstractSet.hashCode()

meysholdt: good catch. fixed in latest patch set. I also noticed that the hasCode only represented the existence of states but ignored the existence of edged and their direction. fixed as well.

meysholdt:

meysholdt:

PR #660, opened 2015-10-07
Fixed two issue with parser rule fragments

szarnekow:


szarnekow:

PR #661, opened 2015-10-08
[generator2] make XtextProjectConfig more structured

oehme:


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/WizardConfig.xtend
```diff
@@ -44,151 +32,82 @@ class WizardConfig extends XtextProjectConfig {
...
+			if (metaInf === null)
+				metaInf = computeMetaInf
+			if (src === null)
+				src = computeSrc
+			if (srcGen === null)
+				srcGen = computeSrcGen
+			if (it instanceof BundleProjectConfig) {
+				if (createEclipseMetaData) {
+					if (manifest === null)
+						manifest = new ManifestAccess
```
szarnekow: I prefer the variant that used the injector to create the manifestAccess since subclasses could be bound. This approach only allows to inject other properties into the *accesses but doesn't provide the advantage of inheritance / polymorphism. Is there a specific reason to do it like this that I don't understand yet?

oehme: The old approach did not injectMembers on Manifests that the users set manually. Also, it is more consistent with how we create the FileSystemAccess. 

We could of course add a combination of both. In that case we should also add a FileSystemAccessFactory so the user can customize that too.

oehme: ...or just add overridable methods, since the ProjectConfig is basically the factory for these types, they are created nowhere else.

szarnekow: Sounds about right to me.

oehme:
Jenkins test

oehme:
Jenkins test

PR #667, opened 2015-10-09
The AbstractLexerBasedConverter might have been used with parser rules in the past. We need to be more forgiving with those cases.

svenefftinge:


szarnekow:
:+1:

PR #670, opened 2015-10-12
[xtext generator] Migration of 'RefactorElementNameFragment'

sailingKieler:


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/ui/refactoring/RefactorElementNameFragment2.xtend
```diff
@@ -0,0 +1,149 @@
...
+class RefactorElementNameFragment2 extends AbstractGeneratorFragment2 {
+
+	@Inject
+	extension XtextGeneratorNaming
+
+	@Inject
+	extension XbaseUsageDetector
+
+	@Accessors
+	private Boolean useJdtRefactoring = null; // no default (depends on whether xbase is used or not)
```
szarnekow: The comment could be JavaDoc and explicitly state that it defaults to 'false' if Xbase is not used and otherwise to true.

PR #671, opened 2015-10-12
[xtext generator] added removal of duplicate imports in 'JavaFileAccess', …

sailingKieler:
and resolved code clone in 'GeneratedJavaFileAccess'.

``getContent()`` in ``JavaFileAccess`` as well as in ``GeneratedJavaFileAccess`` did not check for duplicate class names in ``imports.values``.
Added that filtering in ``JavaFileAccess`` und simplified implementation of ``getContent``.
Resolved the modified clone of ``getContent()`` in ``GeneratedJavaFileAccess``. 

### DIFF
--- a/plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/GeneratedJavaFileAccess.xtend
+++ b/plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/GeneratedJavaFileAccess.xtend
@@ -4,8 +4,6 @@ import java.util.List
 import org.eclipse.xtend.lib.annotations.Accessors
 import org.eclipse.xtext.xtext.generator.CodeConfig
 import org.eclipse.xtext.xtext.generator.model.annotations.IClassAnnotation
-import java.util.Collections
-import com.google.common.collect.Lists
 
 class GeneratedJavaFileAccess extends JavaFileAccess {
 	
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/JavaFileAccess.xtend
```diff
@@ -105,19 +103,20 @@ class JavaFileAccess extends TextFileAccess {
...
-				import �importName��IF appendSemicolons�;�ENDIF�
-			�ENDFOR�
-			
-			�internalContents�
-		'''
+	override getContent() '''
+		�codeConfig.fileHeader�
+		package �javaType.packageName��IF appendSemicolons�;�ENDIF�
+
+���		remove duplicates in required imports and sort them
```
oehme: I think this is pretty clear from the line below. But if you think it is worth pointing out, please extract a method instead of adding a comment.

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/GeneratedJavaFileAccess.xtend
```diff
@@ -28,24 +26,26 @@ class GeneratedJavaFileAccess extends JavaFileAccess {
...
+			importType(annotationImport)
+		]
+		super.getContent()
+	}
+	
+	private def getClassAnnotations() {
+		annotations + codeConfig.classAnnotations.filter[appliesTo(this)]
+	}
+	
+	override getInternalContent() '''
```
szarnekow: Template methods like this are usually called doXYZ, e.g doGetContents in the Xtext codebase. We also sometimes use internalXYZ, e.g. internalGetContents but generally it's not getInternalXYZ.

sailingKieler: ok. actually I just chose a name as close as possible to the variable name and fairly in line with 'getContent'.
Would you prefer `ìnternalGetContent`` here?

szarnekow: In fact I'd prefer a name that describes the result of the method, e.g. getTopLevelClass

PR #675, opened 2015-10-12
Conditionaly split switch in specialStateTransition method

jsebechlebsky:
Switch in lexer in specialStateTransition can get really big, even after pulling code from case statements into separate methods.

Somewhere slightly over 1500 cases the specialStateTransition method can exceed 64kB limit. 

Implemented transformation will split such a method into several methods with defined maximal number of case statements (1000 by default).

Signed-off-by: Jan Sebechlebsky <jan.sebechlebsky@cz.ibm.com>

### DIFF
--- a/tests/org.eclipse.xtext.tests/src/org/eclipse/xtext/generator/parser/splitting/LexerSpecialStateTransitionSplitterTest.java
+++ b/tests/org.eclipse.xtext.tests/src/org/eclipse/xtext/generator/parser/splitting/LexerSpecialStateTransitionSplitterTest.java
@@ -63,6 +63,43 @@ public class LexerSpecialStateTransitionSplitterTest extends Assert {
 			"\n" + 
 			"                        if ( s>=0 ) return s;\n" + 
 			"                        break;\n" +
+			"                    case 2 : \n" +
+			"                        int LA12_0 = input.LA(1);\n" +
+			"\n" +
+			"                        s = -1;" +
+			"                        if ( (LA12_0=='+'||LA12_0=='-') && (( currentState() == State.STATE_A ))) {s = 125;}\n" +
+			"\n" +
+			"                        else if ( ((LA12_0>='0' && LA12_0<='9')) && (( currentState() == State.STATE_A ))) {s = 126;}\n" +
+			"\n" +
+			"                        else if ( (LA12_0=='.') && (( currentState() == State.STATE_A ))) {s = 127;}\n" +
+			"\n" +
+			"\n" +
+			"                        if ( s>=0 ) return s;\n" +
+			"                        break;\n" +
+			"                    case 3 : \n" +
+			"                        int LA12_2 = input.LA(1);\n" +
+			"\n" +
+			"                        s = -1;\n" +
+			"                        if ( (LA12_2=='.') && (( currentState() == State.STATE_A ))) {s = 128;}\n" +
+			"\n" +
+			"                        else if ( ((LA12_2>='0' && LA12_2<='9')) && (( currentState() == State.STATE_A ))) {s = 126;}\n" +
+			"\n" +
+			"                        else if ( (LA12_2=='E'||LA12_2=='e') && (( currentState() == State.STATE_A ))) {s = 128;}\n" +
+			"\n" +
+			"                        if ( s>=0 ) return s;\n" +
+			"                        break;\n" +
+			"                    case 4 : \n" +
+			"                        int LA12_1 = input.LA(1);\n" +
+			"\n" +
+			"                        s = -1;\n" +
+			"                        if ( ((LA12_1>='0' && LA12_1<='9')) && (( currentState() == State.STATE_A ))) {s = 126;}\n" +
+			"\n" +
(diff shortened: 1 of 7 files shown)


szarnekow:
Thanks for the PR. In the parser we observe quite often that many many cases have the same body block in those bigger switch statements. Is that also true for the lexer?
Could you please provide a test language grammar that shows the effect of the splitting?

jsebechlebsky:
Hi Sebastian. 
I haven't searched for identical bodies in lexer switch statements, I'll try to check it - if that's true, maybe it would a be good place to make some more optimizations. 

I've modified unit test for LexerSpecialStateTransitionSplitter to cover also new functionality. If you take a look at the file there is a result after transformation (Is this sufficient as an example of transformation?). 

While writing the test I realized I probably  misused ignoreCaseCountGuard so I removed check for that. I suppose the purpose of the this variable was just to force LexerSpecialStateTransitionSplitter to pull code from cases into methods regardless of number of cases during the test. 

I've squashed changes on LexerSpecialStateTransitionSplitter into one commit, the test is in the separate commit. If you would like to squash it all into single commit just let me know.

tests/org.eclipse.xtext.tests/src/org/eclipse/xtext/generator/parser/splitting/LexerSpecialStateTransitionSplitterTest.java
```diff
@@ -63,6 +63,43 @@
...
+			"                        if ( (LA12_2=='.') && (( currentState() == State.STATE_A ))) {s = 128;}\n" +
+			"\n" +
+			"                        else if ( ((LA12_2>='0' && LA12_2<='9')) && (( currentState() == State.STATE_A ))) {s = 126;}\n" +
+			"\n" +
+			"                        else if ( (LA12_2=='E'||LA12_2=='e') && (( currentState() == State.STATE_A ))) {s = 128;}\n" +
+			"\n" +
+			"                        if ( s>=0 ) return s;\n" +
+			"                        break;\n" +
+			"                    case 4 : \n" +
+			"                        int LA12_1 = input.LA(1);\n" +
```
szarnekow: case 4 appears to have the same body as case 2

szarnekow:
If your test data was taken from a real lexer, it seems to be the same situtation as for the parser. The case block are identical for different case labels. Do you have the time to work on a mechanism that allows to detect those dups and merge them into a switch with fall-throughs for duplicate case blocks?

jsebechlebsky:
You're right, 2 and 4 are semantically identical in this case. I was little worried if the Antlr could generate conditions which would not exclude each other -> in that case the order of the condition testing would matter. But that could be probably easily checked by inspecting Antlr code, or if you are sure about that, I think I can implement the transformation and submit another pull request.

szarnekow:
I'm very confident that the order in the if-else checks within the cases doesn't matter. The if-else cascades represent a normalized (sub-)DFA as far as I can tell thus each if-condition describes the conditions for the transition to another state. These transitions cannot overlap each other.
It would be greatly appreciated if you could come up with a logic that determines the identical cases and merges them. Please feel free to add the commits to this PR.

szarnekow:
Jan, are you investigating in the discussed optimization?

jsebechlebsky:
Hi Sebastian, sorry for a delay, I'm planning to do it on monday.

jsebechlebsky:
Hi Sebastian, I haven't finished it yet but so far it seems that on my lexer sample almost 3/4 cases contains identical statements.

szarnekow:
That sounds very promising :+1:  Do you plan to support that automated refactoring / code cleanup also for the parser?

jsebechlebsky:
Hi Sebastian, It will probably take me some more time to look at it (I probably also need separate permission for this). I would like to ask you - When do you think code from this pull request can be included in the release? We're using automatically generated xtext parsers with custom lexer using semantic predicates which result in non-compilable code in certain cases, I've noticed that quite a lot people using custom lexers are facing similar issue. It would be nice to have this fixed in early Xtext release.

svenefftinge:
hi jan, we are going to have a release in the coming days. Tomorrow there will be an RC if there are no stoppers in it we are going to release it next Monday. 

The change looks good, but could potentially introduce problems and I would prefer to have the splitting disabled by default. So if you could provide a PR that completely skips the newly introduced code unless someone sets the new skipping property, I can imagine to have it in 2.9.1, which will be in two or three weeks. Does that sound ok for you?

jsebechlebsky:
That sounds great, I'll make the modification and push it to this pull request.

jsebechlebsky:
Hi, I've pushed the change - it will turn off special state switch splitting by default (new transformation is completely omitted), it can be explicitly turned on by setting specialStateSwitchSplitting to true.

svenefftinge:
I'd prefer a single property that encodes the number and whether it is set or not. 
Because now people could set the number but it wouldn't have any effect if they forgot to enable the splitting, which is unnecessarily complicated. I think either Integer with null or int with a negative value would be a good value for a disabled switch-splitting.

svenefftinge:
I merged and did another PR #797

jsebechlebsky:
Thanks a lot!

PR #677, opened 2015-10-13
[xtext generator] upgraded 'GeneratedJavaFileAccess' to expect the 'typeComment' as a 'StringConcatenationClient'…

sailingKieler:
… and process imports of referenced classes. The generated Java files affected by this change are amended.

### DIFF
--- a/plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/GeneratedJavaFileAccess.xtend
+++ b/plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/GeneratedJavaFileAccess.xtend
@@ -2,12 +2,13 @@ package org.eclipse.xtext.xtext.generator.model
 
 import java.util.List
 import org.eclipse.xtend.lib.annotations.Accessors
+import org.eclipse.xtend2.lib.StringConcatenationClient
 import org.eclipse.xtext.xtext.generator.CodeConfig
+import org.eclipse.xtext.xtext.generator.model.JavaFileAccess.JavaTypeAwareStringConcatenation
 import org.eclipse.xtext.xtext.generator.model.annotations.IClassAnnotation
 
 class GeneratedJavaFileAccess extends JavaFileAccess {
 	
-	@Accessors
 	CharSequence typeComment
 	
 	@Accessors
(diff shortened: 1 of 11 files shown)


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/GeneratedJavaFileAccess.xtend
```diff
@@ -25,9 +26,17 @@ class GeneratedJavaFileAccess extends JavaFileAccess {
 		if (!markedAsGenerated)
 			throw new IllegalArgumentException("It's always generated");
 	}
-
+	
+	def setTypeComment(StringConcatenationClient javaContent) {
```
sailingKieler: A question on the return type that is now 'CharSequence'. Is that fine or should I explicitly set it to 'void'?
(did it the same way like in 'JavaFileAccess.setContent(...)')

szarnekow: I think setters should return void in general. Also in JavaFileAccess. Is the return value of JavaFileAccess.content used somewhere?

sailingKieler: the setter stem's from 'TextFileAccess' and declaring it explicitly 'void' does not break anything.
Personally I'm not a fan of void, but rather would return the '...FileAccess' instance itself, like it is done in ``GuiceModuleAccess.BindingFactory``. But that may be a bit overdressed here.
Anyway, returning the 'CharSequence' doesn't make sense to me.

szarnekow: Please let the setters return 'void'.

sailingKieler: done for both 'setContent(...)' and 'setTypeComment(...)'.

szarnekow:
Jenkins test

PR #682, opened 2015-10-14
Update 103_domainmodelnextsteps.md

eavy:
There is a bug in the example.

First when you use `getSuperType()` then `superEntity` is always null.
Second there is always at least one feature with such name, so you should skip at least _once_.

May be it's not the most beautiful solution, but it works :)

### DIFF
--- a/xtext-website/documentation/103_domainmodelnextsteps.md
+++ b/xtext-website/documentation/103_domainmodelnextsteps.md
@@ -355,12 +355,18 @@ The second validation rule is straight-forward, too. We traverse the inheritance
 ```xtend
 @Check
 def void checkFeatureNameIsUnique(Feature f) {
-	var superEntity = (f.eContainer() as Entity).getSuperType();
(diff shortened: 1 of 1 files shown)


svenefftinge:
Closing this as to huibo's comment. Please reopen if we've missed something.

eavy:
@shihuibo19 you're right. I was not so attantive during reading. And I was confused by the name of the function `checkFeatureNameIsUnique`. I thought it checks the uniqueness of the names on the same level.
> The second validation rule is straight-forward, too. We traverse the inheritance hierarchy of the Entity and look for features with equal names.

But after reading this text again that function makes sense, but may be someone should rename it so to show that it only checks uniqueness only between features of the current entiry and the features of the inherited entity.
```
entity HasAuthor {
    author: String
  }
entity Post extends HasAuthor {
    title: String
    author: String /* Will show an error */
    title: String /* It won't show an error */
    content: String
    many comments: Comment
  }
```

If I'm missing anything again, please, tell me.

PR #686, opened 2015-10-14
[generator] Simpler language generator configuration for the standard case

svenefftinge:


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/StandardLanguage.xtend
```diff
@@ -0,0 +1,152 @@
...
+			standaloneSetup = new XtextLanguageStandaloneSetup => [
+				addLoadedResource("platform:/resource/org.eclipse.xtext.xbase/model/Xbase.genmodel")
+			]
+		} catch (ClassNotFoundException e) {
+			LOG.info("Skipping registration of Xbase genmodel. Xbase is not on the classpath.")
+		}
+	}
+	
+	override initialize(Injector injector) {
+		val configuredFragments = new ArrayList(fragments)
```
szarnekow: a template method 'getStandardFragments' would be easier to override

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/StandardLanguage.xtend
```diff
@@ -0,0 +1,152 @@
...
+import org.eclipse.xtext.xtext.generator.validation.ValidatorFragment2
+import org.eclipse.xtext.xtext.generator.web.WebIntegrationFragment
+import org.eclipse.xtext.xtext.generator.xbase.XbaseGeneratorFragment2
+import org.eclipse.xtext.xtext.generator.xbase.XtypeGeneratorFragment2
+import org.eclipse.xtext.util.internal.Log
+
+/**
+ * @author Sven Efftinge - Initial contribution and API
+ */
+@Accessors(PUBLIC_SETTER)
```
szarnekow: Protected getters would be nice, too - in case a client wants to have MyConfig extends StandardConfig

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/types/TypesGeneratorFragment2.xtend
```diff
@@ -12,10 +12,20 @@ import org.eclipse.xtext.xtext.generator.AbstractGeneratorFragment2
...
 import static extension org.eclipse.xtext.xtext.generator.model.TypeReference.*
+import org.eclipse.xtext.xtext.generator.xbase.XbaseUsageDetector
+import com.google.inject.Inject
+import org.eclipse.xtend.lib.annotations.Accessors
 
 class TypesGeneratorFragment2 extends AbstractGeneratorFragment2 {
 	
+	@Inject XbaseUsageDetector xbaseUsageDetector
+	
+	@Accessors boolean onlyEnabledIfGrammarIsUsed = false
```
szarnekow: should this property use the term 'Xbase' or 'Xtype' ?

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/StandardLanguage.xtend
oehme: That would also make sense in the superclass LanguageConfig2, which adds the ImplicitFragment.

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/StandardLanguage.xtend
```diff
@@ -0,0 +1,152 @@
...
+				addLoadedResource("platform:/resource/org.eclipse.xtext.xbase/model/Xbase.genmodel")
+			]
+		} catch (ClassNotFoundException e) {
+			LOG.info("Skipping registration of Xbase genmodel. Xbase is not on the classpath.")
+		}
+	}
+	
+	override initialize(Injector injector) {
+		val configuredFragments = new ArrayList(fragments)
+		fragments.clear
```
oehme: This removes all explicitly added Fragments (like ecore2Xtext)

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/StandardLanguage.xtend
svenefftinge: The getters are intentionally left out, so I get compiler warnings for non referenced fields :-)
I don't think this class should be designed for extendability, the super class is.

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/types/TypesGeneratorFragment2.xtend
svenefftinge: Yes, there are more accurate solutions. I went for a quick and 'not' dirty one :-)
Let's go with this and improve later.

plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/StandardLanguage.xtend
svenefftinge: and adds the later again. 
But I agree it's not nice. I could alternatively use fragments.add(i++, bla) starting with null, which I also didn't quite like.

szarnekow:
When we discussed the design, one scenario for clients was to copy-the-whole-thing. I though inheritance would be nice, too thus adding protected getters and a single template method would greatly improve the reusability.

PR #688, opened 2015-10-14
[serializer] introduce IContext

meysholdt:
Previously, the context was of type EObject which actually was a
ParserRule or assigned Action. 

This refactoring paves the road for more advanced contexts, such
as ParserRules with parameter values.


meysholdt:
Jenkins 	xtext-runtime-tests-master-latest

meysholdt:

meysholdt:

meysholdt:

meysholdt:

meysholdt:

meysholdt:

svenefftinge:
Do we really need an interface? Isn't SerializationContext good enough? It also has a better name (notes that it is used in serialization).

Also I saw lots of commented out code. Can that be deleted?

meysholdt:
Using an interface allows us to have public methods *without* making them API (on the class).

szarnekow:
Are there any none-static public methods in the SerializationContext that are not on the interface?
Another thing that I wonder about: The IContext appears in generated code so it should probably be in an exported package? If we mark the serializer itself as API, we don't want to expose all the support packages, too, do we?

tests/org.eclipse.xtext.tests/src/org/eclipse/xtext/serializer/GrammarConstraintProviderTest.java
```diff
@@ -8,20 +8,21 @@
...
 import org.eclipse.xtext.Grammar;
 import org.eclipse.xtext.XtextStandaloneSetup;
 import org.eclipse.xtext.junit4.AbstractXtextTests;
+import org.eclipse.xtext.serializer.analysis.IContext;
 import org.eclipse.xtext.serializer.analysis.IGrammarConstraintProvider;
 import org.eclipse.xtext.serializer.analysis.IGrammarConstraintProvider.IConstraint;
-import org.eclipse.xtext.serializer.analysis.IGrammarConstraintProvider.IConstraintContext;
+import org.eclipse.xtext.serializer.analysis.SerializationContext;
+import org.eclipse.xtext.util.Pair;
 import org.junit.Ignore;
```
szarnekow: Maybe the `@Ignored` tests will now work, too?

meysholdt:
yes, on the long them the context interface should be in an exported package. But making the serializer public API is beyond this pull request. This requires some more rearrangement to separate between API we can keep stable and support classes.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/sequencer/ContextFinder.java
```diff
@@ -91,160 +92,188 @@
...
+			constraints = grammarConstraintProvider.getConstraints(ruleNames.getContextGrammar());
 		}
 	}
 
-	protected boolean intersect(Iterable<EObject> it1, Iterable<EObject> it2) {
-		for (EObject i1 : it1)
-			for (EObject i2 : it2)
-				if (i1 == i2)
+	protected boolean intersect(Iterable<IContext> it1, Iterable<IContext> it2) {
+		for (IContext i1 : it1)
```
szarnekow: I don't know about the avg number of elements in the iterables, but it could be more efficient to use CollectionExtensions.toSet(it1) and than a single loop over second one with set.contains(elem)

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/sequencer/ContextFinder.java
```diff
@@ -50,27 +53,25 @@
...
 	@Inject
 	protected ITransientValueService transientValues;
 
 	@Inject
-	protected ISemanticNodeProvider nodesProvider;
+	protected TransientValueUtil transientValueUtil;
 
-	protected Iterable<AbstractElement> findAssignedElements(EObject obj, EStructuralFeature feature,
-			Iterable<AbstractElement> candidates) {
+	protected Set<AbstractElement> findAssignedElements(EObject obj, EStructuralFeature feature,
```
szarnekow: Should this API require ordering guarantees, e.g. by using LinkedHashMultimap as input?

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/impl/Serializer.java
```diff
@@ -151,6 +173,13 @@ protected EObject getContext(EObject semanticObject) {
 		return contexts.next();
 	}
 
+	protected IContext getIContext(EObject semanticObject) {
```
szarnekow: I don't like the method name though I see that `getContext` was already taken. `getSerializationContext`?

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/diagnostic/SerializationDiagnostic.java
```diff
@@ -24,46 +25,54 @@
...
-	public SerializationDiagnostic(String id, EObject semanticObject, EObject context, Grammar grammar, String message) {
+	/**
+	 * @deprecated use {@link #SerializationDiagnostic(String, EObject, IContext, Grammar, String)}.
+	 */
+	@Deprecated
+	public SerializationDiagnostic(String id, EObject sem, EObject context, Grammar grammar, String message) {
+
+	}
+
+	public SerializationDiagnostic(String id, EObject sem, IContext context, Grammar grammar, String message) {
```
szarnekow: What was wrong with `semanticObject`? :)

plugins/org.eclipse.xtext/src/org/eclipse/xtext/serializer/analysis/SyntacticSequencerPDAProvider.java
```diff
@@ -606,17 +590,25 @@ protected SynTransition createTransition(SynAbsorberState source, SynAbsorberSta
...
+			EClass type = contexts.get(0).getType();
 			Map<ISerState, SynAbsorberState> absorbers = Maps.newHashMap();
 			Map<SynAbsorberState, Map<ISerState, SynState>> emitters = Maps.newHashMap();
-			Grammar grammar = GrammarUtil.getGrammar(context);
-			Pda<? extends ISerState, RuleCall> pda = pdaProvider.getContextTypePDA(grammar, context, type);
-			result = createAbsorberState(pda.getStart(), absorbers, emitters, context, type);
-			cache.put(key, result);
+			SynAbsorberState state = createAbsorberState(pda.getStart(), absorbers, emitters, type);
+			for (IContext ctx : contexts) {
+				result.put(ctx, state);
```
szarnekow: Should we assert that there is no state yet in that map for the given ctx?

szarnekow:
I understand that we don't want to make it public API now, but given that we want to do that eventually, we should go for a single package for all types that are used from generated code.

meysholdt:

szarnekow:
To fix the tests, you may want to add a boolean flag to the FlattenedGrammarAccess, whether the typeRef should be copied or not. In general I'd prefer a solution that is symmetric to the other clients, which would request the originalElement before accessing the tyoe, though.

meysholdt:

meysholdt:
it did. let's hope this was the last obstacle.

meysholdt:

meysholdt:
Jenkins test

PR #691, opened 2015-10-15
[generator] Delegating from old to new serializer fragment

spoenemann:
* Implemented adapter for using the new generator in old fragments
* Removed implementation of old serializer fragment


meysholdt:
thank you! :-) This saves my butt in #688.

PR #692, opened 2015-10-15
[479885] Fixed regression with clean builds

szarnekow:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=479885

oehme:
Can't we use the ProjectConfig to find the project name for the resource? That would make the additional "projectName" argument unnecessary.

szarnekow:
Currently there is no project config attached to the resource set that is used by the builder (as far as I can tell)

oehme:
Ah, you're right, we can only get hold of the WorkspaceConfig, but not of the ProjectConfig so far. Sven wanted to fix that API for 2.9 (get rid of WorkspaceConfig, install ProjectConfig everywhere). I think this is another sign that that's a good idea.

szarnekow:
I'll merge this despite the potential refactoring to have a build that I can test tomorrow.

PR #693, opened 2015-10-16
Regression: workspace root is not always bound

szarnekow:


svenefftinge:
Why or when is it not bound?

PR #696, opened 2015-10-16
[scoping fragment2] scope providers extends DelegatingScope, generate scope provider stub also for Xbase languages.

svenefftinge:


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/StandardLanguage.xtend
```diff
@@ -95,6 +88,14 @@ import org.eclipse.xtext.xtext.generator.xbase.XtypeGeneratorFragment2
 	 
 	XtextAntlrIDEAGeneratorFragment ideaParser = new XtextAntlrIDEAGeneratorFragment
 	
+	TypesGeneratorFragment2 commonTypesSupport = new TypesGeneratorFragment2 => [
+		onlyEnabledIfGrammarIsUsed = true
+	]
+	
+	XtypeGeneratorFragment2 xtypeSupport = new XtypeGeneratorFragment2
+	
+	XbaseGeneratorFragment2 xbaseSupport = new XbaseGeneratorFragment2
```
szarnekow: Reordered fragment will change bindings in generated code, e..g for rename refactoring if xbase is used.

szarnekow: Oh well they won't - got it :)

PR #697, opened 2015-10-16
[idea] fix EMF URI for IDEA

akosyakov:


akosyakov:

PR #702, opened 2015-10-19
[xtext generator] Migrated 'Ecore2XtextValueConverterServiceFragment'

sailingKieler:


svenefftinge:
As this just registers one binding I think this should be done in the runtime module instead in the future.

szarnekow:
Could you please elorate a bit on that? You want to remove the fragment and somehow communicate to clients that the need to add a binding manually instead?

sailingKieler:
rebased, ready for merge. Have set the package ...xtext.xtext.generator.ecore2xtext x-internal, too.

svenefftinge:
I think it could be a bit less loud, e.g. having the fragment configured always and then check whether the binding needs to be done, or even put that converter rule into the base class. But it's not super important, so let's leave it as is.

PR #704, opened 2015-10-19
[480118] Fixed regression in new grammar access fragment

szarnekow:
Not all valid java identifiers are valid in Antlr
action contexts.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=480118

szarnekow:
Probably an arbitrary decision back in the days :)

PR #706, opened 2015-10-20
[builder] IShouldGenerate uses ProjectConfig

svenefftinge:


plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/EclipseBasedShouldGenerate.xtend
```diff
@@ -14,18 +14,20 @@ import org.eclipse.core.runtime.Path
...
 class EclipseBasedShouldGenerate implements IShouldGenerate {
 
-	override shouldGenerate(Resource resource, String projectName, CancelIndicator cancelIndicator) {
+	override shouldGenerate(Resource resource, CancelIndicator cancelIndicator) {
 		val uri = resource.URI
 		if (uri === null || !uri.isPlatformResource)
 			return false
 		val member = ResourcesPlugin.workspace.root.findMember(new Path(uri.toPlatformString(true)))
-		if (member !== null && member.type === IResource.FILE && member.project.name == projectName) {
+		val projectConfig = ProjectConfigAdapter.findInEmfObject(resource.resourceSet)?.projectConfig
```
szarnekow: Minor optimization: Can we postpone the adapter lookup until we are sure that we are processing a resource that was loaded from an IFile, e.g. lookup the projectConfig after member !== null && member.type === FILE?

PR #710, opened 2015-10-20
Renamings and API shrinking for new Generator

oehme:


plugins/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/adapter/FragmentAdapter.xtend
```diff
@@ -289,7 +289,7 @@ class FragmentAdapter extends AbstractGeneratorFragment2 {
 	}
 	
 	protected def LanguageConfig createLanguageConfig() {
-		val config2 = language as LanguageConfig2
+		val config2 = language as XtextGeneratorLanguage
```
szarnekow: cosmetic: we should rename local vars, too, if types are renamed

oehme: I'll do these cosmetics in another round.

PR #720, opened 2015-10-22
Bug 474868 xtext-maven-plugin should ignore target/classes

oehme:
This is the same as the xtend-maven-plugin already.

When models are copied to target/classes, they used to be on the
classpath of the next generator run. This may lead to hard-to-diagnose
errors like "Foo cannot be cast to Foo".


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=474868

PR #723, opened 2015-10-22
IGenerator2 improvements

oehme:


plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/JavaProjectBasedBuilderParticipant.java
```diff
@@ -136,7 +137,7 @@ protected void handleChangedContents(Delta delta, IBuildContext context, IFileSy
 			return;
 		Resource resource = context.getResourceSet().getResource(delta.getUri(), true);
 		if (shouldGenerate(resource, context)) {
-			generator.generate(resource, (IFileSystemAccess2) fileSystemAccess);
+			generator.generate(resource, (IFileSystemAccess2) fileSystemAccess, CancelIndicator.NullImpl);
```
szarnekow: We should try to obtain a valid cancelIndicator from the fileSystemAccess

plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/IGenerator2.java
```diff
@@ -32,24 +33,19 @@
 	 * are non-blocking if the given fsa is implementing that pattern.
 	 * Read operations will happen in the background but
 	 * are blocking.
-	 *
```
szarnekow: My precious `@param` docs ;)

plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/IGenerator2.java
```diff
@@ -32,24 +33,19 @@
...
 	 * is executed.
 	 */
-	void beforeGenerate(Resource input, IFileSystemAccess2 fsa);
+	void beforeGenerate(Resource input, IFileSystemAccess2 fsa, CancelIndicator cancelIndicator);
 
 	/**
 	 * Release any data that is no longer necessary after the generator ran. 
 	 */
-	void afterGenerate(Resource input, IFileSystemAccess2 fsa);
+	void afterGenerate(Resource input, IFileSystemAccess2 fsa, CancelIndicator cancelIndicator);
```
szarnekow: We should document the contract of the life cycle hooks in case of cancellation. E.g. afterGenerate is always called independently from the cancellation state of the generator or smth like that.

oehme:
I'll document the contract of afterGenerate in the next PR where I make sure that the contract actually holds ;)

PR #724, opened 2015-10-22
[NodeModel] replace a first child of a new node with null if an old does not have children

akosyakov:


akosyakov:

szarnekow:

szarnekow:

PR #734, opened 2015-10-22
Fix freeze on FailedPredicateException

jsebechlebsky:
I believe input should be consumed also when FailedPredicateException is thrown from the lexer. If not the while may keep spinning infinitely.


szarnekow:
Jenkins test

szarnekow:

PR #742, opened 2015-10-23
[idea] keep order of events

svenefftinge:
Previously when canceled, the no processed events were added to the tail of the queue. As a result the order got inverted (newer events were preceding older ones). This could potentially have a bad effect when you have add and delete events for the same URI.

This change keeps track of the order and also ensures that deleted URIs don’t show up in the changed set and vice versa.


svenefftinge:
Jenkins test

svenefftinge:

PR #765, opened 2015-10-29
[gradle] make all idea plugins exclude ‘org.eclipse.xtend.lib’ because they would otherwise be on the classpath of the annotation processors classloader and be prioritized over the ones on the module classpath.

svenefftinge:


svenefftinge:

PR #766, opened 2015-10-29
fixed IShouldGenerate.OnlyWithoutErrors

oehme:


plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/IShouldGenerate.xtend
```diff
@@ -35,7 +36,9 @@ interface IShouldGenerate {
 		@Inject IResourceValidator resourceValidator
 		
 		override shouldGenerate(Resource resource, CancelIndicator cancelIndicator) {
-			return resourceValidator.validate(resource, CheckMode.NORMAL_AND_FAST, cancelIndicator).empty
+			return resourceValidator.validate(resource, CheckMode.NORMAL_AND_FAST, cancelIndicator)
+				.filter[severity == Severity.ERROR]
```
szarnekow: exists?

szarnekow: A full validate is rather expensive in practice (especially for broken models), I first check of the resource.errors or XtextResource.parseResult.hasSyntaxErrors could be worth it.

svenefftinge: +1 to both suggestions

PR #771, opened 2015-10-30
remove dependency from o.e.xtext to org.eclipse.equinox.common

meysholdt:
This simplifies the classpath for Standalone, IntelliJ and Web.

However, it's still on the classpath as re-exported dependency of 
xtext's optional dependency org.eclipse.core.runtime.

### DIFF
--- a/plugins/org.eclipse.xtext/src/org/eclipse/xtext/service/OperationCanceledManager.xtend
+++ b/plugins/org.eclipse.xtext/src/org/eclipse/xtext/service/OperationCanceledManager.xtend
@@ -7,7 +7,6 @@
  *******************************************************************************/
 package org.eclipse.xtext.service
 
-import org.eclipse.core.runtime.OperationCanceledException
 import org.eclipse.xtext.util.CancelIndicator
 
 /**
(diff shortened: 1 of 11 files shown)


meysholdt:

plugins/org.eclipse.xtext/src/org/eclipse/xtext/service/OperationCanceledManager.xtend
```diff
@@ -68,8 +67,11 @@ class OperationCanceledManager {
 		throw asWrappingOperationCanceledException(platformSpecificOperationCanceledException)
 	}
 	
-	protected def RuntimeException getPlatformSpecificOperationCanceledException() {
-		return new OperationCanceledException()
+	/**
+	 * @since 2.9
+	 */
+	protected def Throwable getPlatformSpecificOperationCanceledException() {
+		return new OperationCanceledError(null)
```
oehme: There is lots of code that calls

```throw e.wrapped```

which would fail with this change.

oehme:
If we do this change (I thought we decided not to do it?), then there are some more todos:

- remove equinox.common from poms and gradle files
- make binding of XtendReferenceFinder lazy, since it depends on ProgressMonitor
- remove the "OperationCanceledException not found" check from mwe2

meysholdt:
I started this change mostly to get a more clear picture of the real costs of removing this dependency. The dependency has been a pain in the past and it will become even more so now that we're about to claim true platform independence. 

Thank you for your pointers.

svenefftinge:
Could you close this PR again, please?

PR #773, opened 2015-11-01
Simplify gradle build

oehme:
There is now only one top-level build script that categorizes the subprojects and applies cross-cutting configuration. The configuration itself is split into small, focused scripts. These scripts have been documented and any outdated and duplicated logic has been removed.

The individual projects only declare dependencies and sometimes specialty tasks (like the javascript compilation in the web project).

This enables us to use --configure-on-demand by default. With that option, Gradle only configures those projects that actually take part in the build. I also enabled --parallel and --daemon by default.

Lastly, I found signed versions of equinox.common and jdt.core on maven central. I replaced all references in our pom/gradle files with these.


.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
szarnekow: Do we need this at the parent level?

gradle/dependency-hygiene.gradle
```diff
@@ -0,0 +1,7 @@
+/*
+ * Put any unwanted transitive dependencies here, they will be excluded from all projects
```
szarnekow: Curiosity: Where do they usually come from in the first place?

gradle/eclipse-project-layout.gradle
```diff
@@ -3,19 +3,21 @@
...
 		java.srcDirs = ['src', 'xtend-gen', 'src-gen']
 		resources {
 			srcDirs = ['src', 'src-gen']
-			exclude (
-				'**/*._trace',
-				'**/*.xtendbin',
-				'**/*.xtend',
-				'**/*.mwe2',
-			)
+			exclude '**/*._trace', '**/*.xtendbin', '**/*.xtend', '**/*.mwe2'
```
svenefftinge: Why do we exclude these files?

gradle/jar-from-build-properties.gradle
```diff
@@ -0,0 +1,39 @@
+/*
+ * Lazily reads build.properties to configure the contents of the jar
+ */
+
+task configureJarFromBuildProperties {
+	doLast {
+		jar {
+			//TODO get rid of this distinction by reconfiguring
+			//the Eclipse output folder for the mavenized projects
+			if(isBuildServer) {
```
svenefftinge: It would be nice to leave a comment on why we have this here.

gradle/versioning.gradle
```diff
@@ -0,0 +1 @@
+version = '2.9.0-SNAPSHOT'
```
szarnekow: Can we put the intellij version here, too? And the versions of the used gradle plugins?

plugins/org.eclipse.xtend.core/META-INF/MANIFEST.MF
```diff
@@ -65,8 +64,6 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="3.6.0",
  org.eclipse.xtext.smap,
  org.apache.ant;bundle-version="1.7.1";resolution:=optional;x-installation:=greedy,
  org.objectweb.asm;bundle-version="[5.0.1,6.0.0)";resolution:=optional,
- org.eclipse.xtext.xbase.junit;resolution:=optional,
- org.eclipse.xtext.generator;resolution:=optional,
- org.eclipse.emf.mwe2.launch;bundle-version="2.7.1";resolution:=optional
+ org.eclipse.xtext.xbase.junit;resolution:=optional
```
szarnekow: I doubt that we could get rid of the dep to xbase.junit from xtend.core :(

settings.gradle
```diff
@@ -2,19 +2,60 @@ rootProject.name = 'xtext'
...
+		apply from: "$rootDir/gradle/build-server-detection.gradle"
+		apply from: "$rootDir/gradle/eclipse-settings.gradle"
+		apply from: "$rootDir/gradle/java-compiler-settings.gradle"
+		apply from: "$rootDir/gradle/versions-from-parent-pom.gradle"
+		apply from: "$rootDir/gradle/versioning.gradle"
+		apply from: "$rootDir/gradle/dependency-hygiene.gradle"
+	
+		def isEclipsePlugin = project.file('build.properties').exists()
+		
+		if (isEclipsePlugin) {
```
szarnekow: Personally I'd prefer clean entry points rather than the 'take them all and try to figure out later what they were' approach.

web/build.gradle
```diff
@@ -1,27 +0,0 @@
...
-	apply plugin: 'java'
-	apply from: "$rootDir/gradle/eclipse-settings.gradle"
-
-	sourceSets {
-		main.java.srcDir 'src/main/xtext-gen'
-		main.java.srcDir 'src/main/xtend-gen'
-		test.java.srcDir 'src/test/xtend-gen'
-	}
-	
-	if (JavaVersion.current().isJava8Compatible()) {
```
szarnekow: Why is this no longer necessary?

gradle/eclipse-project-layout.gradle
szarnekow: I think it should be symmetric to the compiler settings in eclipse. There we filter '*.launch,*.xtend,*._trace,*.xtendbin' in org.eclipse.xtext, for example. The settings should be made consistent across all the compiler configs.

.project
oehme: I'll remove this later

gradle/dependency-hygiene.gradle
oehme: Transitively from other dependencies like Guice.

gradle/versioning.gradle
oehme: Yes, good idea

plugins/org.eclipse.xtend.core/META-INF/MANIFEST.MF
oehme: Well, it was there before, so we would break clients =/

settings.gradle
oehme: What do you mean by 'clean entry points'?

web/build.gradle
oehme: I guess you are referring to the JavaDoc line? That seems to have been a workaround for a broken JavaDoc classpath. I don't see any problems without it.

gradle/jar-from-build-properties.gradle
oehme: I'll reformulate the TODO into a reasoning.

plugins/org.eclipse.xtend.core/META-INF/MANIFEST.MF
szarnekow: Yes, I think one of the existing in-memory-compilers uses some xbase.junit tools :(

settings.gradle
szarnekow: It used to be smth like `[:intellij:, :web:].foo { .. } ` and now its like `all.foo { if (eclipse) else if (intellij) else if (web) }`

gradle/eclipse-project-layout.gradle
oehme: We use 'src' as a resource folder. Gradle would copy those files to the jar if we don't exclude them.

settings.gradle
oehme: Oh I see. We are in the init phase here, so there are no projects yet. Instead, we add a hook on the "Gradle" object, which is project agnostic. This is why we have to do the if-expressions instead.
The performance gain is worth it, though. Instead of always iterating over all projects and eagerly configuring them, we now only configure them one by one when they are needed. That makes builds from subprojects much faster.

oehme:
Jenkins test

oehme:
Jenkins test

oehme:

plugins/org.eclipse.xtend.core/build.properties
```diff
@@ -12,6 +12,8 @@ bin.includes = META-INF/,\
 exclude.. = **/GenerateXtend.mwe2
 src.includes = about.html
 additional.bundles = org.eclipse.xtext.generator,\
-                     org.eclipse.xtext.xtext.generator
+                     org.eclipse.xtext.xtext.generator,\
+                     org.eclipse.emf.mwe2.launch,\
```
dhuebner: org.eclipse.emf.mwe2.launch is treated as required dependency.
mwe2 is a downstream project that depends on xtext and can not be consumed during xtext build.

svenefftinge:
Can we wait with this PR untill after the release?

oehme:
It's not important for the release, so yes.

szarnekow:
Since the release is still 2 weeks away, should we reconsider merging this?

svenefftinge:
ok with me.

PR #782, opened 2015-11-05
[generator] Introduced GeneratorOption

spoenemann:
Solved problem: when using the StandardLanguage and overriding a fragment with its corresponding setter, the default options of the fragment were used instead of the default options of the StandardLanguage. Now it can detect whether explicit values have been set by the user, and apply its own default if not.

### DIFF
--- a/plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
+++ b/plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
@@ -44,26 +44,27 @@ import org.eclipse.xtext.xtext.generator.model.FileAccessFactory
 import org.eclipse.xtext.xtext.generator.model.GuiceModuleAccess
 import org.eclipse.xtext.xtext.generator.model.IXtextGeneratorFileSystemAccess
 import org.eclipse.xtext.xtext.generator.model.JavaFileAccess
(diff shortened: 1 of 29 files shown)


plugins/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
```diff
@@ -44,26 +44,27 @@ import org.eclipse.xtext.xtext.generator.model.FileAccessFactory
...
+	@Accessors(PUBLIC_SETTER)
 	boolean debugGrammar
 	
-	@Accessors
-	Boolean combinedGrammar = null
+	val combinedGrammar = new BooleanGeneratorOption
 	
-	@Accessors
+	@Accessors(PUBLIC_SETTER)
 	boolean removeBacktrackingGuards
```
szarnekow: It would be nice to have the values of these flags available in subtypes, too.

spoenemann: Then we should add a protected getter. I removed public getters where possible in order to reduce the exposed api.

szarnekow: :+1:  for `PUBLIC_SETTER, PROTECTED_GETTER`

PR #786, opened 2015-11-06
[wizard#481478] disable dead maven repos in generated POMs

meysholdt:
this
- upgrades the maven-exec-plugin from 1.2.1 to 1.4.0
- sets cleanupDaemonThreads=false for the maven-exec-plugin
- registers http://nexus.codehaus.org/snapshots/ as disabled

see
https://bugs.eclipse.org/bugs/show_bug.cgi?id=481478
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475098#c3


oehme:
Change looks good, but the CliWizardIntergrationTest is not updated yet.

PR #789, opened 2015-11-09
[xtext][generator] Fixed regression in GrammarAccessFragment

szarnekow:
The serializer got better with regards to keeping escape
sequences which could produce invalid java code.


svenefftinge:
+1

PR #791, opened 2015-11-09
[NodeModel] Wrong result from findActualNode with actions and fragments

szarnekow:


szarnekow:

PR #803, opened 2015-11-14
[482116] Fixed bug in GrammarAccessFragment

szarnekow:
Super rule calls could confuse the getter impl in the
grammar access.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=482116

szarnekow:

svenefftinge:
If a super language overrides a certain rule later on, then it will no longer be considered by the sub language. I don't think it's a big issue, as sublanguages would usually need to be regenerated anyway.

It would be good to have the failing scenario added as a test language.

szarnekow:
Regarding newly added override in some super lang: This scenario was always a breaking change since the since the the return type of `getMyRuleAccess` would have changed from `SuperLangGrammarAccess.MyRuleAccess` to `SubLangGrammarAccess.MyRuleAccess`. IMHO it's more robust after that change.

Will add a test-case.

svenefftinge:
ok, good!

szarnekow:

PR #810, opened 2015-11-19
[482514] Fixed ConcurrentModException

szarnekow:
Also: make return types explicit


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=482514

PR #811, opened 2015-11-19
Web integration improvements

spoenemann:
* Extended hover service so it can compute hovers for selected content assist entries
* Implemented support for editing multiple resources
* Changed session API so a session id is generated only when necessary
* Updated Orion integration to consider latest bugfixes


spoenemann:
Jenkins test

plugins/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/IdeContentProposalProvider.xtend
```diff
@@ -51,7 +51,18 @@ class IdeContentProposalProvider {
 	/**
 	 * Create content assist proposals and pass them to the given acceptor.
 	 */
-	def void createProposals(List<ContentAssistContext> contexts, IIdeContentProposalAcceptor acceptor) {
+	def void createProposals(Collection<ContentAssistContext> contexts, IIdeContentProposalAcceptor acceptor) {
```
sailingKieler: Is there any reason for not using just 'Iterable' instead of 'Collection'?

spoenemann: No, not really

PR #813, opened 2015-11-19
[idea build config] registered 'src-gen2' in 'configure(ideaTestProjects)' in build file of 'intellij' project set

sailingKieler:


sailingKieler:

PR #817, opened 2015-11-20
oomph redirection to local Xtext.setup

LorenzoBettini:
As suggested by Ed and Eike:
This way, once the IDE is setup and has cloned the repository, the setup
from the clone will be used in that IDE, so you can make changes to it,
and perform the tasks to test that the changes are good before you
commit them.

### DIFF
--- a/devtools/org.eclipse.xtext.contributor/Xtext.setup
+++ b/devtools/org.eclipse.xtext.contributor/Xtext.setup
@@ -27,6 +27,11 @@
       value="2g"
       vm="true"/>
   <setupTask
+      xsi:type="setup:EclipseIniTask"
+      option="-Doomph.redirection.xtext="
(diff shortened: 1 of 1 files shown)


svenefftinge:
nice!

oehme:
I did this some time ago and we reverted it.

@dhuebner do you still know the reason?

dhuebner:
Maybe some buggy behavior in the earlier oomph versions? I don't know :p
The benefit here is that the contributor can edit the setup file in his WS and test (mainly the dev team) The disadvantage is that you have to pull the changes to get fixed setup version. Imagine you are in your older  branch and there is an important setup fix in master. How to update your oomphed eclipse?

LorenzoBettini:
This morning I was blocked because of the EMF milestones update site; I modified the setup file but I couldn't update my workspace because the setup was still taken from the remote one.
In case there is an important setup fix in master, the developer working on an older branch could still cherry pick the updated setup from master, couldn't he?

dhuebner:
The most of the clients who uses our setup are not very familiar with oomph. Even if they know how oomph works, they will most likely not be able to fix it. And the experts could add the redirection to the local user setup model.

PR #818, opened 2015-11-20
Extracted method for adding possible var decl to the scope

LorenzoBettini:
This method can be overridden in DSLs that introduce new containers for
variable declarations.  Currently, in such contexts, the whole
_computeTypes for the container, e.g., XBlockExpression, must be
overridden to intercept possible variable declarations.

This also reduces some code duplication ;)

This was also discussed with @szarnekow in chat.


szarnekow:
Jenkins test

PR #819, opened 2015-11-20
Fix for #482665

ghaith:
Modified the bigger than check to bigger than or equal on
isInRequestedRange as it would otherwise skip the last whitespace region
in a document.

Link to original bug : 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=482665

### DIFF
--- a/plugins/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/TextReplacerContext.java
+++ b/plugins/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/TextReplacerContext.java
@@ -183,7 +183,7 @@ public class TextReplacerContext implements ITextReplacerContext {
 		if (regions.isEmpty())
 			return true;
 		for (org.eclipse.xtext.util.ITextRegion region : regions)
-			if (region.getOffset() <= repl.getOffset() && region.getOffset() + region.getLength() > repl.getEndOffset())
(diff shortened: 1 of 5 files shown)


szarnekow:
Jenkins test

ghaith:
The two failing tests appear to be caused by the fact that the Xtend formatter appends a new line to the xtendfile (line 66 in XtendFormatter -> xtendFile.append[newLine])
I would assume this behaviour was never noticed because the last hidden region was never formatted.

I'm assuming we should expect these new lines in the two failing tests, however stopping the new line will also make both tests pass (with no additional tests failing).

I'm not sure what the expected behaviour should be here, should I fix the tests to expect new lines at the end as is requested by the Xtend formatter (And add a test case for Xtend there) or should the Xtend formatter not append new lines to the end of the file?

szarnekow:
Jenkins test

meysholdt:

meysholdt:
hi Ghaith, thank you for the pull request. You found the right spot for the fix. 

Before we can merge this, xtext-runtime-tests-master-latest integration test build should run. Not sure when this will be happening since there are currently troubles with our build server.

Also, let's put the test into the Xtext code base and not into the Xtend code base. I've prepared a pull request for that here: #836

meysholdt:
we decided to let the CI build run the tests since the build server currently doesn't receive the requests for building pull requests.

PR #844, opened 2015-12-02
[483080] proposal of fix for bug 483080 in xtext builder

sailingKieler:
I identified an issue with external referenced resources being available as plain files, not jar entries for example. The issue occurs if those files will receive markers by the 'MarkerUpdaterImpl', like TODO markers. My proposal is to exclude those resources from the marker processing.
To that end I introduced an additional method in 'IStorage2UriMapper' allowing to query only non-external resources. Let me know you thoughts on that.

I did not streamline the changes yet in order to simplify review.

### DIFF
--- a/plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/MarkerUpdaterImpl.java
+++ b/plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/MarkerUpdaterImpl.java
@@ -65,12 +65,17 @@ public class MarkerUpdaterImpl implements IMarkerUpdater {
 		IMarkerContributor markerContributor = getMarkerContributor(uri);
 		CheckMode normalAndFastMode = CheckMode.NORMAL_AND_FAST;
 
-		for (Pair<IStorage, IProject> pair : mapper.getStorages(uri)) {
+		for (Pair<IStorage, IProject> pair : mapper.getStorages(uri, false)) {
 			if (monitor.isCanceled()) {
 				throw new OperationCanceledException();
 			}
 			if (pair.getFirst() instanceof IFile) {
 				IFile file = (IFile) pair.getFirst();
+				
+				// if (file.isLinked(IResource.CHECK_ANCESTORS)) {
+				//		continue;
+				// }
+				
 				if (delta.getNew() != null) {
 					if (resourceSet == null)
 						throw new IllegalArgumentException("resourceSet may not be null for changed resources.");
--- a/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperImpl.java
+++ b/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperImpl.java
@@ -199,24 +199,40 @@ public class Storage2UriMapperImpl implements IStorage2UriMapperExtension {
 
 	@Override
 	public Iterable<Pair<IStorage, IProject>> getStorages(URI uri) {
+		return getStorages(uri, true);
+	}
+
+	/**
+	 * @since 2.9
+	 */
+	@Override
+	public Iterable<Pair<IStorage, IProject>> getStorages(URI uri, boolean includeArchivesAndExternals) {
 		if (!uri.isPlatformResource()) {
 			// support storage lookup by absolute file URI as it is possibly resolved by importURI references
 			if (uri.isFile()) {
 				IPath path = new Path(uri.toFileString());
 				if (path.isAbsolute()) {
 					IFile file = getWorkspaceRoot().getFileForLocation(path);
-					return getStorages(uri, file);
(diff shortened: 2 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=483080

plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperImpl.java
```diff
@@ -199,24 +199,40 @@ protected boolean isHandled(IFolder folder) {
...
-		return getStorages(uri, file);
+		return getStorages(uri, file, includeArchivesAndExternals);
 	}
 
-	private Iterable<Pair<IStorage, IProject>> getStorages(/* @NonNull */ URI uri, IFile file) {
+	private Iterable<Pair<IStorage, IProject>> getStorages(/* @NonNull */ URI uri, IFile file, boolean includeArchivesAndExternals) {
 		if (file == null || !file.isAccessible()) {
-			return contribution.getStorages(uri);
+			if (includeArchivesAndExternals) {
+				return contribution.getStorages(uri);
```
szarnekow: This won't work since it's not necessarily an archive or an external folder that is processed by a contribution

plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/MarkerUpdaterImpl.java
```diff
@@ -65,12 +65,17 @@ private void processDelta(Delta delta, /* @Nullable */ ResourceSet resourceSet,
...
 		CheckMode normalAndFastMode = CheckMode.NORMAL_AND_FAST;
 
-		for (Pair<IStorage, IProject> pair : mapper.getStorages(uri)) {
+		for (Pair<IStorage, IProject> pair : mapper.getStorages(uri, false)) {
 			if (monitor.isCanceled()) {
 				throw new OperationCanceledException();
 			}
 			if (pair.getFirst() instanceof IFile) {
 				IFile file = (IFile) pair.getFirst();
+
```
szarnekow: I think it would make sense to compare the `pair.getSecond` with the currently processed project and use that as an additional criteria.

szarnekow: Is the uri a platform:/resource or a platform:/plugin URI?

sailingKieler: platform:/resource
And indeed `pair.getSecond` gets the project being built, not the project containing resource in that case.

plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperImpl.java
sailingKieler: But I would like to restrict on project local resources only in this case, and ignore all project external resources here. Why should that be a problem? Sure, for a lot of use cases this limited set of resources is not sufficient.

szarnekow: I wanted to point out that nothing in the contribution.getStorages(uri) suggest that the contributed storages are really external. So clients are in general free to contribute arbitrary storages that are potentially considered to be local, too. In other words: The contribution would need to make the decision on its own and take / process the boolean flag. Which in turn would be an even more sever API change.

plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/MarkerUpdaterImpl.java
svenefftinge: We should try to identify these to be skipped resources without touching the mapper.
What is file.getProject() returning in your particular case?
Also what is file.isReadOnly() returning?

szarnekow: Christian is just providing another patchset that does special treatment for the external.folder jdt dummy project which bites us here.

sailingKieler: @svenefftinge `file.isReadOnly()` even returns `true` in that case
I'm going to open another PR

szarnekow: For completeness: `file.isReadonly` is by no means an indicator whether a file should be processed or not, e.g. pessimistic VCS will checkout files as readonly by default.

PR #847, opened 2015-12-03
typo

kthoms:


spoenemann:
I fixed that line in [commit 951c008](https://github.com/eclipse/xtext/commit/951c008b072c89829946adfe89a64fb385dd5da2). I think the part "If not specified other" was just the remainder of a sentence that was not eliminated properly.

By the way, pull requests about fixing documentation should be merged into maintenance in order to have an immediate effect on the website.

PR #852, opened 2015-12-07
[web] Fixed search for generator artifacts

spoenemann:
This allows to omit the DEFAULT_OUTPUT prefix in the 'artifact' parameter.


spoenemann:
Jenkins test

PR #868, opened 2015-12-11
Bug 484215 - [builder] Performance and OOM problems with QualifiedName while building large projects

iloveeclipse:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=484215

### DIFF
--- a/plugins/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
+++ b/plugins/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
@@ -12,6 +12,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import org.eclipse.emf.common.util.CommonUtil;
 import org.eclipse.emf.ecore.resource.impl.BinaryResourceImpl.EObjectInputStream;
 import org.eclipse.emf.ecore.resource.impl.BinaryResourceImpl.EObjectOutputStream;
 import org.eclipse.xtext.util.Strings;
@@ -32,6 +33,8 @@ public class QualifiedName implements Comparable<QualifiedName> {
 	private final String[] segments;
 
 	private QualifiedName lowerCase;
+
+	private static final boolean USE_INTERNING = Boolean.getBoolean("xtext.qn.interning");
 
 	public static final QualifiedName EMPTY = new QualifiedName() {
 		@Override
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=484215

dhuebner:
Jenkins test

plugins/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
```diff
@@ -102,25 +123,25 @@ public static QualifiedName createFromStream(EObjectInputStream eObjectInputStre
...
 		} else {
-			return new QualifiedName(segments);	
+			return new QualifiedName(segments);
 		}
 	}
 
 	private static String[] readSegmentArray(EObjectInputStream from, int count, String first) throws IOException {
 		String[] segments = new String[count];
-		segments[0] = first;
+		segments[0] = first == null? null : intern(first);
```
szarnekow: This null check shouldn't be necessary. Did you have any issues here?

plugins/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
```diff
@@ -102,25 +123,25 @@ public static QualifiedName createFromStream(EObjectInputStream eObjectInputStre
...
 		for (int i = 1; i < count; i++) {
 			String segment = from.readSegmentedString();
-			segments[i] = segment;
+			segments[i] = segment == null? null : intern(segment);
```
szarnekow: Same here

szarnekow:
Jenkins test

plugins/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
iloveeclipse: Looking at the code of EObjectInputStream the "from" stream can return null.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
iloveeclipse: Same here.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java
szarnekow: It can theoretically return null, but that would indicate that we wrote a null value into the string in the first place. Since no segment of a QN may ever be null, it would be wrong to tolerate null values here. Thus I'd prefer to raise an exception.

iloveeclipse: You mean, just skip the NP check?

szarnekow: Exactly.

iloveeclipse: Exactly *what*? Skip or throw?

szarnekow: Sorry to be unclear here. At first I though skipping the NPE check is good enough. But to be on the safe side: Please throw an IllegalStateException if the segment is null when it's to be put into the array.

iloveeclipse:
I've updated the pull request.
BTW, how one can see in github if the changeset is updated or not? Is there any hint in the UI?

szarnekow:
It's possible to deduce the fact 'changeset updated' the comment history. If you do not amend but push a new commit to your branch, the PR will reflect that, too.

szarnekow:
Jenkins test

szarnekow:
One follow up question (and probably smth for a follow-up PR): Do you see many duplicates in your qualified name instances? Would it be worthwhile to intern the entire array?

iloveeclipse:
Since the QN's are hierarchical in our case, and all the files have one common root package, with 26000 QN's we have at least 25999 duplicates of the root package. If we go down and expect something like a binary distribution, we will have 25999 + 12999 + 6499 + ... duplicates. Also interning arrays doesn't make sense since all QN's are unique and so each array will have at least one segment different to others.

dhuebner:
@iloveeclipse could you please reopen your PR for the maintenance branch instead of master?
We consider to add it for the 2.9 stream.

iloveeclipse:
How I can do this on github?

dhuebner:
@iloveeclipse https://github.com/isaacs/github/issues/18 
I would probably open a new PR against maintenance and link this PR in comments.

iloveeclipse:
I've cherry-picked my commits to the maintenance branch, see pull request #888

dhuebner:
@iloveeclipse Thanks, Andrey.

dhuebner:
Merged into master from maintenance with https://github.com/eclipse/xtext/issues/888

PR #869, opened 2015-12-11
[484201] don't try to load synthetic classes with Java source language

oehme:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=484201

PR #872, opened 2015-12-15
Docs for parameters of Check

olenz:
I noted that the parameter of the Check annotation is missing from the xtext docs. I added some docs.


svenefftinge:
thanks!

PR #879, opened 2015-12-16
[483088] allow nested types in implementationClass in ecore models.

svenefftinge:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=483088

PR #883, opened 2015-12-17
[bug 483931] Relativize output paths even for non existent folders 

akosyakov:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=483931

PR #902, opened 2016-01-18
[bug 469446] Make identifier computation for WildcardTypeReference and JvmWildcardTypeReference symmetric

akosyakov:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=469446

szarnekow:
The change looks wrong to me. GetIdentifier shoud return an unambiguous string representation whereas <?> may be ambiguous for types like `T<E extends CharSequence>`. T<?> will usually be interpreted as being `T<? extends CharSequence>` rather than `T<? extends Object>`.

szarnekow:
Could be that the overrideHelper should use `LightweigthTypeReference.getUniqueIdentifier`

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/references/WildcardTypeReference.java
```diff
@@ -336,8 +336,10 @@ private String getAsString(Function<LightweightTypeReference, String> format) {
 		if (lowerBound != null) {
 			return "? super " + format.apply(lowerBound);
 		}
-		if (upperBounds != null && upperBounds.size() == 1 && upperBounds.get(0).isType(Object.class)) {
-			return "?";
+		if (format == SimpleNameFunction.INSTANCE) {
```
szarnekow: Why not change the signature of the private method and add a boolean flag to make the behavior more explicit?

PR #903, opened 2016-01-19
[486120] fixes concurrency issues causing NPEs during 'Resource.eNotify(...)'

sailingKieler:

### DIFF
--- a/plugins/org.eclipse.xtext.util/src/org/eclipse/xtext/util/OnChangeEvictingCache.java
+++ b/plugins/org.eclipse.xtext.util/src/org/eclipse/xtext/util/OnChangeEvictingCache.java
@@ -15,6 +15,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.log4j.Logger;
+import org.eclipse.emf.common.notify.Adapter;
 import org.eclipse.emf.common.notify.Notification;
 import org.eclipse.emf.common.util.WrappedException;
 import org.eclipse.emf.ecore.resource.Resource;
@@ -109,11 +110,19 @@ public class OnChangeEvictingCache implements IResourceScopeCache {
 	 * @return the cache adapter for the given resource. Never <code>null</code>.
 	 */
 	public CacheAdapter getOrCreate(Resource resource) {
-		CacheAdapter adapter = (CacheAdapter) EcoreUtil.getAdapter(resource.eAdapters(), CacheAdapter.class);
-		if (adapter == null) {
-			adapter = new CacheAdapter();
-			resource.eAdapters().add(adapter);
-			adapter.setResource(resource);
+		// ask for the list of adapters twice in order to avoid the usage of a
+		//  potentially stale list because of a race of several involved threads
+		resource.eAdapters();
+		final List<Adapter> adapters = resource.eAdapters();
+		
+		CacheAdapter adapter;
+		synchronized (adapters) {			
+			adapter = (CacheAdapter) EcoreUtil.getAdapter(adapters, CacheAdapter.class);
+			if (adapter == null) {
+				adapter = new CacheAdapter();
+				adapters.add(adapter);
+				adapter.setResource(resource);
+			}
 		}
 		return adapter;
 	}


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=486120

meysholdt:

meysholdt:

dhuebner:

dhuebner:

meysholdt:
+1 but it would be nice if more people would look at this.

plugins/org.eclipse.xtext.util/src/org/eclipse/xtext/util/OnChangeEvictingCache.java
```diff
@@ -109,11 +110,19 @@ protected void cacheHit(CacheAdapter adapter) {
...
 	 */
 	public CacheAdapter getOrCreate(Resource resource) {
-		CacheAdapter adapter = (CacheAdapter) EcoreUtil.getAdapter(resource.eAdapters(), CacheAdapter.class);
-		if (adapter == null) {
-			adapter = new CacheAdapter();
-			resource.eAdapters().add(adapter);
-			adapter.setResource(resource);
+		// ask for the list of adapters twice in order to avoid the usage of a
+		//  potentially stale list because of a race of several involved threads
+		resource.eAdapters();
```
szarnekow: This does not prevent a race. Consider two thread T1 and T2:
T1 enters this method and calls resource.eAdapter and is suspended before eAdapters returns and the field of the resource is set.
T2 enters this method, calls eAdapters, creates a new list, calls eAdapter again and obtains the same list and syncs on that one.
T1 is now overriding the adapter list of the resource and syncs on a different instance.

szarnekow:
:-1: 
This does not fix the problem (see comment in code). A unit test would be nice, e.g. two/five threads, a list with 10k resources and both threads loop over the list and obtain a cache value. It's to be asserted that the adapter lists contain exactly one element afterwards for all resources.

szarnekow:
Also note that other clients may install adapters on the resource concurrently, too and their code would not use this method.

plugins/org.eclipse.xtext.util/src/org/eclipse/xtext/util/OnChangeEvictingCache.java
sailingKieler: I'm absolutely aware of that scenario, and I discussed that problem with Moritz. Nonetheless, the list is most likely already available because of the uri change adapter installed by the XtextResourceSet. Can the access to the adapters list of arbitrary resources be coordinated, at all?

sailingKieler:
The more important concern of this change is coordinating the search and insertion into a single adapters list, rather than coordinating the initialization of the adapters list. Loosing an adapters list because of the race of several threads is bad but won't hurt wrt. to the CacheAdapters. Breaking the list's basic assumption of containing no 'null' entries, however, does hurt a lot!

szarnekow:
Simply sync'ing on the resource itself does not work?

sailingKieler:
LazylinkingResource.getEObject() already syncs on the resource, so that would be very deadlock prone

szarnekow:
> Loosing an adapters list because of the race of several threads is bad but won't hurt wrt. to the CacheAdapters. Breaking the list's basic assumption of containing no 'null' entries, however, does hurt a lot!

It's not about loosing the adapter list but about two threads that sync on different instances which will yield the exact same null entry in the array that you want to get rid off. If you want to make sure that the adapter list is always initialized, I'd add a call to `eAdapters()` to the constructor of XtextResource and avoid the `ask twice` thing here, since it's giving a false impression of security.

svenefftinge:
I agree with Sebastian.
In the bugzilla it sounds like you can reproduce the issue.
Do you have a full dump for when the NPE happens?
I wonder why multiple threads can access the resource simultaneously if they go through the XtextDocument#readOnly API.

sailingKieler:
This happens when several resources are asked to resolve a reference to another resource. I observed scenarios were 2 or even 3 resources (each in a different thread) tried to access a 4th one  treads more or less simultaneously. (I also could believe that this actually happens, that might be facilitated by the JVM in debug mode). This was caused by the builder, the semantic highlighting calculator of an open xtext editor, and an additional view that synchronizes on another open editor.

Accessing the adapters list was not a problem as it was already initialized but accessing/adding the CacheAdapter.

sailingKieler:
> It's not about loosing the adapter list but about two threads that sync on different instances which will yield the exact same null entry in the array that you want to get rid off.
I don't think that this is correct since only that list is manipulated from which the lock is taken. In case the lock is taken from different lists, the manipulation would happen on different lists.

> If you want to make sure that the adapter list is always initialized, I'd add a call to eAdapters() to the constructor of XtextResource and avoid the ask twice thing here, since it's giving a false impression of security.
That's true, I'll better do it that way ;-)

svenefftinge:
> This happens when several resources are asked to resolve a reference to another resource. I observed scenarios were 2 or even 3 resources (each in a different thread) tried to access a 4th one more or less simultaneously. (I also couldn't believe that this actually happens, that might be facilitated by the JVM in debug mode). This was caused by the builder, the semantic highlighting calculator of an open xtext editor, and an additional view that synchronizes on another open editor.

Note that the builder and each editor have their own ResourceSet and don't share any state. So if this was really the case then your setup is broken. 

We don't access the resourceset from the builder in parallel, with the exception of the ParrallelBuilderParticipant (currently only used for xtend). The state of an editor is wrapped by an XtextDocument, and one always needs to go through XtextDocument.readOnly (or modfy) to access it. These methods synchronize on the Document itself (or the resource if it implements ISynchronizable), so there also is no parallel access.

I can imagine that this happens, if you break these contracts, i.e. share the same resource between multiple XtextDocument or keep references to the state some how and access it outside a transaction.

sailingKieler:
> Note that the builder and each editor have their own ResourceSet and don't share any state.

hmmm... that's true. That leads me to the conclusion that this must be due to (or at least facilitated by) our employment of those ReferencingResourceSets, which are currently not coordinated wrt. accessing the referenced resource sets.

PR #907, opened 2016-01-22
[web] Fixed package imports and exports in generated manifests

spoenemann:


spoenemann:
See [Gradle OSGi plugin documentation](https://docs.gradle.org/current/userguide/osgi_plugin.html).

PR #908, opened 2016-01-22
[bug 464563] Check that a called feature does not refer invalid types

akosyakov:
Change-Id: Ieec5b2e092557d9c05d46e385d40ad86df5ff13d
Signed-off-by: akosyakov <anton.kosyakov@itemis.de>

### DIFF
--- a/plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/RefereredInvalidTypeFinder.xtend
+++ b/plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/RefereredInvalidTypeFinder.xtend
@@ -1,0 +1,130 @@
+/** 
+ * Copyright (c) 2016 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.xtext.xbase.validation
+
+import com.google.inject.Inject
+import com.google.inject.Singleton
+import java.util.List
+import org.eclipse.xtext.common.types.JvmExecutable
+import org.eclipse.xtext.common.types.JvmField
+import org.eclipse.xtext.common.types.JvmIdentifiableElement
+import org.eclipse.xtext.common.types.JvmOperation
+import org.eclipse.xtext.common.types.JvmTypeReference
+import org.eclipse.xtext.xbase.typesystem.references.ArrayTypeReference
+import org.eclipse.xtext.xbase.typesystem.references.CompoundTypeReference
+import org.eclipse.xtext.xbase.typesystem.references.LightweightTypeReference
+import org.eclipse.xtext.xbase.typesystem.references.LightweightTypeReferenceFactory
+import org.eclipse.xtext.xbase.typesystem.references.ParameterizedTypeReference
+import org.eclipse.xtext.xbase.typesystem.references.StandardTypeReferenceOwner
+import org.eclipse.xtext.xbase.typesystem.references.TypeReferenceVisitorWithResult
+import org.eclipse.xtext.xbase.typesystem.references.WildcardTypeReference
+import org.eclipse.xtext.xbase.typesystem.util.CommonTypeComputationServices
+
+/** 
+ * @author kosyakov - Initial contribution and API
+ */
+@Singleton
+class RefereredInvalidTypeFinder extends TypeReferenceVisitorWithResult<LightweightTypeReference> {
+
+	@Inject CommonTypeComputationServices services
+
+	def LightweightTypeReference findRefereredInvalidType(JvmIdentifiableElement element) {
+		return element.internalFindRefereredInvalidType
+	}
+
+	def protected dispatch LightweightTypeReference internalFindRefereredInvalidType(JvmIdentifiableElement field) {
+		return null
+	}
+
+	def protected dispatch LightweightTypeReference internalFindRefereredInvalidType(JvmField field) {
+		val type = field.type.toLightweightTypeReference
+		if (type.primitiveVoid)
+			return type
+		
+		return type.findUnknownType
+	}
+
+	def protected dispatch LightweightTypeReference internalFindRefereredInvalidType(JvmOperation operation) {
+		val unknownType = operation.returnType.findUnknownType 
+		if(unknownType !== null)
+			return unknownType
+		
+		return _internalFindRefereredInvalidType(operation as JvmExecutable)
+	}
+
+	def protected dispatch LightweightTypeReference internalFindRefereredInvalidType(JvmExecutable executable) {
+		for (typeReference : executable.typeParameters.map[constraints].flatten.map[typeReference]) {
+			val unknownType = typeReference.findUnknownType
+			if(unknownType !== null)
+				return unknownType
+		}
+		for (parameterType : executable.parameters.map[parameterType.toLightweightTypeReference]) {
+			if(parameterType.primitiveVoid)
+				return parameterType
+			
+			val unknownType = parameterType.findUnknownType
+			if(unknownType !== null)
+				return unknownType
+		}
+		for (exception : executable.exceptions) {
+			val unknownType = exception.findUnknownType
+			if(unknownType !== null)
+				return unknownType
+		}
+		return null
+	}
+
+	def protected LightweightTypeReference findUnknownType(JvmTypeReference reference) {
+		return reference.toLightweightTypeReference.findUnknownType
+	}
+
+	def protected LightweightTypeReference findUnknownType(LightweightTypeReference type) {
+		return type.accept(this)
+	}
+
+	override protected LightweightTypeReference doVisitTypeReference(LightweightTypeReference reference) {
+		if(reference.unknown)
+			return reference
+
+		return null
+	}
+
+	override protected LightweightTypeReference doVisitArrayTypeReference(ArrayTypeReference reference) {
+		return reference.componentType.accept(this)
+	}
+
+	override protected LightweightTypeReference doVisitWildcardTypeReference(WildcardTypeReference reference) {
+		val unknownType = reference.upperBounds.visit
+		if(unknownType !== null)
+			return unknownType
+
+		val lowerBound = reference.lowerBound
+		if(lowerBound !== null)
+			return lowerBound.accept(this)
+
+		return null
+	}
+
+	override protected LightweightTypeReference doVisitParameterizedTypeReference(ParameterizedTypeReference reference) {
+		return reference.typeArguments.visit
+	}
+
+	override protected LightweightTypeReference doVisitCompoundTypeReference(CompoundTypeReference reference) {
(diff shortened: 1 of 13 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=464563

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/RootResolvedTypes.java
```diff
@@ -167,6 +167,12 @@ protected boolean isIntentionalEarlyExit(/* @Nullable */ XExpression expression)
 	}
 	
 	protected AbstractDiagnostic createTypeDiagnostic(XExpression expression, LightweightTypeReference actualType, LightweightTypeReference expectedType) {
+		if (expectedType.isUnknown() && actualType.isAny())
```
szarnekow: This looks wrong rather ad-hoc to me. Any should already be conformant to unknown, otherwise the fix should go into the conformance check.

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/RootResolvedTypes.java
```diff
@@ -167,6 +167,12 @@ protected boolean isIntentionalEarlyExit(/* @Nullable */ XExpression expression)
...
+			return null;
+		
+		if (expectedType.isType(Object.class) && actualType.isUnknown())
```
szarnekow: Which problem is this special treatment supposed to solve?

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/RootResolvedTypes.java
```diff
@@ -131,7 +131,7 @@ protected void addTypeDiagnostic(XExpression expression, /* @Nullable */ TypeDat
 				if (expectedType != null) {
 					if (!expectedType.isPrimitiveVoid()) {
 						int conformanceFlags = getConformanceFlags(typeData, false);
-						if (!((conformanceFlags & ConformanceFlags.SUCCESS) != 0)) {
+						if ((conformanceFlags & ConformanceFlags.SUCCESS) == 0 || (conformanceFlags & ConformanceFlags.UNKNOWN_TYPE_PARTICIPATED) != 0) {
```
szarnekow: Why would you create a conformance problem for unknown types? They should be flagged as missing on the type ref itself.

akosyakov: — Bar.xtend —
```
class Bar {
  def void bar(Unknown it) {}
}
```

This file has an error for an unresolved type.

— Foo.xtend —
```
class Foo {
  def void foo(Bar bar) {
    bar.bar(notify)
  }
}
```

This file does not have any errors, but compilation fails, since notify of void type. Adding a conformance problem for the notify call will fix the issue.

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/internal/RootResolvedTypes.java
svenefftinge: @Anton: If these case handling is neccessary here,  could you put a comment into the code, that explains the cases?

svenefftinge:
The change looks good to me. The two failing test cases in FeatureCallValidationTest need to be adjusted, though.

szarnekow:
I'm skeptical about this change and would rather try another approach, e.g use NonVoidExpectation instead of an UnknownType if an expectation is to be set for a feature call argument or if the expectation is-to-set for a thing that definitely expects a value.

svenefftinge:
Could you explain why you don't like the current approach?
It sounds natural that these expectations are of UnkownTypes, as they are unknown. 
Just to be sure we are on the same page: the problem we want to fix here, is that we usually treat everything we do on UnknownTypes as a silent follow up problem. But this approach doesn't work for non-local elements, as Anton explained. 
E.g. if you call a method that is declared (and flagged erroneous) somewhere else, you want to have an error on the callsite, too.

szarnekow:
Yes, you want to see an error but something like 'void is not assignable to DoesNotExist' is not the most intuitive user feedback. Java offers something like 'The method myMethod(DoesNotExist) from the type OtherType refers to the missing type DoesNotExist' which is much easier to grasp by the programmer. A validation along these lines wouldn't require strange special treatment in the generic type conformance validation and wouldn't introduce follow up errors that shadow the real problem.

svenefftinge:
Yes, flagging the actual feature references that have unknown types on its declaration signature seems better than flagging the incompatible arguments.

svenefftinge:

svenefftinge:
Jenkins test

akosyakov:
Could someone review the change?

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/RefereredInvalidTypeFinder.xtend
```diff
@@ -0,0 +1,130 @@
...
+import org.eclipse.xtext.xbase.typesystem.references.StandardTypeReferenceOwner
+import org.eclipse.xtext.xbase.typesystem.references.TypeReferenceVisitorWithResult
+import org.eclipse.xtext.xbase.typesystem.references.WildcardTypeReference
+import org.eclipse.xtext.xbase.typesystem.util.CommonTypeComputationServices
+
+/** 
+ * @author kosyakov - Initial contribution and API
+ */
+@Singleton
+class RefereredInvalidTypeFinder extends TypeReferenceVisitorWithResult<LightweightTypeReference> {
```
svenefftinge: It is 'referred' or better 'referenced'

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/UIStrings.java
```diff
@@ -153,40 +146,16 @@ protected String referencesToString(Iterable<? extends JvmTypeReference> element
...
-		}
-		return defaultLabel;
+		if (typeRef == null)
+			return defaultLabel;
+		
+		if (typeRef.eResource() == null)
+			return typeRef.getSimpleName();
+		
+		StandardTypeReferenceOwner owner = new StandardTypeReferenceOwner(services, typeRef);
+		LightweightTypeReferenceFactory factory = new LightweightTypeReferenceFactory(owner, false);
```
szarnekow: Minor: the owner has already a factory that can be used via owner.toLightweight...

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/RefereredInvalidTypeFinder.xtend
```diff
@@ -0,0 +1,130 @@
...
+		return reference.multiTypeComponents.visit
+	}
+
+	def protected LightweightTypeReference visit(List<LightweightTypeReference> references) {
+		return references.map[accept(this)].filterNull.head
+	}
+
+	def protected LightweightTypeReference toLightweightTypeReference(JvmTypeReference typeRef) {
+		val owner = new StandardTypeReferenceOwner(services, typeRef)
+		val factory = new LightweightTypeReferenceFactory(owner, false)
```
szarnekow: Same here, no need to deal with the internals of the owner manually

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/UIStrings.java
```diff
@@ -153,40 +146,16 @@ protected String referencesToString(Iterable<? extends JvmTypeReference> element
...
-							return defaultLabel;
-						}
-					}
-				} else {
-					return typeRef.getSimpleName();
-				}
-			}
-		}
-		return defaultLabel;
+		if (typeRef == null)
```
szarnekow: The refactoring of this method breaks the string represenation for JvmAnyTypeRef that is not contained in a resource.

akosyakov: Is there a test for it?

szarnekow: Not that I'm aware of. Since the code had 'full' support for dangling JvmTypeRefs before it was refactored, it should either add an assertion (which would likely break xbase...UIStringsTest) or new test cases should be added in that test class and the proper handling of these should be restored.

akosyakov: ok, I will add new tests. it was not just refactored, the method was bogus for handling unresolved type, e.g. for List<Unknown> it used to produce List<void> and so on.

akosyakov:
Could someone review the change?

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/validation/UIStrings.java
```diff
@@ -13,33 +13,27 @@
...
-import org.eclipse.xtext.util.ITextRegion;
 import org.eclipse.xtext.util.Strings;
-import org.eclipse.xtext.util.Triple;
 import org.eclipse.xtext.xbase.XAbstractFeatureCall;
 import org.eclipse.xtext.xbase.XConstructorCall;
 import org.eclipse.xtext.xbase.XExpression;
 import org.eclipse.xtext.xbase.typesystem.IBatchTypeResolver;
 import org.eclipse.xtext.xbase.typesystem.IResolvedTypes;
 import org.eclipse.xtext.xbase.typesystem.references.LightweightTypeReference;
+import org.eclipse.xtext.xbase.typesystem.references.LightweightTypeReferenceFactory;
```
dhuebner: unused import?

dhuebner: In corresponding bug report the fix is aimed for 2.9.2. You should reopen PR against the maintenance stream or change whiteboard to v2.10.

akosyakov: @svenefftinge Should it be a part of 2.9.2?

dhuebner:
@svenefftinge I think at least the change in XbaseBatchCompiler should be well tested for 2.9.x

svenefftinge:
Yes, it should go into 2.9.2. 
@dhuebner What kind of problem do you expect from the fix in the batchcompiler? It looks good to me. If you have any specific scenarios in mind, please go ahead and check them.

dhuebner:
@svenefftinge I didn't said it is wrong. It looked to me like it may change the commandLine string.

svenefftinge:
Ok, I don't think that's the case.

PR #909, opened 2016-01-22
[wizard] Bug 486323: add dependencies to old generator when creating a Gradle project from an existing Ecore model

spoenemann:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=486323

PR #914, opened 2016-01-26
[486527] Respect logical container in type scope

svenefftinge:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=486527

tests/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/macro/AbstractReusableActiveAnnotationTests.xtend
```diff
@@ -3363,6 +3363,87 @@ abstract class AbstractReusableActiveAnnotationTests {
...
+							it.visibility = method.visibility
+							it.synchronized = method.synchronized
+							it.deprecated = method.deprecated
+							it.static = method.static
+							it.exceptions = method.exceptions
+							it.varArgs = method.varArgs
+							it.strictFloatingPoint = method.strictFloatingPoint
+						]
+				
+						for (t : method.typeParameters) {
```
akosyakov: If i remove copying of type parameters then T will be resolved to the type parameter T of the original method and we will get quite confusing error: 'Cannot make a static reference to the non-static type T'. I wonder should the resolution of T fail, since the body is moved to another method?

PR #916, opened 2016-01-26
[xbase compiler] Bug 485556: fields referenced from switch cases must be static in Java

spoenemann:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=485556

spoenemann:
Jenkins test

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/SwitchConstantExpressionsInterpreter.xtend
```diff
@@ -50,7 +50,7 @@ class SwitchConstantExpressionsInterpreter extends AbstractConstantExpressionsIn
 					if (feature.constant) {
 						return feature.constantValue
 					}
-				} else if (feature.final) {
+				} else if (feature.final && feature.static) {
```
szarnekow: I think final and static is not sufficient, e.g. a field must be a compile time constant (see isSetConstant and isConstant).

tests/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ValidationBug427432Test.xtend
```diff
@@ -31,7 +31,7 @@ class ValidationBug427432Test extends AbstractXtendTestCase {
 		'''
 			class Foo {
 			
-				val x = 1
+				static val x = 1
```
szarnekow: This static keyword should not be necessary to issue the warning.

spoenemann: Without the static keyword the switch case constant values aren't evaluated anymore, so the validation does not work anymore for this case. This is ok since the compiler now does not create a switch, but an if-else chain. I removed the static modifier and changed the assertion to "no errors".

szarnekow: I don't share the opinion that this is ok, since the second case in the test model is dead code but no longer flagged as such.

spoenemann: The same applies to other test cases: `finalInstanceField_2()`, `finalInstanceField_3()`, `notFinalInstanceField()`, `notFinalStaticField()`, `notFinalStaticField_2()`, etc.
If we want the dead code to be detected in all these cases, we have to change the implementation of that validation. But that would be worth another bug IMO.

tests/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/ValidationBug427432Test.xtend
```diff
@@ -41,7 +41,7 @@ class ValidationBug427432Test extends AbstractXtendTestCase {
 				}
 			
 			}
-		'''.parse.assertError(XbasePackage.Literals.XFEATURE_CALL, IssueCodes.DUPLICATE_CASE)
+		'''.parse.assertNoErrors
```
szarnekow: Maybe I just don't get it. We still have dead code in the test model but now we assert that we don't find it?

spoenemann: I just adapted the expectation to other similar cases. The validation uses `SwitchConstantExpressionsInterpreter`, which is the  point where I fixed the compiler bug. When that interpreter is unable to evaluate the switch cases, the validation simply does nothing.

szarnekow: My gut feeling is that the SwitchConstantExpressionsInterpreter needs two modes - one for the compiler and another for the validation. No longer detecting problematic code seems to be a bad side effect of the compiler fix that's worth to reconsider the nature of the fix itself.

spoenemann: My point was that we weren't able to detect all problematic cases already before. But ok, it makes sense to keep detecting at least those cases that can be realized with affordable effort.

PR #925, opened 2016-02-03
[documentation] Updated and reworked the documentation

spoenemann:
Restructured main chapters as follows:
 1. The Grammar Language
 2. Configuration
 3. Basic Language Features
 4. Integration with Java
 5. Integration with EMF
 6. Eclipse Support
 7. IntelliJ IDEA Support (TODO)
 8. Web Editor Support
 9. Continuous Integration

The chapters "MWE 2" and "Typical Language Configurations" are now hidden from the overview. The pages are still there, but their content should be moved elsewhere ASAP.


svenefftinge:
Thinking about it once more, I don't like "basic language features" as a header for the non-ui features. The term language in it suggests that it's features in some language, but it is about features in a compiler or language implementation. Maybe we should call it "language implementation features"?

svenefftinge:
otherwise i like the restructuring

spoenemann:
Or simply "Language Implementation"?

svenefftinge:
:+1:

PR #926, opened 2016-02-11
add scope to maven dependency example

jangalinski:
since xtend does not need to be a transitive dependency at runtime, it's a good idea to mark it provided ...

### DIFF
--- a/xtend-website/download.md
+++ b/xtend-website/download.md
@@ -81,6 +81,7 @@ If you already have a project, you need to add the Xtend library:
   <groupId>org.eclipse.xtend</groupId>
   <artifactId>org.eclipse.xtend.lib</artifactId>
   <version>2.9.0</version>
+  <scope>provided</scope>
 </dependency>
 ```
 


meysholdt:
Hi Jan, thank you for your contribution. However, the Eclipse Foundation requires all contributors to sign the CLA (see https://www.eclipse.org/legal/CoO.php, this can be done online) before we can accept contributions. And commits need to be Signed-Off with the same email address that has been used to sign the CLA.

meysholdt:
Regarding your change: marking dependency "org.eclipse.xtend.lib" as provided tells maven that at runtime this dependency will already be available, i.e. coming from somewhere else. However, I don't see where it would come from.

"org.eclipse.xtend.lib" is a really small artifact with utility methods that are often used by the Java code that's produced by the Xtend compiler. I.e. to make TemplateExpressions work. So it needs to be there at runtime.

"org.eclipse.xtend.lib" is *not* the Xtend parser/compiler.

jangalinski:
Thanks Moritz, I misunderstood the concept (and implemented a wrong example :-) ) .... I will close this one.

PR #941, opened 2016-02-25
Quickfix for Spaces and Tabs in Keyword

cdietrich:
…458). Added a Quickfix that replaces spaces and tabs in keyword with _

### DIFF
--- a/plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
+++ b/plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
@@ -14,6 +14,7 @@ import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.util.AbstractList;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
@@ -27,10 +28,13 @@ import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.emf.codegen.ecore.genmodel.GenModel;
 import org.eclipse.emf.codegen.ecore.genmodel.GenModelPackage;
 import org.eclipse.emf.codegen.ecore.genmodel.GenPackage;
+import org.eclipse.emf.common.util.BasicEList;
+import org.eclipse.emf.common.util.EList;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EDataType;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.EPackage;
+import org.eclipse.emf.ecore.EStructuralFeature;
 import org.eclipse.emf.ecore.EStructuralFeature.Setting;
 import org.eclipse.emf.ecore.EcorePackage;
 import org.eclipse.emf.ecore.plugin.EcorePlugin;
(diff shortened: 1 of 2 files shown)


svenefftinge:
Replacing with underscore is usually not what people want, but splitting a keyword into multiple ones.

cdietrich:
i thought of that but that fix is not generally applicable (unassigned keywords)

JanKoehnlein:
You mean assigned keywords?
I'd rather have QuickFix "a split into two keywords" for the cases where it is applicable (I guess that is unassigned keywords, correct me if I am wrong), and no fix for the other. An issue in both cases.

cdietrich:
i mean the quickfix works for "my kw" but not for assign="my kw"

cdietrich:
now i have a quickfix for the "simple" cases

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
```diff
@@ -164,6 +169,42 @@ public void apply(final EObject element, IModificationContext context) {
...
+	public void fixSpacesInKeyword(final Issue issue, IssueResolutionAcceptor acceptor) {
+		if (issue.getData().length > 0 && Boolean.valueOf(issue.getData()[0]))
+		acceptor.accept(issue, "Replace Keyword with multiple Keywords", "Splits a Keyword that contains spaces or tabs into multiple Keywords", NULL_QUICKFIX_IMAGE,
+				new ISemanticModification() {
+			@Override
+			public void apply(final EObject element, IModificationContext context) {
+				Keyword keyword = (Keyword) element;
+				EStructuralFeature eContainingFeature = keyword.eContainingFeature();
+				EObject eContainer = keyword.eContainer();
+				String[] parts = keyword.getValue().split("[\\t ]");
```
JanKoehnlein: How about
  ...split("\\s*");

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
```diff
@@ -164,6 +169,42 @@ public void apply(final EObject element, IModificationContext context) {
 					}
 				});
 	}
+	@Fix(SPACES_IN_KEYWORD)
+	public void fixSpacesInKeyword(final Issue issue, IssueResolutionAcceptor acceptor) {
+		if (issue.getData().length > 0 && Boolean.valueOf(issue.getData()[0]))
+		acceptor.accept(issue, "Replace Keyword with multiple Keywords", "Splits a Keyword that contains spaces or tabs into multiple Keywords", NULL_QUICKFIX_IMAGE,
+				new ISemanticModification() {
```
JanKoehnlein: Don't you think a textual modification would be simpler, also for the case where you have to insert perentheses?

cdietrich:
so is it ok if we live without a quickfix?

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
cdietrich: but where to put the logic? inside the validator? calling xtextdocument.readonly?

svenefftinge:
Yes, I agree. It doesn't seem to be worth the effort.

PR #942, opened 2016-02-26
[bug 488380]  revised dirty state support

JanKoehnlein:
Possible fix. We should defer the updating of the IResourceDescription of a dirty resource when the editor is marked as clean to the moment when the appropriate change is reconciled.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=488380

PR #946, opened 2016-02-29
Bug 484465: Improved serializer performance

spoenemann:
In one specific example with partial serialization via `XtextDocument.modify(..)` the average time for applying the modification was reduced from 80.7 seconds (before this change) to 5.8 seconds (after this change). Cloning the array and hash set seems to keep the GC quite busy in some cases.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=484465

meysholdt:
:+1:

spoenemann:
The change does not improve the asymptotic complexity (5.8 seconds is still too much), so I would leave the bug open.

cdietrich:
there is one problem with this fix. we currently update to xtext 2.9.1 and patched the collectBounds as done in this pullrequest but we still get an infinite loop. the problem is that the visited set does not seem to have any influence on the recursion, since newVisit = false does not stop the recusion.

spoenemann:
Yes, the fix does not change the algorithm, but only its memory consumption. We will analyze the algorithm more closely soon.

PR #949, opened 2016-03-01
Bug 486615: completer function fetches the services from the given editor

spoenemann:
The problem: when `xtext.createServices()` is called a second time, the new completer is not applied by Ace. This change makes the completer independent of its closure variables.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=486615

PR #950, opened 2016-03-01
Bug 488635: Removed leading '/' in resource path

spoenemann:
Also avoided usage of Servlet 3.0 API method `ServletRequest.getServletContext()`.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=488635

PR #956, opened 2016-03-03
488966: provide an AbstractOutlineTest

LorenzoBettini:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=488966

This allows to easily test the outline of a DSL: it gives an indented string representation of the outline that can be easily compared with multiline strings.


svenefftinge:
Nice, thanks!

PR #972, opened 2016-03-28
[490530] Change variable type of abortOnMissingSource

kthoms:
Change-Id: I13317deeb0718c04df42341c415a73b37fbc53f6


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=490530

svenefftinge:
This will break source compatibility with earlier versions. Could you provide a version that compiles fine with the latest JDT and older versions, too.

kthoms:
I have checked that the version compiles against an older version of org.eclipse.jdt.core, and against a current JDT (which is installed in my dev env). The difference is only that the @SuppressWarnings("unchecked“) becomes obsolete with the new JDT and gets a warning. With an older JDT, the @SuppressWarnings is not marked with a warning.

In JDT code, the field is only set with a Boolean instance and checked against that. Thus I consider this fix as compatible.

svenefftinge:
Jenkins test

svenefftinge:
Thanks!

PR #974, opened 2016-03-29
[325658] Introduce AbstractGrammarElement as common base class for AbstractRule and AbstractElement

kthoms:
The PR is organized in 2 commits: The first only introduces the AbstractGrammarElement EClass, the second commit changes INode#getGrammarElement and dependent code.

Note that changing the return type of INode#getGrammarElement causes API changes. Since the returned type is a subtype of the previous one (EObject) this should be OK. I have created API filters for the places detected by API tooling.


svenefftinge:
> Note that changing the return type of INode#getGrammarElement causes API changes. Since the returned type is a subtype of the previous one (EObject) this should be OK. I have created API filters for the places detected by API tooling.

It breaks binary compatibility so it is not ok. 

In general this PR includes lots of changes that could introduce problems without a significant improvement on the other side.

kthoms:
So bug#325658 can’t be resolved? Are both changes invalid or only the second commit?

How is the proceeding when the metamodel is extended, like it was in the past?

svenefftinge:
Yes, I think we should close that old bug. It's just too much noise for the little benefit today.

> How is the proceeding when the metamodel is extended, like it was in the past?

Most changes have a relatively high risk of breaking existing code these days and we always spend a lot of time on reviewing changes to make this as unlikely as possible. So every change needs to carry it's own weight, which means it should be worth going through the time consuming process and risk.

PR #976, opened 2016-03-30
[431435] Bind generated EPackage and EFactory

kthoms:
Change-Id: I8590497bbb1f25292f185d630e4e8c9f485d7f3b

### DIFF
--- a/examples/org.eclipse.xtext.xtext.ui.examples/contents/org.eclipse.xtext.example.arithmetics/src-gen/org/eclipse/xtext/example/arithmetics/AbstractArithmeticsRuntimeModule.java
+++ b/examples/org.eclipse.xtext.xtext.ui.examples/contents/org.eclipse.xtext.example.arithmetics/src-gen/org/eclipse/xtext/example/arithmetics/AbstractArithmeticsRuntimeModule.java
@@ -13,6 +13,8 @@ import com.google.inject.name.Names;
 import java.util.Properties;
 import org.eclipse.xtext.Constants;
 import org.eclipse.xtext.IGrammarAccess;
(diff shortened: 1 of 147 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=431435

svenefftinge:
When changing the code generator we need to regenerate all the languages, too. We usually do that in a separate commit in the same PR.

kthoms:
OK. I’ll do that.

svenefftinge:
Please also regenerate xtend, xtext, xbase and the examples.

kthoms:
The bindings are only added when EMFGeneratorFragment / EMFGeneratorFragment2 is executed in the workflow. For Xtext, Xbase & Xtend the bindings are not automatically added, since they do not generate the EMF metamodel.

Is the EMFGeneratorFragment2 the right fragment or should this be handled more generic. But how to distinguish between the EMF packages "owned" by the language and those that are just used?

Should I add the bindings to the concrete runtime class for those languages manuall?

svenefftinge:
My bad. It's ok to not do that for external EPackages.

One last thing, could you add a property to the fragment to turn the generation off the bindings on, and have it off by default. The reason I'd prefer it to be off by default, is that we almost never use DI for these two types because there is no good reason to do it (especially not for EPackages) and I prefer to encourage the use of the standard EMF-way (use singletons) instead, as there is too much code that relies on that already. And we do it like that, as well.
That said its ok to have that option, if you want to have it.

kthoms:
Due to repository restructuring replaced by
https://github.com/eclipse/xtext-core/pull/75
https://github.com/eclipse/xtext-extras/pull/34

PR #980, opened 2016-04-01
[387119] Allow to deliver notifications at runtime

kthoms:
Added a class DeliverNotificationAdapter. This class can be used to switch the default value (false) at runtime when eSetDeliver is called.

The new behavior is covered by test org.eclipse.xtext.linking.lazy.LazyLinkingResourceTest.testResolveLazyCrossReferences_03()
The other tests in this test class would fail if an unexpected notification was sent.

I have changed all places where eSetDeliver is called. When eSetDeliver is called with value 'true' in the finally block, I left this unchanged. For places where the value is resetted to the previous state of eDeliver(), the DeliverNotificationAdapter#resetDeliver() method is called.

### DIFF
--- a/plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DeliverNotificationAdapter.java
+++ b/plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DeliverNotificationAdapter.java
@@ -1,0 +1,119 @@
+/*******************************************************************************
+ * Copyright (c) 2016 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.resource;
+
+import org.eclipse.emf.common.notify.Adapter;
+import org.eclipse.emf.common.notify.Notifier;
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.emf.ecore.resource.ResourceSet;
+import org.eclipse.emf.ecore.util.EContentAdapter;
+import org.eclipse.xtext.util.Pair;
+import org.eclipse.xtext.util.Tuples;
+
+/**
+ * <p>
+ * An adapter that is used to decide whether notifications should be delivered to Notifiers.
+ * </p>
+ * <p>
+ * Usage example:
+ * <pre>
+ * 
+ * DeliverNotificationAdapter.DEFAULT.setDeliver(true);
+ * 
+ * ...
+ * 
+ * Notifier notifier = ...;
+ * Resource resource = ...;
+ * 
+ * try {
+ *    DeliverNotificationAdapter.get(resource).setDeliver(notifier);
+ *    // do something with notifier
+ * } finally {
+ *    DeliverNotificationAdapter.get(resource).resetDeliver(notifier);
+ * }
+ * 
+ * @author thoms - Initial contribution and API
+ * @since 2.10
+ */
+public class DeliverNotificationAdapter extends EContentAdapter {
+	private boolean deliver = false;
+	private Pair<Notifier, Boolean> memento;
+	
+	public static DeliverNotificationAdapter DEFAULT = new DeliverNotificationAdapter();
+	
+	/**
+	 * Gets an instance for a Resource. Searches the Resource's and the ResourceSet's adapters for
+	 * an instance in their adapters lists. When not found, the {@link #DEFAULT} singleton is returned.
+	 * @return The active instance for the Resource. Will not return <code>null</code>.
+	 */
+	public static DeliverNotificationAdapter get (Resource resource) {
+		if (resource != null) {
+			for (Adapter adapter : resource.eAdapters()) {
+				if (adapter instanceof DeliverNotificationAdapter) {
+					return (DeliverNotificationAdapter) adapter;
+				}
+			}
+			return get(resource.getResourceSet());
+		} else {
+			return DEFAULT;
+		}
+	}
+	public static DeliverNotificationAdapter get (ResourceSet resourceSet) {
+		if (resourceSet != null) {
+			for (Adapter adapter : resourceSet.eAdapters()) {
+				if (adapter instanceof DeliverNotificationAdapter) {
+					return (DeliverNotificationAdapter) adapter;
+				}
+			}
+		}
+		return DEFAULT;
+	}
+
+	/**
+	 * Enable notifications.
+	 */
+	public void enable() {
+		this.deliver = true;
+	}
+
+	/**
+	 * Disable notifications.
+	 */
+	public void disable() {
+		this.deliver = false;
+	}
+	
+	/**
+	 * Notification enablement state.
+	 */
+	public boolean isDeliver() {
+		return deliver;
+	}
+
+	/**
+	 * Calls {@link Notifier#eSetDeliver(boolean)} on the {@link Notifier} with the
+	 * current enablement state value. The current state of {@link Notifier#eDeliver()}
+	 * is remembered and can be set to the previous state when {@link #resetDeliver(Notifier)}
+	 * is called.
+	 */
+	public void setDeliver (Notifier notifier) {
+		memento = Tuples.pair(notifier, notifier.eDeliver());
+		notifier.eSetDeliver(isDeliver());
+	}
+	
+	/**
+	 * Resets the {@link Notifier} notification delivery state to the value when {@link #setDeliver(Notifier)}
+	 * is called. Usually this method will be called in a finally block after {@link #setDeliver(Notifier)} on 
+	 * the same <code>Notifier</code>.
+	 */
+	public void resetDeliver (Notifier notifier) {
+		if (memento != null && memento.getFirst()==notifier) {
+			notifier.eSetDeliver(memento.getSecond());
+		}
+	}
+}
--- a/plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DeliverNotificationAdapter.java
+++ b/plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DeliverNotificationAdapter.java
@@ -9,11 +9,15 @@ package org.eclipse.xtext.resource;
 
 import org.eclipse.emf.common.notify.Adapter;
 import org.eclipse.emf.common.notify.Notifier;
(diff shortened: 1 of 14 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=387119

plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DeliverNotificationAdapter.java
```diff
@@ -0,0 +1,119 @@
...
+ *    DeliverNotificationAdapter.get(resource).setDeliver(notifier);
+ *    // do something with notifier
+ * } finally {
+ *    DeliverNotificationAdapter.get(resource).resetDeliver(notifier);
+ * }
+ * 
+ * @author thoms - Initial contribution and API
+ * @since 2.10
+ */
+public class DeliverNotificationAdapter extends EContentAdapter {
```
svenefftinge: Why a content adapter? It will pollute the whole EMF tree.

plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DeliverNotificationAdapter.java
```diff
@@ -0,0 +1,119 @@
...
+	private boolean deliver = false;
+	private Pair<Notifier, Boolean> memento;
+	
+	public static DeliverNotificationAdapter DEFAULT = new DeliverNotificationAdapter();
```
svenefftinge: Why a static global one?

plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DeliverNotificationAdapter.java
```diff
@@ -0,0 +1,119 @@
...
+	}
+	
+	/**
+	 * Resets the {@link Notifier} notification delivery state to the value when {@link #setDeliver(Notifier)}
+	 * is called. Usually this method will be called in a finally block after {@link #setDeliver(Notifier)} on 
+	 * the same <code>Notifier</code>.
+	 */
+	public void resetDeliver (Notifier notifier) {
+		if (memento != null && memento.getFirst()==notifier) {
+			notifier.eSetDeliver(memento.getSecond());
```
svenefftinge: When this adapter is use in two resourcesets in parallel the notification might not be enabled again

plugins/org.eclipse.xtext/src/org/eclipse/xtext/resource/DeliverNotificationAdapter.java
kthoms: Usually it won’t be installed anyway, it is expected that usually DEFAULT is retrieved, but a simple Adapter should do also the job. I’ll fix that, if the general approach sounds like the right way for you.

svenefftinge:
The initial bug report was about events for lazy linking, this patch treats all suppressed notifications the same. Some, like e.g. when we clear the resource set in the builder are really not interesting to anyone.

Also I don't like the design with the adapter, especially not as a content adapter. Knut proposed a load option for the resource set.

kthoms:
> The initial bug report was about events for lazy linking, this patch treats all suppressed notifications the same. Some, like e.g. when we clear the resource set in the builder are really not interesting to anyone.
> 
> 
True. The locations where this was changed disabled notifications anyway. The approach allows to get notifications for these cases as well when desired.
> Also I don't like the design with the adapter, especially not as a content adapter. Knut proposed a load option for the resource set.
> 

As sorted out in comment#14 <https://bugs.eclipse.org/bugs/show_bug.cgi?id=387119#c14>, a load option is not sufficient. The EMF-IncQuery project needs to change the behavior at runtime.

" - In EMF-IncQuery, one of our user scenarios is that the query engine is attached to an already loaded Resource / ResourceSet. So it would be nice if this option could be set to 'false' at any time, not just at load time.“

> When this adapter is use in two resourcesets in parallel the notification might not be enabled again
> 

True. Should this be bound to a ThreadLocal?

> Why a static global one?
> 

It would be sufficient to switch behavior globally.

ujhelyiz:
Thanks for the help, we will try it out as soon as possible.

@svenefftinge when we have opened the bug, we were not aware that these notifications are turned off in a lot of places, that is why we have only asked for a Resource load option. However, anybody  who turns off the notifications may invalidate our internal indexes, thus our results might become stale (furthermore, we don't have any way to detect this). There are two ways we could provide correct results: (1) the notifications are not turned off at all (for that we are fine with an opt-in solution in the sense that e.g. when we initialize our indexes, we turn off any Xtext-specific flag/singleton/etc.) or (2) we need some callback to reindex the models on changes.

We'd rather prefer option (1), and we are open to any solution we can do without knowing any specific language.

svenefftinge:
Does "on change" refer to changes in the editor or to resource changes?

bergmanngabor:
Changes in the editor.

svenefftinge:
Have you checked whether IXtextModelListener would work for you?

kthoms:
Adressed the comments from @svenefftinge
- base class is AdapterImpl
- last state is remembered in a ThreadLocal to avoid problems when updates happen in parallel
- introduced a Provider interface that is injected and used to get an adapter from a Notifier. By default it will still return just the static singleton, but without searching through the Resource's and ResourceSet's adapters

If the IXtextModelListener does not work for you, please check if my proposal does.

ujhelyiz:
@svenefftinge  For an editor, IXtextModelListener is a very promising callback.

However, querying over open editors is only one use case we support. I am not sure whether opening an Xtext model (in EMF form) in any other way (e.g. not through an XtextDocument), are there any supported operations that might turn off notifications?

@kthoms Thank you for the, we will try this out as soon as possible. I try to report back on Monday about our results.

ujhelyiz:
Sorry for being so slow to report back. We have tried out the implementation on the reproducible case with Yakindu and found that the culprit was a copied version of the original LazyLinkingResource defined by Yakindu, and not some other mechanism in Xtext. Updating only that Resource implementation fixed the single issue we have found. However, we are still evaluating whether the other changes are required for correct behaviour of our indexes in all cases

Furthermore, we have found a few issues with this pull request:

- LazyLinkingResource line 179 (finally block) was not updated (see https://github.com/eclipse/xtext/pull/980/files#diff-c360125b325d89d90edad9167e619c08R179)
- The finally block in DerivedStateAwareResourceDescriptionManager was not updated (see https://github.com/eclipse/xtext/pull/980/files#diff-9df271bde1927767805c117bc73aa43fL53)
- We are not sure whether it is intentional or not, but XtextBuilder also turns on/off notifications, but was not included in the change.
- Additionally, we have found a few methods cleaning up the Resource that also turns off notifications. This might not be problematic, but we have to test them as well.

kthoms:
> LazyLinkingResource line 179 (finally block) was not updated (see https://github.com/eclipse/xtext/pull/980/files#diff-c360125b325d89d90edad9167e619c08R179 <https://github.com/eclipse/xtext/pull/980/files#diff-c360125b325d89d90edad9167e619c08R179>)
> The finally block in DerivedStateAwareResourceDescriptionManager was not updated (see https://github.com/eclipse/xtext/pull/980/files#diff-9df271bde1927767805c117bc73aa43fL53 <https://github.com/eclipse/xtext/pull/980/files#diff-9df271bde1927767805c117bc73aa43fL53>)
The finally blocks were left by intention. The thought behind this is that it might be desired to turn events off when desired, but it is in any case desired to switch them on again in the finally block as it was implemented before. Other code locations have remembered the previous state and then set the state back (which might be false). This is where I called resetDeliver() in the finally block.

> We are not sure whether it is intentional or not, but XtextBuilder also turns on/off notifications, but was not included in the change.
This is left by intention. I thought that during a build the notifications should not be send. What’s your use case here?

> Additionally, we have found a few methods cleaning up the Resource that also turns off notifications. This might not be problematic, but we have to test them as well.
Would be great if you could test that.


Have you tried IXtextModelListener?

Does the approach with the adapter address your issue currectly and works in principle as desired?

ujhelyiz:
> The finally blocks were left by intention. The thought behind this is that it might be desired to turn events off when desired, but it is in any case desired to switch them on again in the finally block as it was implemented before. Other code locations have remembered the previous state and then set the state back (which might be false). This is where I called resetDeliver() in the finally block.

My mistake; the Adapter does have very limited internal state, so simply calling eSetDeliver(true) does everything adapter#resetDeliver is expected to do.

>>  We are not sure whether it is intentional or not, but XtextBuilder also turns on/off notifications, but was not included in the change.
> This is left by intention. I thought that during a build the notifications should not be send. What’s your use case here?

In general, EMF-IncQuery wants to support any case where model queries are useful (incremental model compiling is one of these use cases) over any EMF models. Our API might be included even in the Xtext compiler as well. However, every time notifications are turned off, there is a chance that our indexes become stale. These eSetDeliver calls are especially problematic, as we have no way to detect whether we have missed something or not.

In other words, if we want to use EMF-IncQuery/VIATRA to write the code generator/derived models, this might be problematic for us. As of now, we have never even tried to write such code using this stack because of the missing notifications, but we are interested in being able to do that.

>Have you tried IXtextModelListener? 

Not yet, but I know its internal behavior and am reasonable sure it works for editors (we already have some similar infrastructure available for derived feature handling we could trigger when the .
However, I am not sure whether such missing notifications are possible when not having an open editor; the XtextBuilder case suggests it is. In such cases, we might need additional ways to avoid suppressing these notifications as well.

> Does the approach with the adapter address your issue currectly and works in principle as desired?

Yes, we have executed quick tests, and getting the notifications makes sure that our indexes remain up-to-date. So I can confirm, this adapter clearly helps. Thank you.

svenefftinge:
I am very skeptical that getting updated on all events is what you really need, as there are many changes that happen during loading, linking and model inference which are not based on any user actions. So effectively you would need to ignore all of them. Registering for resource changes using the hooks Xtext provides is IMHO a much better idea.

ujhelyiz:
If our queries are not used during loading/linking/model inference, a resource change callback works, as then we would have some information about when to reindex the entire content of the file. However, our current architecture relies on gathering all elementary model change events, and maintains a consistent index based on this information; reindexing is possible but not our main use case.

On the other hand, what kind of hooks does Xtext provide for this reason that is not UI-dependent (such as the IXtextModelListener)?

svenefftinge:
You should also look at org.eclipse.xtext.ui.notification.IStateChangeEventBroker, which gives you change events from the builder and the dirty state manager (open editors)

spoenemann:
If the change is still relevant, please reapply it to xtext-extras and xtext-core.

PR #981, opened 2016-04-02
Port SimpleProjectWizardFragment to the new mwe2 generator

LorenzoBettini:
This ports the SimpleProjectWizardFragment to the new mwe2 generator.
In the new generator, this fragment does not rely on Xpand anymore: the created stub class for initial contents is an Xtend class.
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=490796
the fragment is part of the StandardLanguage, but it kicks in only if it's feature 'generate' is set to true, e.g.,

```
newProjectWizardForEclipse = {
	generate = true
}
```

while working on this, I also did some cleanup of the abstract classes involved in the generated wizard.

Moreover, I fixed also the old bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=338419 
Now, if the user sets an explicit path, that path is used to create the project. With this respect, I also fixed the old SimpleProjectWizardFragment (for the old mwe2 generator).


svenefftinge:
Could you please rename the property in the mwe2 file to something that clearly communicates what this is? I'd prefer 'newProjectWizardForEclipse'.

LorenzoBettini:
done:

```
newProjectWizardForEclipse = {
  generate=true
}
```

plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/wizard/AbstractProjectCreator.java
```diff
@@ -153,21 +156,16 @@ protected String getPrimaryModelFileExtension() {
         return Collections.emptyList();
     }
 
-    protected String[] getProjectNatures() {
-        return new String[] {
-        	JavaCore.NATURE_ID,
```
akosyakov: Should not we keep java nature for non plugin projects?

plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/wizard/AbstractProjectCreator.java
```diff
@@ -153,21 +156,16 @@ protected String getPrimaryModelFileExtension() {
...
-			"org.eclipse.pde.PluginNature", //$NON-NLS-1$
+	protected String[] getProjectNatures() {
+		return new String[] {
 			XtextProjectHelper.NATURE_ID
 		};
-    }
-    
-    protected String[] getBuilders() {
-    	return new String[]{
-    		JavaCore.BUILDER_ID,
```
akosyakov: Should not we keep java builder for non plugin projects?

LorenzoBettini:
@akosyakov I thought that that base class was meant for "General Project" so I removed the Java nature and builder. If I'm not wrong there's another project creator class specific for Java projects.

akosyakov:
@LorenzoBettini I've just browsed the github, it seems to be ok. The most clients override it anyway and add java nature if it is required.

PR #983, opened 2016-04-04
avoid calls to xbase.lib.IterableExtension.forEach()

meysholdt:
because in Java 8 that call is resolved to Iterable.forEach()
which causes the code in src-gen to change.

We'd like to avoid these changes when working in mixed Java 7 / 8
environments. 

Signed-off-by: Moritz Eysholdt <moritz.eysholdt@typefox.io>


meysholdt:
Jenkins test

sailingKieler:
👍

svenefftinge:
Moritz and I discussed this once more and decided to not do this. Moritz pointed out that adding another method with a different case that will show up even in Java 8 is not really a compelling idea. OTOH just deprecating will turn existing code yellow without a real reason, as it's ok for Java 7 and if a project migrates to Java 8 the same code will work fine and just generate differently (with same semantics). 

All in all it would just make it easier to live with a wrong JDK on the classpath which should be fixed rather than tolerated.

meysholdt:
true :-(

PR #984, opened 2016-04-04
Fix index guard in getPreviousHiddenPart()

lenidh:
Fixes an index guard that prevents
org.eclipse.xtext.formatting2.regionaccess.internal.NodeHidden#getPreviousHiddenPart()
from returning the first part of a hidden region.

Old behavior:
Given a hidden region consisting of two or more parts
When getPreviousHiddenPart() is called on the second part
Then null is returned

New behaviour:
Given a hidden region consisting of two or more parts
When getPreviousHiddenPart() is called on the second part
Then first part ist returned

Signed-off-by: Moritz Heindl <lenidh@gmail.com>


meysholdt:
good catch, thank you!

PR #985, opened 2016-04-05
Kth/bug438625

kthoms:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=438625

svenefftinge:
Jenkins test

kthoms:
Is the fix OK?

svenefftinge:
It's unclear what problem you want to solve, but the solution is odd, because you inject an interface of which you actually don't call a single method. The two methods which are used in this context are Eclipse specific and not part of ITraceForStorageProvider (which should be file system independent). So injecting that unused type and later cast down to the type we really want to use is strange.

If there was a reason we could introduce an interface that has only the two mentioned methods and inject that, but this only makes sense if there is a reason for it. Unless there is one I prefer to not do anything and close the bugzilla as won't fix. Another approach would be to make the injection optional and work with null, but again I'd like to understand the motivation behind that.

Please let me know if I missed something.

spoenemann:
If the change is still relevant, please reapply it to xtext-extras.

PR #988, opened 2016-04-06
484043: Add .feature and .repository projects to the wizard

LorenzoBettini:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=484043

This adds two checkboxes in the Xtext wizard, under "Eclipse project":

- Create SDK Feature Project
- Create p2 Repository Project (the lowercase for p2 is intentional, since, as far as I know, that's the right way of spelling it)

The creation of a p2 repository project implies the creation of the SDK feature project.

The p2 repository project contains a category.xml which includes the .sdk and the .sdk.source project.

If Maven/Tycho build is enabled the corresponding poms will be generated as well. In this case, the tycho plugins for generating sources for bundles and features are also added to the parent pom.

I also updated the CliWizardIntegrationTest with additional project structure expectations.


svenefftinge:
Jenkins test

svenefftinge:
@dhuebner could you please review?

dhuebner:
@svenefftinge I've started reviewing the changes.

@eclipse/eclipse-xtext For others here is a screen shot which shows the new UI.
I would like to have feedback about the new introduced group.
 
bildschirmfoto 2016-04-07 um 10 34 45

sailingKieler:
IMO the 2 additional entries can be combined to a single one, as most both artifacts are required. In the rare cases of requiring only one of them the superfluous one can be removed by the user.

sailingKieler:
Instead of an SDK Feature (containing sources as well) just a plain feature should be offered (an mentioned).

sailingKieler:
Will the new part be deactivated or vanish at all when 'Eclipse Plugin' is deactivated?

LorenzoBettini:
The SDK feature does not contain the sources as well: it is the p2 project that triggers the generation of sources. I generate the SDK feature following the old strategy of the Xtext wizard.

LorenzoBettini:
I thought about deactivating the two checkboxes when 'Eclipse Plugin' is deactivated, but I followed to current strategy not to disable anything and report errors and warnings.

dhuebner:
@sailingKieler @LorenzoBettini  I would also prefer to rename "sdk" to simply "feature". The common meaning of SDK is "sources included".  But that is not the case here.

LorenzoBettini:
I agree that SDK usually includes sources, as I said, I followed the previous strategy of the Xtext wizard. So should I rename it to simply ...mydsl.feature?

I think it would be good to keep both options available (feature and p2), but I'm also OK in keeping only one.

dhuebner:
@LorenzoBettini Yes, mydsl.features sounds good :)

dhuebner:
I've played a bit with the new wizard.
Following some enhancements for discussion:  

1. Uncheck (and/or disable) the group when "Eclipse plugin" is disabled by user.
2. Change "p2 repository" to "Update site" or "Software site". p2 repository is technically the right name, but in a wizard it may be better to choose a more common name (compare the naming in the eclipse update site dialogs) 
3. Disallow bogus configuration, e.g. following combination: 
 -  eclipse plugin - off
 + create update site - on
 + gradle build - on
4. info "The creation of a p2 repository project will automatically create an SDK feature project"  should not hide other important informations like "...An additional Gradle build will be created", so append info instead of override

LorenzoBettini:
OK, then I'll work on that in the next few days. I'll keep both checkboxes (feature and update site), right?

dhuebner:
@LorenzoBettini We discussed about labels and we would suggest to leave the "Project" part. So the end result would be:
- Create Feature
- Create Update Site

Also if user unchecks "Eclipse Plugin" the whole sub group should be disabled (greyed out)  but the checkboxes selection should remain.

LorenzoBettini:
@dhuebner I think I applied all your suggested modifications, could you please check that?

dhuebner:
@LorenzoBettini Thanks Lorenzo, I will try it out this Friday

dhuebner:
Looks good so far except of two things:
1. User selection in the new group (i.e. selecting "Create feature") get lost if user deselects "Eclipse plugin" and select it again.
2. "Create Updatesite" - on and  "Gradle Build" - on results in a combined message = good. But IMHO the severity icon (currently Info) should reflect the highes Severity of the combined messages, in this case it should be Warning.

LorenzoBettini:
I've addressed the two points you mentioned.

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.xtend
```diff
@@ -134,12 +156,16 @@ class AdvancedNewProjectPage extends WizardPage {
 			reportIssue(INFORMATION,
 				AdvancedNewProjectPage_ideaReqGradleInfo)
 		}
+		if (createP2Project.selection && !createSDKProject.selection) {
```
dhuebner: A check for "Eclipse Plugin" is disabled is missing.

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.xtend
```diff
@@ -282,6 +314,14 @@ class AdvancedNewProjectPage extends WizardPage {
 		createWebProject.selection
 	}
 
+	def boolean isCreateSdkProject() {
+		createSDKProject.selection
```
dhuebner: Also here, return false if the whole section is disabled. The same for isCreateP2Project

LorenzoBettini:
I addressed your comments; previously I was not checking for eclipse project in the isCreateSdkProject and isCreateP2Project because in the project descriptors feature and update site project require the Eclipse project descriptor so it was safe, but I agree that it's cleaner to avoid returning true if that configuration in the wizard does not make sense :)

dhuebner:

PR #992, opened 2016-04-07
[404114] Change message texts for configuration/unconfiguration of Xtext

kthoms:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=404114

plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/nature/messages.properties
```diff
@@ -1,3 +1,3 @@
-NatureAddingEditorCallback_MessageDialog_Title=Add Xtext Nature
-NatureAddingEditorCallback_MessageDialog_Msg0=Do you want to add the Xtext nature to the project '
+NatureAddingEditorCallback_MessageDialog_Title=Configure Xtext
+NatureAddingEditorCallback_MessageDialog_Msg0=Do you enable Xtext for the project '
```
spoenemann: Incomplete sentence: should be
```
Do you want to enable Xtext for the project ...
```
or, in order to be more consistent with the terms used in the context menu,
```
Do you want to convert ... to an Xtext project?
```

dhuebner:
It's a very good idea, our menu entry looks a bit alien:
bildschirmfoto 2016-04-07 um 16 50 10

plugins/org.eclipse.xtext.ui.shared/plugin.properties
```diff
@@ -2,7 +2,7 @@ OpenModelElementCommand.name=Open Model Element
 OpenModelElementCommand.description=Open the declaration of an Xtext element
 XtextBuilder.name=Xtext Project Builder
 XtextNature.name=Xtext Project Nature
-XtextNatureAddAction.label=Add Xtext Nature
-XtextNatureRemoveAction.label=Remove Xtext Nature
+XtextNatureAddAction.label=Convert to Xtext project
```
svenefftinge: uppercase P

plugins/org.eclipse.xtext.ui.shared/plugin.properties
```diff
@@ -2,7 +2,7 @@ OpenModelElementCommand.name=Open Model Element
...
+XtextNatureRemoveAction.label=Unconfigure Xtext
```
svenefftinge: Should we remove this all together? Other plugins doesn't seem to support removing as well.

plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/nature/messages.properties
```diff
@@ -1,3 +1,3 @@
-NatureAddingEditorCallback_MessageDialog_Title=Add Xtext Nature
-NatureAddingEditorCallback_MessageDialog_Msg0=Do you want to add the Xtext nature to the project '
+NatureAddingEditorCallback_MessageDialog_Title=Configure Xtext
```
svenefftinge: I think this should use the same lingo than the configure action, i.e. "Convert to Xtext Project" and Miro's suggestion.

plugins/org.eclipse.xtext.ui.shared/plugin.properties
kthoms: Maybe makes sense. It is unlikely that someone disables Xtext once it was configured. But if someone does want to do this, how should it be done? I think it is too prominent to have this option always in the "Configure" menu. How about having this in the project preference page? If so, shall I open a separate bug?

plugins/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/nature/messages.properties
```diff
@@ -1,3 +1,3 @@
-NatureAddingEditorCallback_MessageDialog_Title=Add Xtext Nature
-NatureAddingEditorCallback_MessageDialog_Msg0=Do you want to add the Xtext nature to the project '
-NatureAddingEditorCallback_MessageDialog_Msg1='?
+NatureAddingEditorCallback_MessageDialog_Title=Configure Xtext
+NatureAddingEditorCallback_MessageDialog_Msg0=Do you want to convert '
+NatureAddingEditorCallback_MessageDialog_Msg1='to an Xtext project?
```
spoenemann: Cosmetic: I think there should be a space between `'` and `to`.

PR #998, opened 2016-04-15
[491768] Change classpathScope to 'runtime'

kthoms:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=491768

svenefftinge:
Looks ok to me. @dhuebner can we merge this?

dhuebner:
@svenefftinge I'm not sure that this is a right solution, I'm even not sure that this is a problem at all. We explicitly set the scope to "compile" after we updated the exec plugin version to  > 1.1.2 where the default cp scope was changed from "runtime" to "compile". Now we are about to change it back. 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=483667

@kthoms Is it possible to create a small example maven project for the current issue?

dhuebner:
@kthoms  I've just noticed that you are still using 2.9.1 in your build. The issue you have was fixed in 2.9.2 see:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=485129

kthoms:
Issue is fixed with 2.9.2

PR #774, opened 2015-11-01
Xtend Web Editor

huibo-shi:

PR #1000, opened 2016-04-18
Fixed documentation for Web API: property artifacts not documents

vorburger:
Signed-off-by: Michael Vorburger <mike@vorburger.ch>

### DIFF
--- a/xtext-website/documentation/330_web_support.md
+++ b/xtext-website/documentation/330_web_support.md
@@ -256,7 +256,7 @@ The following functions are available, provided that the respective services are
  * `format()`
    Formats the current selection (or the whole document if the selection has zero length) and returns the formatted text.
  * `generate()`
-   Returns the document generated from the associated resource. If more than one document has been generated, an object with a property `documents` is returned, which is an array of objects with the properties `name`, `contentType`, and `content`.
+   Returns the document generated from the associated resource. If more than one document has been generated, an object with a property `artifacts` is returned, which is an array of objects with the properties `name`, `contentType`, and `content`.
  * `loadResource()`
    Loads the associated resource and returns an object with properties `fullText` and `dirty`. If the resource has been modified during the current session, the modified version is returned even if the page is reloaded.
  * `saveResource()`


vorburger:
@spoenemann as far as I can tell from having played around a little in https://github.com/eclipse/xtext/compare/master...vorburger:xtend.web for https://github.com/eclipse/xtext/pull/774 fixed https://github.com/eclipse/xtext/commit/d2c3e2833d75d62df12327057339e4c32291ede5#diff-3eb2b115e4d9d5e386dce6eec0348cc2L171, there is a small error in the documentation, which this PR rectifies - do you agree and would like to merge this?

BTW: Thank You for having written 330_web_support.md in the first place - very useful to get started understanding the (cool!) new Xtext Web support.

cdietrich:
i think this is not working since the generator service uses the artifactId to determine what generation result should be returned and if you dont specify it you get an error. thus i think the generatorService should expose compute as well to return all generated artifacts

vorburger:
@cdietrich I fixed (worked around?) the point you raise via https://github.com/eclipse/xtext/commit/d2c3e2833d75d62df12327057339e4c32291ede5#diff-5b1e78ebdf10acff0ec33045e2f3a548R14

@spoenemann iirelevant of the above, the current doc is still wrong, so I would suggest this either way.

spoenemann:
Documentation updates should be merged into maintenance. If they are merged into master, they won't become visible until the next release (2.10.0).

That part of the documentation is outdated: In one of the beta releases, the generator service returned all generated resources, while now it returns only one resource. This means that your proposed fix is also wrong. I will fix that part on the maintenance branch so we have a correct version on the website soon.

It is true that sometimes the names of the generated artifacts are not known in the client. We could add a service such as `generatorArtifactIds` that returns the names of all artifacts, which can then be fetched with the `generate` service.

spoenemann:
The fix is online: https://www.eclipse.org/Xtext/documentation/330_web_support.html

spoenemann:
Thanks for pointing to the outdated documentation!

PR #1002, opened 2016-04-18
xtend.web

vorburger:
continuation of https://github.com/eclipse/xtext/pull/774 (which should probably be closed now)

@svenefftinge & @spoenemann is this of interest to merge? OK as is (not much tested..), or need more work?

FYI For me this was more an opportunity to learn how to create an Xtend Web editor than a goal in itself..  I'd like like to apply what I've seen here in another project (BTW I'm wondering if there is any point in have any sort of org.eclipse.xtend.web.lib - as the org.eclipse.xtend.web proposed here is more of a sample application, perhaps there could be a place to keep things that are common between this example and a "real" Xtend Web IDE somewhere else - what do you think?). I've listed what's pending on https://github.com/vorburger/xtext/blob/xtend.web/web/org.eclipse.xtend.web/TODO.md, and made a short video of this on https://youtu.be/QtW6WBJztPw ..

[Xtend Editor in Web IDE](https://www.youtube.com/watch?v=QtW6WBJztPw)

### DIFF
--- a/web/org.eclipse.xtend.web/src/main/java/org/eclipse/xtend/web/XtendServiceDispatcher.xtend
+++ b/web/org.eclipse.xtend.web/src/main/java/org/eclipse/xtend/web/XtendServiceDispatcher.xtend
@@ -1,0 +1,46 @@
+package org.eclipse.xtend.web
+
+import com.google.inject.Inject
+import com.google.inject.Singleton
+import org.eclipse.xtend.web.server.java2xtend.Java2XtendService
+import org.eclipse.xtext.util.internal.Log
+import org.eclipse.xtext.web.server.IRequestData
+import org.eclipse.xtext.web.server.ISessionStore
+import org.eclipse.xtext.web.server.InvalidRequestException
+import org.eclipse.xtext.web.server.XtextServiceDispatcher
+
+@Singleton
+@Log
+class XtendServiceDispatcher extends XtextServiceDispatcher {
+	
+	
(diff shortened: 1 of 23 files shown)


web/org.eclipse.xtend.web/src/main/java/org/eclipse/xtend/web/XtendServiceDispatcher.xtend
```diff
@@ -0,0 +1,62 @@
...
+			default:
+				super.createServiceDescriptor(serviceType, context)	
+		}
+	}
+	
+	/** 
+	 * Xtend may generate several *.java from one *.xtend,
+	 * so we don't have an 'artifact' request parameter and
+	 * instead of generatorService.getArtifact() need to compute().
+	 */
```
spoenemann: I think it would be better to create a new service for this instead of changing the semantics of the original generator service.

spoenemann:
My feeling is that it's too early to release Xtend into the web, since the current web integration of Xtext provides only language support for text widgets, and not an IDE. It would be interesting to see Xtend integrated with an web IDE such as [Che](https://eclipse.org/che/). Until then I would prefer to move this contribution into a separate GitHub repo. Nevertheless, it would be cool to have this example online so people can play around with it. I already had the idea of setting up a server for language demos.

What do others think?

svenefftinge:
I agree.

vorburger:
==> https://github.com/vorburger/ch.vorburger.xtend.web

PR #1004, opened 2016-04-19
Explicit dependency on org.eclipse.rcp in ui.tests

LorenzoBettini:
If you have a ui test that needs the workbench or an editor, if you do
not have org.eclipse.osgi.compatibility.state, introduced in Luna, in
the dependencies the workbench and editors will be null and the UI tests
will fail. By adding org.eclipse.rcp feature as a dependency, then
org.eclipse.osgi.compatibility.state will be dragged in and UI tests
will work, since the workbench will be available.  Adding
org.eclipse.rcp is safe also for previous Eclipse target platforms,
while adding org.eclipse.osgi.compatibility.state only will not work in
Eclipse versions earlier than Luna.

At least, that's what I do to run UI tests that require the workbench.

### DIFF
--- a/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TestProjectDescriptor.xtend
+++ b/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TestProjectDescriptor.xtend
@@ -94,7 +94,8 @@ abstract class TestProjectDescriptor extends ProjectDescriptor {
 										<extraRequirements>
 											<!-- to get the org.eclipse.osgi.compatibility.state plugin
 											if the target platform is Luna or later.
-											(backward compatible with kepler and previous versions) -->
+											(backward compatible with kepler and previous versions)
+											see https://bugs.eclipse.org/bugs/show_bug.cgi?id=492149 -->
 											<requirement>
 												<type>eclipse-feature</type>
 												<id>org.eclipse.rcp</id>
--- a/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TestProjectDescriptor.xtend
+++ b/plugins/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TestProjectDescriptor.xtend
@@ -85,6 +85,26 @@ abstract class TestProjectDescriptor extends ProjectDescriptor {
 									«ENDIF»
 								</configuration>
 							</plugin>
+							«IF needsUiHarness»
+							<plugin>
+								<groupId>org.eclipse.tycho</groupId>
+								<artifactId>target-platform-configuration</artifactId>
+								<configuration>
+									<dependency-resolution>
+										<extraRequirements>
+											<!-- to get the org.eclipse.osgi.compatibility.state plugin
+											if the target platform is Luna or later.
+											(backward compatible with kepler and previous versions) -->
+											<requirement>
+												<type>eclipse-feature</type>
+												<id>org.eclipse.rcp</id>
+												<versionRange>0.0.0</versionRange>
+											</requirement>
+										</extraRequirements>
+									</dependency-resolution>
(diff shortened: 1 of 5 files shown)


dhuebner:
Is there a corresponding bugzilla report for this issue?
However, this is a bit strange, because we have org.eclipse.platform.feature.group in the target platform that also includes org.eclipse.rcp feature. We should find out why there are parts missing or not automatically exposed to the  tycho-surefire-plugin. If we have a solution for that, we should add it to the target platform or parent pom.xml. That will also help users, when they create an additional test project.

LorenzoBettini:
If you want I can file a bug report related to the PR.

I found this problem also mentioned on the web: http://lists.jboss.org/pipermail/jbosstools-dev/2013-December/008387.html and http://stackoverflow.com/questions/34969544/how-to-spot-and-fix-differences-between-eclipse-and-maven-build

From what I understand is related to the fragment org.eclipse.osgi.compatibility.state which is not automatically picked up, even if org.eclipse.platform.feature.group (which includes the feature org.eclipse.rcp feature) is in the TP.

That's the only solution I found to make UI tests working...

dhuebner:
@LorenzoBettini I assume that tycho-surefire plugin already adds some dependencies which are not declared in the test project and it's transitive dependencies. otherwise it wouldn't start at all

LorenzoBettini:
Yes, but only at bundle level, not at feature level, if I understand correctly.

If you want to have a feature as a dependency for tests, the only way I know is to add it in the pom as an additional dependency.

And org.eclipse.osgi.compatibility.state is a fragment so you can't add it to the required bundles in the manifest.

dhuebner:
@LorenzoBettini Keeping in mind that we are talking about a wizard and the fact, that we should count with beginners as users, I would prefer to not increase the complexity of the build system configuration, especially for maven.
If we assume that the target platform is at least Mars (2.10 will be shipped with Neon) is there a non-pom solution for that? It's also important to try keeping PDE junit launch and tycho surefire execution behave the same as good as possible. I will try to reproduce the maven build failure locally to find out more.
A bugzilla report including the maven error message would be nice for documentation reasons.

dhuebner:
@LorenzoBettini I will merge your PR as soon as we can link it to a bug/enhancement report where we can describe the issue and work on a probably better solution.

LorenzoBettini:
@dhuebner I created a bug with a project to recreate the problem: https://bugs.eclipse.org/bugs/show_bug.cgi?id=492149

My original intention was exactly to relieve the user from worrying about a strange error with an UI test which works in Eclipse but not with Maven. The solution I propose is just to add that feature in the pom and everything will work both in Eclipse and in Maven. It's similar to what you get in the runtime project's pom, e.g., 

```
<cleanupDaemonThreads>false</cleanupDaemonThreads><!-- see https://bugs.eclipse.org/bugs/show_bug.cgi?id=475098#c3 -->
```

or

```
<groupId>org.eclipse.m2e</groupId>
	<artifactId>lifecycle-mapping</artifactId>
```

As far as I know there cannot be a solution without touching the pom, due to the differences between tycho surefire plugin and PDE: you need that fragment when running Eclipse, and PDE does that automatically. tycho-surefire does not. And since that is a fragment, you have no way to have that in your test run configuration apart from a feature. And to add a feature to the tycho-surefire running application you can only add that as an explicit dependency.

dhuebner:
Jenkins xtext-runtime-tests-master-oldest

dhuebner:
Jenkins test

dhuebner:

LorenzoBettini:
Sorry, I forgot to update CliWizardIntegrationTest and the build fails. I'll update the pr later this afternoon

LorenzoBettini:
I've updated the failing test and I also added the link to the bug in the pom, following the same approach as in other parts of the poms

LorenzoBettini:

LorenzoBettini:
@dhuebner I thought I could run the Jenkins tests myself but that's not the case :)
could you please run them again?

dhuebner:

PR #1005, opened 2016-04-25
[486838] fixed bug in ErrorTreeAppendable

sailingKieler:
Signed-off-by: Christian Schneider <christian.schneider@typefox.io>

Please review!

### DIFF
--- a/plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/output/ErrorTreeAppendable.java
+++ b/plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/output/ErrorTreeAppendable.java
@@ -19,6 +19,7 @@ import org.eclipse.xtext.generator.trace.ITraceURIConverter;
 import org.eclipse.xtext.linking.lazy.LazyLinkingResource;
 import org.eclipse.xtext.linking.lazy.LazyURIEncoder;
 import org.eclipse.xtext.nodemodel.INode;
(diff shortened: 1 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=486838

spoenemann:
As discussed earlier, the current solution uses a single leaf node, which probably works only for simple types, but not for parameterized types etc.

Don't merge yet.

sailingKieler:
Now, fixed the issue by relying on `NodeModelUtil.getTokenText()`. Added a bunch of tests with plain types, inner types, and parameterized types.

plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/output/ErrorTreeAppendable.java
```diff
@@ -58,7 +59,7 @@ public TreeAppendable append(JvmType type) {
 			if(unresolvedLink != null) {
 				INode linkNode = unresolvedLink.getThird();
 				if(linkNode != null) {
-					append(linkNode.getText().trim());
+					append(NodeModelUtils.getTokenText(linkNode).trim());
```
meysholdt: calling .trim() is not necessary as getTokenText() returns already-trimmed strings.

meysholdt:
besides the unnecessary call of trim() the change looks good to me

sailingKieler:
OK, I'll drop it and merge.

PR #1007, opened 2016-04-25
[web] Added generator service options 'allArtifacts' and 'includeContent'

spoenemann:
If `allArtifacts` is set to `true`, the generator service returns all artifacts instead of a specific one. If `includeContent` is set to `false`, only the name and the content type of the requested artifact(s) is returned. That can be useful if the artifact names are not known in the client, e.g. when the JvmModelInferrer is used to generate Java types and the names of these types depend on the document contents. Then you can either fetch all generated types at once or first fetch their names and then send a dedicated request for each type.

Please review.


spoenemann:
Jenkins test

PR #1009, opened 2016-04-26
Bug 492234 - [wizard] Include sources in the generated .target file

LorenzoBettini:
The target definition file generated by the Xtext project wizard is not set to include the sources, that is, includeSource="false".

Instead, I think that sources should be included by default in a target definition file: if the developer uses the .target file for defining the target platform of his workbench, then having the sources while developing is really helpful.  I know that for some features the generated .target file includes the .sdk which already includes sources (e.g., Xtext and EMF), but that's not the case for the platform, pde and jdt.

On the other hand, including sources will not disturb the Tycho build, since the includeSource is ignored by default.  This means that sources will not have to be downloaded during the target platform resolution of the Maven/Tycho build.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=492234


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=492234

dhuebner:
Jenkins xtext-runtime-tests-master

dhuebner:

PR #1012, opened 2016-04-28
Bug 492648 - [wizard] generate environment for 32bit in the parent pom

LorenzoBettini:
The project wizard generates a parent pom with target platform environments for 64bit only.

I think it would be nice if also the 32bit version for Linux and Windows were generated as well (Mac 32bit is not supported anymore in Mars). Otherwise, if a 32bit OS user is using the Maven/Tycho build and has a ui test, that test will fail with some kind of obscure error which is not easy to track (it just says it can't find some SWT classes).

https://bugs.eclipse.org/bugs/show_bug.cgi?id=492648


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=492648

JanKoehnlein:
See comment in https://bugs.eclipse.org/bugs/show_bug.cgi?id=492648

PR #1015, opened 2016-04-29
Bug 492728 - [wizard][gradle] Project Wizard, IntelliJ Plugin, Gradle and Buildship problems

LorenzoBettini:
The Xtext project wizard does not work together with Buildship when selecting Intellij Plugin and Gradle: Buildship throws an exception because it can't find org.xtext.example.mydsl.parent/org.xtext.example.mydsl.idea/.settings/org.eclipse.buildship.core.prefs

This PR generates that file instead of gradle.prefs (which, anyway, would be converted to org.eclipse.buildship.core.prefs. The convertion would not take place for the .idea project because JavaProjectFactory.enhanceProject would fail to set the classpath, since Buildship throws the above exception).

ATTENTION: this also requires ProjectCreator.createProject to enhance the project ONLY AFTER having called all the contributors.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=492728

### DIFF
--- a/plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
+++ b/plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
@@ -42,6 +42,7 @@ import org.eclipse.xtext.xtext.wizard.AbstractFile;
 import org.eclipse.xtext.xtext.wizard.BinaryFile;
 import org.eclipse.xtext.xtext.wizard.ParentProjectDescriptor;
 import org.eclipse.xtext.xtext.wizard.ProjectDescriptor;
+import org.eclipse.xtext.xtext.wizard.ProjectLayout;
 import org.eclipse.xtext.xtext.wizard.TextFile;
 
 import com.google.common.collect.Lists;
@@ -256,6 +257,8 @@ public class XtextProjectCreator extends WorkspaceModifyOperation implements IPr
 		private String getConnectionLogicalPath() {
 			if (descriptor instanceof ParentProjectDescriptor) {
 				return "";
+			} else if (descriptor.getConfig().getProjectLayout() == ProjectLayout.FLAT) {
+				return "../" + descriptor.getConfig().getParentProject().getName();
 			} else {
 				return "..";
 			}
--- a/plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
+++ b/plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
@@ -8,7 +8,9 @@
  *******************************************************************************/
 package org.eclipse.xtext.xtext.ui.wizard.project;
 
+import static com.google.common.collect.Iterables.*;
 import static com.google.common.collect.Sets.*;
+import static org.eclipse.xtext.xbase.lib.IterableExtensions.*;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -44,6 +46,8 @@ import org.eclipse.xtext.xtext.wizard.ParentProjectDescriptor;
 import org.eclipse.xtext.xtext.wizard.ProjectDescriptor;
 import org.eclipse.xtext.xtext.wizard.TextFile;
 
+import com.google.common.base.Function;
+import com.google.common.base.Predicate;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.inject.Inject;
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=492728

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
```diff
@@ -234,25 +238,57 @@ public GradleContributor(ProjectDescriptor descriptor) {
...
 			} else {
 				return ":" + descriptor.getName();
 			}
 		}
+
+		private String getConnectionLogicalPath() {
+			if (descriptor instanceof ParentProjectDescriptor) {
+				return "";
+			} else {
+				return "..";
```
oehme: This will fail for a flat layout.

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
```diff
@@ -234,25 +238,57 @@ public GradleContributor(ProjectDescriptor descriptor) {
...
-				"	}\n" +
-				"}\n",
-			".settings/gradle.prefs");
+				"connection.arguments=\n" +
+				"connection.gradle.distribution=GRADLE_DISTRIBUTION(WRAPPER)\n" +
+				"connection.gradle.user.home=null\n" +
+				"connection.java.home=null\n" +
+				"connection.jvm.arguments=\n" +
+				"connection.project.dir=" + getConnectionLogicalPath() +"\n" +
+				"derived.resources=" + getDerivedResources() + "\n" +
```
oehme: Buildship will update these the first time the project is synchronized anyway, so no need to duplicate Buildship logic here.

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
LorenzoBettini: Right! I still have to consider the flat layout case. This could be related to your previous comment.

plugins/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
LorenzoBettini: @oehme do you mean the locations? So I can simply not put any location here? Or even completely remove the properties connection.project.dir and derived.resources?

oehme: You need all `connection` parameters, but nothing else.

LorenzoBettini:
@oehme I updated the PR according to your comments.

If flat layout is used, the generated .prefs file reads like this

```
connection.arguments=
connection.gradle.distribution=GRADLE_DISTRIBUTION(WRAPPER)
connection.gradle.user.home=null
connection.java.home=null
connection.jvm.arguments=
connection.project.dir=../org.xtext.example.mydsl.parent
eclipse.preferences.version=1
project.path=\:org.xtext.example.mydsl
```

Otherwise it is

```
connection.arguments=
connection.gradle.distribution=GRADLE_DISTRIBUTION(WRAPPER)
connection.gradle.user.home=null
connection.java.home=null
connection.jvm.arguments=
connection.project.dir=..
eclipse.preferences.version=1
project.path=\:org.xtext.example.mydsl
```

I tested manually with Buildship 1.0.15 and it does not complain anymore. But derived.resources is not generated anyway, is that expected?

oehme:
It will be generated the next time you synchronize the project. It's just information that Buildship uses internally to keep track of which resources the user marked as "derived" manually and which were marked by Buildship.

Change looks good to me 👍

LorenzoBettini:
I think I can't still run Jenkin tests myself...

svenefftinge:
Jenkins test

LorenzoBettini:
I'd go on merging this PR.

Before that, @dhuebner @svenefftinge do you think it could be worthwhile to have some Plug-in tests for the project wizard? I mean, something that makes sure that the project creation terminates and that projects are created? Such tests would have caught this integration problem with Buildship (I think the problem has been there for some time, since I also experienced that with Buildship that is in the Mars repository). To this aim, though, Buildship should be in the target platform.

oehme:
Just a quick overall note: Buildship has backwards-compatibility code, so the old format would have worked too. The actual root cause is trying to set the classpath on the project before the Buildship settings were written. They need to be there *before* JDT tries to query the Buildship classpath container.

LorenzoBettini:
@oehme Right! Sorry I forgot to mention that :)

dhuebner:
@LorenzoBettini 
> To this aim, though, Buildship should be in the target platform.
I think it's a good idea. The server tests TP (latest) as well as oomph targlets should have that. We also need m2e UI dependency in order to test the full m2e integration

LorenzoBettini:
@dhuebner in the current oomph targlet there's m2e.core, and indeed, when running a new Eclipse instance from the development workbench I could see the errors in the pom files of https://bugs.eclipse.org/bugs/show_bug.cgi?id=493120.
But buildship is not in the oomph targlet, nor in any .target file of the workspace from what I can see... it's not even in the p2 ide plugins installed in the development eclipse.

In the meantime, I'd merge this PR; then, could you please modify the targlet and .target in order to include buildship? I think we should use the latest one 1.0.15. Then, in another PR, I'll try to setup the UI tests (probably SWTBot tests).

svenefftinge:
@LorenzoBettini please merge

PR #1019, opened 2016-05-04
[488958] no validation issue for whitespaces in keyword in terminal r…

JanKoehnlein:
…ules


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=488958

spoenemann:
What about the second observation in the tracked issue:

> 2) If it was a parser rule, it should have four warnings, as line breaks in keywords are bad as well.

Currently only `" "` and `"\t"` are checked.

JanKoehnlein:
Let's leave it the way the PR is. I think people who are adding newlines in keywords know what they're doing.

PR #1024, opened 2016-05-09
why remove 'convert to xtend' ?

flyzb618:
There is a good function 'java convert to xtend' in xtend 2.8.   why remove it in 2.9 ?


svenefftinge:
the action hasn't been removed. please explain.

PR #1028, opened 2016-05-11
[434982] Escape Java Keywords inside dispatch if else cascade

cdietrich:

### DIFF
--- a/plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
+++ b/plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
@@ -21,6 +21,7 @@ import org.eclipse.xtext.common.types.JvmTypeReference;
 import org.eclipse.xtext.common.types.util.TypeReferences;
 import org.eclipse.xtext.util.Strings;
 import org.eclipse.xtext.xbase.compiler.IAppendable;
+import org.eclipse.xtext.xbase.compiler.JavaKeywords;
 import org.eclipse.xtext.xbase.compiler.Later;
 import org.eclipse.xtext.xbase.compiler.TreeAppendableUtil;
 import org.eclipse.xtext.xbase.compiler.output.ITreeAppendable;
@@ -34,6 +35,9 @@ import org.eclipse.xtext.xbase.typesystem.util.CommonTypeComputationServices;
 import com.google.inject.Inject;
 
 public class DispatchMethodCompileStrategy implements Procedures.Procedure1<ITreeAppendable> {
+	
+	@Inject
+	private JavaKeywords keywords;
 	
 	@Inject
 	private TypeReferences typeReferences;
@@ -205,11 +209,14 @@ public class DispatchMethodCompileStrategy implements Procedures.Procedure1<ITre
 		a.append(")");
 	}
 
-	/**
-	 * TODO: we assume that names from the inferred JVM model are the same in the Java code. Does that assumption
-	 * hold?
-	 */
 	protected String getVarName(JvmIdentifiableElement ex, IAppendable appendable) {
-		return ex.getSimpleName();
+		final String name = ex.getSimpleName();
+		if (name == null) {
+			return "__unknown__";
+		} else if (keywords.isJavaKeyword(name)) {
+			return name+"_";
+		} else {
+			return name;
+		}
 	}
 }
--- a/tests/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DispatchCompilerTest.xtend
+++ b/tests/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/DispatchCompilerTest.xtend
@@ -15,6 +15,46 @@ import org.junit.Test
 class DispatchCompilerTest extends AbstractXtendCompilerTest {
 	
 	/**
+	 * https://bugs.eclipse.org/bugs/show_bug.cgi?id=434982
+	 */
+	@Test
+	def testReservedJavaKeywords() {
+		assertCompilesTo('''
+			class Foo {
+				def dispatch foo(Object assert) {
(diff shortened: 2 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=434982

cdietrich:
Jenkins test

cdietrich:
can someone review this change

svenefftinge:
+1

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -205,11 +209,14 @@ protected void generateActualDispatchCall(JvmOperation dispatchOperation, JvmOpe
...
 	}
 
-	/**
-	 * TODO: we assume that names from the inferred JVM model are the same in the Java code. Does that assumption
-	 * hold?
-	 */
 	protected String getVarName(JvmIdentifiableElement ex, IAppendable appendable) {
-		return ex.getSimpleName();
+		final String name = ex.getSimpleName();
+		if (name == null) {
```
akosyakov: When does it happen? Could we provider a test for this case?

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -205,11 +209,14 @@ protected void generateActualDispatchCall(JvmOperation dispatchOperation, JvmOpe
...
+			return "__unknown__";
+		} else if (keywords.isJavaKeyword(name)) {
```
akosyakov: Could we replace else-if cascade just with if check?
if (keywords.isJavaKeyword(name)) {
	return name + '_';
}
return name;

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
```diff
@@ -205,11 +209,14 @@ protected void generateActualDispatchCall(JvmOperation dispatchOperation, JvmOpe
...
+			return name+"_";
```
akosyakov: What if there is another argument with such name, e.g. 'assert_'? Would be generated code valid?

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
cdietrich: see testReservedJavaKeywords

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
cdietrich: well we have the same behavior implemented for method parameters

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
akosyakov: I meant for the case when generated code contains '__unknown__'.

cdietrich: i just lazily copy & pasted the if else cascade

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
akosyakov: I've just tried it.
Input:
class Foo {
	def dispatch foo(Object assert, Object assert_) {}
        def dispatch foo(String assert, Object assert_) {}
}

Output:
public Object foo(final Object assert_, final Object assert__1) {
    if (assert_ instanceof String) {
      return _foo((String)assert_, **assert_**);
    } else if (assert_ != null) {
      return _foo(assert_, **assert_**);
    } else {
      throw new IllegalArgumentException("Unhandled parameter types: " +
        Arrays.<Object>asList(assert_, **assert_**).toString());
    }
  }

It seems that bogus second argument is used inside the dispatch method.

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
akosyakov: From where did you copy it? Could you point me?

plugins/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/DispatchMethodCompileStrategy.java
akosyakov: @cdietrich I reopened the associated bug, you can have a look at it if you want

cdietrich: sry i right busy now. can you have a look. i took it from
org.eclipse.xtext.xbase.compiler.JvmModelGenerator.makeJavaIdentifier(String)

akosyakov: @cdietrich it is not so urgent, you can have a look when you have time

PR #1029, opened 2016-05-11
493399: [wizard] plugin.xml in bin.includes in all generated projects

LorenzoBettini:
I'm afraid that with this PR https://github.com/eclipse/xtext/pull/1015 for bug 492728 I introduced a small bug that makes the Xtext project wizard generate a build.properties where plugin.xml is always bin.included, leading to warnings in the generated projects. Sorry about that O:)

In the PR I swapped the execution of project contributors and enhanceProject (aa766a03f3a346d25152e634cc8a5cfed8719cea), and the latter will always generate a build.properties with that inclusion, overriding what the project contributors have generated.

To fix this, while keeping the fix for bug 492728 (which requires that the buildship properties file is generated *before* enhancing the projects) I introduced early contributors, which are executed before enhancing the project.

Thus the workflow would be:

execute early contributors
enhance the project
execute all the other contributors

Task-Url: https://bugs.eclipse.org/bugs/show_bug.cgi?id=493399

### DIFF
--- a/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/ProjectFactory.java
+++ b/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/ProjectFactory.java
@@ -64,6 +64,8 @@ public class ProjectFactory {
 
 	private List<IProjectFactoryContributor> contributors;
 
+	private List<IProjectFactoryContributor> earlyContributors;
+
 	public ProjectFactory addBuilderIds(String... builderIds) {
 		if (this.builderIds == null)
 			this.builderIds = Lists.newArrayList();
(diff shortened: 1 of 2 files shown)


LorenzoBettini:
If no one has objections, I'd merge this so that this fix will be available in the next nightly

plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/ProjectFactory.java
```diff
@@ -142,25 +156,34 @@ public IProject createProject(IProgressMonitor monitor, Shell shell) {
...
+				}
+			};
 
-					@Override
-					public IFile writeToFile(CharSequence chars, String fileName) {
-						return ProjectFactory.this.writeToFile(chars, fileName, project, subMonitor);
-					}
-				};
-				for (IProjectFactoryContributor contributor : contributors) {
+			// first run early contributors...
```
akosyakov: Are comments in this method necessary? References to bugs can be found in the history.

plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/ProjectFactory.java
```diff
@@ -142,25 +156,34 @@ public IProject createProject(IProgressMonitor monitor, Shell shell) {
...
+			// in some cases it is crucial to enhance the project
+			// only after contributions, like in the case of gradle projects
+			// see https://bugs.eclipse.org/bugs/show_bug.cgi?id=492728
+			if (earlyContributors != null) {
```
akosyakov: Can we extract a protected method that accepts contributors and project to removed duplicate code here and below?

LorenzoBettini:
OK, I'll apply your comments

LorenzoBettini:
@akosyakov please have a look at https://github.com/eclipse/xtext/pull/1041

PR #1030, opened 2016-05-11
[437497] Prevent NPE in GlobalURIEditorOpener 

cdietrich:

### DIFF
--- a/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/GlobalURIEditorOpener.java
+++ b/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/GlobalURIEditorOpener.java
@@ -64,7 +64,10 @@ public class GlobalURIEditorOpener implements IURIEditorOpener {
 	public IEditorPart open(URI uri, boolean select) {
 		IResourceServiceProvider resourceServiceProvider = resourceServiceProviderRegistry.getResourceServiceProvider(uri.trimFragment());
 		if(resourceServiceProvider instanceof IResourceUIServiceProvider) {
-			return ((IResourceUIServiceProvider) resourceServiceProvider).getURIEditorOpener().open(uri, select);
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=437497

plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/GlobalURIEditorOpener.java
```diff
@@ -73,7 +76,10 @@ public IEditorPart open(URI uri, boolean select) {
...
 		IResourceServiceProvider resourceServiceProvider = resourceServiceProviderRegistry.getResourceServiceProvider(referenceOwnerURI.trimFragment());
 		if(resourceServiceProvider instanceof IResourceUIServiceProvider) {
-			return ((IResourceUIServiceProvider) resourceServiceProvider).getURIEditorOpener().open(referenceOwnerURI, reference, indexInList, select);
+			IURIEditorOpener editorOpener = ((IResourceUIServiceProvider) resourceServiceProvider).getURIEditorOpener();
+			if (editorOpener != null) {
+				return editorOpener.open(referenceOwnerURI, reference, indexInList, select);
+			}
 		}
 		return openDefaultEditor(referenceOwnerURI, reference, indexInList, select);
 	}
```
LorenzoBettini: The two modifications highlight that the lines are basically the same; would it be worth taking the opportunity to extract the logic in a method?

cdietrich: the method called is different

LorenzoBettini: Sorry, I was referring to the logic for retrieving the editorOpener (starting from the instanceof)

PR #1035, opened 2016-05-12
[generator] Fix race condition when creating directories in JavaIoFileSystemAcces

stadlerb:


plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/JavaIoFileSystemAccess.java
```diff
@@ -178,7 +178,7 @@ protected File getFile(String fileName, String outputConfigName) {
 	}
 
 	protected void createFolder(File parent) {
-		if (parent != null && !parent.exists() && !parent.mkdirs())
+		if (parent != null && !parent.mkdirs() && !parent.exists())
```
svenefftinge: I think we could just remove the exists() call, no?

svenefftinge:
Could you give a pointer or explanation of the issue you are fixing here?

plugins/org.eclipse.xtext/src/org/eclipse/xtext/generator/JavaIoFileSystemAccess.java
stadlerb: No, because mkdirs() returns false if the directory already exists.

stadlerb:
The problem is that if several JavaIoFileSystemAccess instances in parallel threads invoke the affected method simultaneously, two or more threads may (and in our case did) see the directory as non-existing during the exists() call, but the mkdirs() call will succeed only in one of the threads and return false in all others, so an exception is thrown although the directory was successfully created.

svenefftinge:
Ok, thanks

PR #1044, opened 2016-05-18
[493784] Clean JvmTypeChangeDispatcher when discarding derived state

cdietrich:
Signed-off-by: Christian Dietrich christian.dietrich@itemis.de


cdietrich:

kthoms:
Method cleanJvmTypeChangeDispatcher could be private

svenefftinge:
We need a clean up strategy for TypeResources, too. I think a general 'unload' method on JvmDeclaredTypeImplCustom that removes the registered runnables would be good. That should in turn be called by the JvmModelAssociator and the unload hook in TypeResource.

If you have a workaround I'd like to postpone this change for after 2.10.

spoenemann:
If the change is still relevant, please reapply it to xtext-extras and xtext-xtend.

cdietrich:
it is still relevant but i still dont have any new input from sven on this topic
(what is our cleanup strategy for type resources)

PR #1048, opened 2016-05-21
Typo in log message

Bananeweizen:
CLA available as michael.keppler@gmx.de, while using a different primary github mail


cdietrich:
There were some tests to be fixed too and the commit is no longer mergeable through repo split. thus i commited this as https://github.com/eclipse/xtext-core/commit/f8e2ceabc27880adb7b6685ac4cac94ad550aa8d

PR #1069, opened 2016-07-28
Workflow for milestones and issue labels

spoenemann:
As discussed in the committer meeting of July 22nd, I wrote a proposal for how we could leverage milestones and labels in the Xtext project.

I investigated the usage of labels in [vscode](https://github.com/Microsoft/vscode/labels), [jupyterlab](https://github.com/jupyter/jupyterlab/labels), and [guava](https://github.com/google/guava/labels). You can see the result in [this table](https://drive.google.com/file/d/0B7HlHK64o5nLLW5EQ2x0Ymc0QUE/view?usp=sharing). All three projects have three kinds of labels in common:

* Type (`bug`, `enhancement`, etc.)
* Status (`help wanted`, `duplicate`, etc.)
* Component (`build system`, `editor`, etc.)

My proposal follows this categorization and includes all labels that are created by GitHub for new repositories. This is meant as a starting point and can of course be changed later. In particular, I didn't propose any component labels yet, but these should be added as necessary. Furthermore, the described workflow should be evaluated after a few months of working with the new structure.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,41 @@ Please create any new issue in the most appropriate repository (see full list on
 
 Eclipse Bugzilla issues are deprecated for Xtext. However, we still accept them.
 
+### Milestones
+
+We have one milestone for each Xtext release. Fixed issues should be assigned to the milestone of the nearest release in which the fix will be included. For example, if an issue is fixed on the `maintenance` branch, the corresponding milestone is the next service release (increasing the third version component). If an issue is fixed on the `master` branch, the next major or minor release should be assigned.
+
+It is possible to assign issues to a milestone before they are fixed, which can be useful for planning. However, make sure someone will actually work on that issue when you do that!
+
+The list of issues assigned to a milestone will give a nice overview of new features and fixed bugs after a release has been done, facilitating the creation of release notes and helping users to know when a fix will be available.
+
+### Issue Labels
+
+Issue labels can serve several purposes:
+
+* Indicate the issue type: `bug`, `enhancement`, `new feature`, `question`
+* Communicate the current status of an issue: `confirmed`, `help wanted`, `needs review`
+* Communicate why an issue has not been fixed: `duplicate`, `invalid`, `wontfix`
+* Categorize issues for better overview, e.g. to assign an issue to a specific part of the software. Committers may add categorization labels as required.
+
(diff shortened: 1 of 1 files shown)


meysholdt:
I like it. One minor thing: let's avoid spaces in label names, that makes it easier to use them in query URLs. We can use an underscore instead: `help_wanted`

meysholdt:
You described the idea of the "next required action". How does it reflect in labels that the next required action is clear? is that `help_wanted` ? 

And another thought about `needs_review`. What kind of review does this request? In case of a code review, there will be a PR; I don't think we need a label that indicated that there is a PR.

CONTRIBUTING.md
```diff
@@ -6,6 +6,41 @@ Please create any new issue in the most appropriate repository (see full list on
 
 Eclipse Bugzilla issues are deprecated for Xtext. However, we still accept them.
 
+### Milestones
+
+We have one milestone for each Xtext release. Fixed issues should be assigned to the milestone of the nearest release in which the fix will be included. For example, if an issue is fixed on the `maintenance` branch, the corresponding milestone is the next service release (increasing the third version component). If an issue is fixed on the `master` branch, the next major or minor release should be assigned.
+
+It is possible to assign issues to a milestone before they are fixed, which can be useful for planning. However, make sure someone will actually work on that issue when you do that!
```
cdietrich: so we do no "release planning" for bugs or features?

spoenemann:
@meysholdt:
> You described the idea of the "next required action". How does it reflect in labels that the next required action is clear? is that `help_wanted` ? 

The next required action can be a lot of different things depending on the specific issue and its status, so I proposed to describe the action with a comment instead of creating labels for all kinds of actions. `help_wanted` can be used to attract special attention to issues where it is unclear how to proceed.

> And another thought about `needs_review`. What kind of review does this request? In case of a code review, there will be a PR; I don't think we need a label that indicated that there is a PR.

Right, I'll remove that label.

CONTRIBUTING.md
spoenemann: As written in the proposal, you can use the milestones to do some release planning. We should just avoid piling up too many issues in a milestone, but instead make sure the planning can be realized by the committers who are currently available. In the past we had the Bugzilla whiteboard for this kind of planning, and there we had too many issues that were assigned to a release but couldn't be solved, so they were pushed to the next release.

PR #1081, opened 2016-11-01
Update README.md

waynebeaton:
README must conform to the branding guidelines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Xtext
+# Eclipse Xtext™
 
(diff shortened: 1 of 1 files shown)


meysholdt:
hi Wayne! Thank you for the correction. Unfortunately, edits made via github web interface seem to use an email address `noreply@github.com` that ip-validation doesn't like.

waynebeaton:
That flashed up on my screen for a second after I submitted, and then disappeared. I assumed that something else was happening. I'll resubmit.

PR #1093, opened 2016-12-21
Link to the 2nd edition of the Xtext book

scharf:
I think it makes sense to refer to the newest version of the book.

Note: The link to the book image is a bit more complicated than the old link.

(by following the packtpub link of the image, I discovered that their ebook version is much cheaper than the ebook version I just bought at amazon....)


scharf:
Heck, I have signed the ECA and my github account is linked to my eclipse account.

kthoms:
I have the same problem. If possible, check out your branch, amend your commit with sign-off signature and force update the branch.

PR #1109, opened 2017-03-24
Replaced === with .equals

danielemaddaluno:
Replaced `===` with `.equals` for matching correctly the names in the `checkFeatureNameIsUnique`.


danielemaddaluno:
I tried the tutorial and the `===` does not return an error (which is the purpose of the tutorial), the `.equals` does.

PR #1130, opened 2017-10-18
fix typo

EDDYMENS:


kthoms:
Thanks! Unfortunately the contribution mechanism through the GitHub web interface does not work with Eclipse policies yet. For contributions you would need to sign an [ECA](https://www.eclipse.org/legal/ECA.php) and sign off the commit, and push it from remote.

For this change I hop it is OK that I have done it directly in source and pushed it, so I will close this PR.

PR #1148, opened 2017-11-13
Add info about signing off commits to quality guide

brodo:
The change:
```
9. [Sign off](https://git-scm.com/docs/git-commit#git-commit--s) your commits using the same email address you are using for your Eclipse account.
```


kthoms:
This is actually not a quality criteria. I think this is better added to the contribution guide.

brodo:
Maybe like this?

kthoms:
Thanks!

PR #1188, opened 2018-04-18
Update community.html

pragmatic-modeling:
Hi,
Is it possible to include pxDoc to the list of tools using xtext ?

Thanks in advance,
Kind Regards
Romain Bernard


cdietrich:
@pragmatic-modeling yes it is possible

Issue #1194, opened 2018-05-04
Maintain Website / GDPR / Rules / etc


cdietrich:
things to evaluate

- can we port our jekyll themes to the new eclipse quicksilver (did not find any template/doc yet)
- do we fullfill the header footer whatever stuff the eclipse fqn wants to have
- do we need tagmanager and analytics
- do we need embedded twitter
- do we need embedded vimeo (xtend)
- can we avoid 3rd party cookies (twitter/vimeo)
- can we avoid 1st party cookies (eclipse)
- do we need this inform user on cookies thingy
- can the fork me on github thing done without pulling third party stuff directly

If we are not compliant at May 23nd then websites will be shut down by the fdn

cdietrich:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=534384

szarnekow:
See PR #1196

cdietrich:
we now have a better version of the websites live and will see what the audit by the foundation will give

kthoms:
[GDPR Audit results](https://docs.google.com/spreadsheets/d/1ZICnSgT1H6fAlbLITsHbSTK6gNzP0TuP_l72h48ofD0/edit?usp=sharing) for Xtext & Xtend websites are available. Compliance is approved.

<img width="1421" alt="screenshot 93" src="https://user-images.githubusercontent.com/265597/40033953-0d596cd6-57fb-11e8-87f3-25992c441676.png">

PR #1196, opened 2018-05-07
[#1194] Maintain Website / GDPR / Rules / etc

max-leuthaeuser:
- do we fullfill the header footer whatever stuff the eclipse fqn wants to have? yes, looks good. Exactly like on the eclipse.org main page.
 - do we need tagmanager and analytics? tagmanager JS stuff has thrown exceptions - I removed it for now, may be reverted if needed
 - do we need embedded twitter? No really, as it was only used to display inline tweets - I removed all twitter stuff and added simple links to @xtext and @xtendlang instead
 - do we need this inform user on cookies thingy? I added a Cookie Consent thing using cookieconsent.insites.com; TODO: what to do if the users does not want to opt-in? See TODOs in e.g., xtext-website/_includes/head.html
 - can the fork me on github thing done without pulling third party stuff directly? I removed the link to the external IMG source and used a local one instead. (xtext-website/images/forkme_right_white_ffffff.png)


xtext-website/_includes/footer.html
```diff
@@ -19,22 +19,13 @@ <h3 class="footer-links-header">Quick Links</h3>
...
-                fjs.parentNode.insertBefore(js,fjs);
-              }
-            }(document,"script","twitter-wjs");
-            </script>
-
+            <!-- Social Media Links -->
+            <h3 class="footer-links-header"">Social Media</h3>
+            <ul class="footer-links clearfix">
+              <li>
+                <a href="https://twitter.com/xtext"><img src="{{ site.baseurl }}/images/Twitter-bird-darkgray.png" class="img-responsive" style="margin-right: 5px;height: 1em;" alt="Twitter icon">@xtent on Twitter</a>
```
cdietrich: typo: Xtext on Twitter

cdietrich:
xtend
icon missing?

cdietrich:
it looks like the both now behave differently regarding analytics. do i overlook someth

max-leuthaeuser:
I did not change anything w.r.t. Google analytics, only removed the tagmanager thing.

cdietrich:
then in one it is commented out so we shoild do that in both

max-leuthaeuser:
Ok, fixed that with a22b43b.

cdietrich:
looks better now. am still not sure if/how we can avoid the vimeo cookie.
do we have means to use a own simple videoplayer in times of html5 ?

cdietrich:
do i see it correct; besides vimeo we dont have cookies anymore so we wont even need the consent plugin?

max-leuthaeuser:
I will remove the iframes embedding the vimeo videos with the next commit. Instead the html video element will be utilized showing those video (270p, 360p, and 720p - served locally).
After that I will investigate all remaining cookies.

cdietrich:
nice :+1: was playing around with the video this afternoon but i missed the `embedded-video` css class so the video had the wrong size

PR #1213, opened 2018-05-18
[411549]  Release Notes Xtend: Multi-catch and multi-type guard included

HBuender:
A release note for multi-catch and multi-type guard support for Xtend


cdietrich:
Could you add something to the documentation itself too

kthoms:
Good idea. `xtend-website/documentation/202_xtend_classes_members.md` would be the right place.

HBuender:
@kthoms thanks for the hint. I am on it :)

HBuender:
After taking a closer look I would rather put the enhancements in xtend-website/documentation/203_xtend_expressions.md since switch and catch are explained there as well.

xtend-website/documentation/203_xtend_expressions.md
```diff
@@ -919,6 +932,17 @@ val name = try {
   }
 ```
 
+Handling multiple exceptions in a single catch block can be realized using a *multi-catch clause*. The *multi-catch clause* can handle an arbitrary number of exceptions separated by `|`. Within the catch block, the intersection exception type of the exceptions caught by the *multi-catch clause* is available for further processing.
+
+```xtend
+def readFile(String path){
```
kthoms: please use whitespaces around the brackets

xtend-website/documentation/203_xtend_expressions.md
```diff
@@ -919,6 +932,17 @@ val name = try {
   }
 ```
 
+Handling multiple exceptions in a single catch block can be realized using a *multi-catch clause*. The *multi-catch clause* can handle an arbitrary number of exceptions separated by `|`. Within the catch block, the intersection exception type of the exceptions caught by the *multi-catch clause* is available for further processing.
```
kthoms: Please add a note that error is raised when one exception covers another, e.g. `NullPointerException | RuntimeException`

kthoms:
Thx @HBuender

PR #1260, opened 2018-08-07
Mention IF...ELSE...ENDIF expression in documentation

damschristian:
Hey guys,
I use XTend for generating HTML Code and (i may be stupid) didnt know for a longer time, that i can use ELSE in the template expression.
I always used IF like

```
IF condition1
  'code'
ENDIF

IF !condition1
  'code'
[ENDIF]
```

I found the documentation on [xtendtech](http://www.xtendtech.com/ivr/help/Script%20Language%20Reference/Conditional%20Statements/conditional%20statements.htm)

Maybe a mention of the alternative uses of IF template expressions may help other (stupid) users of xtend

### DIFF
--- a/xtend-website/documentation/203_xtend_expressions.md
+++ b/xtend-website/documentation/203_xtend_expressions.md
@@ -1044,7 +1044,24 @@ def someHTML(Paragraph p) '''
   </html>
 '''
 ```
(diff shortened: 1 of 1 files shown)


cdietrich:
hi why is this a pull request? and what is your actual question

cdietrich:
created https://github.com/eclipse/xtext/pull/1261

PR #1306, opened 2018-10-01
Update index.html

audinue:
Add quick fix for the fork button.

PR #1312, opened 2018-10-12
Update index.html

audinue:
This is the updated pull request of #1306


cdietrich:
thank you very mich @audinue

PR #1320, opened 2018-10-17
Fixed typos

tivervac:


CONTRIBUTING.md
```diff
@@ -48,11 +48,11 @@ Before starting the setup process, make sure that you have both an [Eclipse](htt
 When you run your freshly installed Eclipse IDE for the first time it will clone the relevant repositories and automatically set up your workspace with a number of projects.
 
 ## Contribute via Fork
-You need a [GitHub](https://github.com/) and a [Eclipse](https://accounts.eclipse.org/user/register) account plus a signed [Eclipse Contributor Agreement](https://accounts.eclipse.org/user/login?destination=user/eca).
+You need a [GitHub](https://github.com/join) and an [Eclipse](https://accounts.eclipse.org/user/register) account for which you signed the [Eclipse Contributor Agreement](https://accounts.eclipse.org/user/login?destination=user/eca).
```
cdietrich: should the first a be an an too?

tivervac: No, you use `a` when the next letter is a vowel and `an` when the next letter is a consonant.
e.g. `an apple` because the a in apple is a vowel, `a bucket` because the b in bucket is a consonant

PR #1356, opened 2018-12-03
Proofreading the releasenotes

tivervac:


tivervac:
Note that the Xtext release notes mentions that >100 issues were fixed, yet only 89 were fixed. Are we still waiting on 10 issues or is this just wrong?

Also note that the github queries do not include xtext-umbrella and that both xtend and xtext releasenotes use the same queries.

cdietrich:
i assume the corrent number still needs to be calculcated and the text adjusted

PR #1380, opened 2019-01-14
Update 103_domainmodelnextsteps.md

schubmat:
If the project in the runtime environment does not have an Xtext nature, code generation won't work afaik. At least that's what happened to me recently.

### DIFF
--- a/xtext-website/documentation/103_domainmodelnextsteps.md
+++ b/xtext-website/documentation/103_domainmodelnextsteps.md
@@ -133,7 +133,7 @@ First of all, locate the file *DomainmodelGenerator.xtend* in the package *org.e
     '''
     ```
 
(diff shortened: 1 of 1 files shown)


schubmat:
@cdietrich Done. I guess it takes some time before it becomes updated?

cdietrich:
i dont see a signoff in your commit

cdietrich:
(edits via github web are not supported by the eclipse foundation right how. they are working on it.

schubmat:
@cdietrich initially I only signed the ECA on the eclipse page. 

I had a little trouble with my mail aliases / gitconfig, as I use them on multiple git servers. I'm sorry for the mess with all those commits. I tried to signoff the file, but the check still fails. I guess I'll have to do some reading on how to properly signoff, or this sufficient already?

cdietrich:
It cannot find a eca for any of the emails in signoff. And there should be only one commit. Maybe it's best to create a new branch with a new commit and or. Sorry for the inconveniences and that the foundation makes contribution so hard

PR #1381, opened 2019-01-15
If the project in the runtime environment does not have an Xtext nature, code generation won't work afaik. At least that's what happened to me recently.

schubmat:
Signed-off-by: Mathias Schubanz <M.Schubanz@b-tu.de>

@cdietrich If I login here 
* https://www.eclipse.org/legal/ECA.php

and use the validation mechanism provided on the top right of my user page for `M.Schubanz@b-tu.de`, it tells me that there is a valid ECA. Do I need to sign the ECA somewhere else?


schubmat:
Now the check turns out positive. I guess problems are solved now.

cdietrich:
thx @schubmat

PR #1401, opened 2019-02-22
Change Jenkins URL

DennisRoeck:


szarnekow:
Is this smth that should be merged eventually?

kthoms:
Yes, but we'll do this with a separate PR. This was a test for how the new ECA tool feels when submitting through the web editor.

PR #1453, opened 2019-05-14
bumped xtext versions in docs

cdietrich:


kthoms:
Could you add a hint to "Preparing Milestones and Releases" in Builds.md that this file has to be edited for a release? Or shall we document all the potential files to edit with version updates elsewhere?

cdietrich:
i dont have an overview. and it depends on the usecase where and what needs to be adopted

PR #1559, opened 2019-10-12
Change the link of the event to point to the Xtext talks directly.

miklossy:


miklossy:
@cdietrich @szarnekow @kthoms What do you think about this change?

miklossy:
I updated the PR so that the link points to the `Xtext talks` directly.

Issue #1140, opened 2017-10-21
Document Multi Quick Fixes


cdietrich:

miklossy:
Fixed in 2.22.

PR #1723, opened 2020-04-02
[#1140] Add Multi Quick Fixes documentation to the "Eclipse Support".

miklossy:

### DIFF
--- a/xtext-website/documentation/310_eclipse_support.md
+++ b/xtext-website/documentation/310_eclipse_support.md
@@ -208,6 +208,21 @@ public void fixName(final Issue issue,
 }
 ```
 
+### Multi-Quickfixes
+Xtext also provides a Quickfix API for the definition of multi-quickfixes. Its implementation is similar to the quick fix above, but you need to call `acceptor.acceptMulti()` instead of `acceptor.accept()`.
+```java
+@Fix(IssueCodes.INVALID_FEATURE_NAME)
+public void fixFeatureName(Issue issue,
+                           IssueResolutionAcceptor acceptor) {
+  acceptor.acceptMulti(issue,
+    "Uncapitalize name", // label
+    "Uncapitalize name", // description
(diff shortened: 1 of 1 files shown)


nbhusare:
LGTM

xtext-website/documentation/310_eclipse_support.md
```diff
@@ -208,6 +208,21 @@ public void fixName(final Issue issue,
 }
 ```
 
+### Multi-Quickfixes
+Xtext also provides a Quickfix API for the definition of multi-quickfixes. It's implementation is quite similar to the quick fix above, but you need to call `acceptor.acceptMulti()` instead of `acceptor.accept()`.
```
cdietrich, line 212: Its

xtext-website/documentation/310_eclipse_support.md
```diff
@@ -208,6 +208,21 @@ public void fixName(final Issue issue,
...
+```java
+@Fix(IssueCodes.INVALID_FEATURE_NAME)
+public void fixFeatureName(Issue issue,
```
cdietrich, line 215: would be nice to have an example for the second api too

cdietrich: https://www.typefox.io/blog/xtext-2-13-0-released-semantic-editing-made-easy

xtext-website/documentation/310_eclipse_support.md
miklossy: Fixed.

xtext-website/documentation/310_eclipse_support.md
miklossy: To keep the documentation as small as possible, I think it is enough to have only one example that makes the user aware of that feature. Once they start using that Multi-Quickfix API, they will search for further documentation/examples and will find the blog post above.

Issue #1095, opened 2017-01-08
Table contents of "Documentation" misses a lot of pages on Xtext home page


scharf:
I was searching for the "seven languages xtext" and [google found it](https://eclipse.org/Xtext/documentation/201_sevenlang_introduction.html), but I wonder why this is not part of the documentation index.

Looking into the HTML for the table of contents, it seems that  quite a few parts are commented out:

Is that intentional or by accident?

```html
<ul id="nav-outline" style="margin-left: 0px;">
	<li class="nav-part">Getting Started</li>
	<!--li><a href="101_five_minutes.html">5 Minutes Tutorial</a>
	</li-->
	<li><a href="102_domainmodelwalkthrough.html">15 Minutes Tutorial</a>
	</li>
	<li><a href="103_domainmodelnextsteps.html">15 Minutes Tutorial - Extended</a>
	</li>
	<li><a href="104_jvmdomainmodel.html">Five simple steps to your JVM language</a>
	</li>
	<!--li class="nav-part">Seven JVM Languages Built With Xbase</li>
	<li><a href="201_sevenlang_introduction.html">Introduction</a>
	</li>
	<li><a href="202_scripting.html">Scripting Language</a>
	</li>
	<li><a href="203_builddsl.html">Build Language</a>
	</li>
	<li><a href="204_mongodb.html">DSL for MongoDB</a>
	</li>
	<li><a href="205_guice.html">DSL for Guice</a>
	</li>
	<li><a href="206_httprouting.html">Http Routing Language</a>
	</li>
	<li><a href="207_template.html">Template Language</a>
	</li>
	<li><a href="208_tortoise.html">Little Tortoise</a>
	</li-->
	<li class="nav-part">Reference Documentation</li>
	<li><a href="301_grammarlanguage.html">The Grammar Language</a>
	</li>
	<li><a href="302_configuration.html">Configuration</a>
	</li>
	<li><a href="303_runtime_concepts.html">Language Implementation</a>
	</li>
	<li><a href="305_xbase.html">Integration with Java</a>
	</li>
	<!--li><a href="306_mwe2.html">MWE2</a>
	</li-->
	<!--li><a href="307_special_languages.html">Typical Language Configurations</a>
	</li-->
	<li><a href="308_emf_integration.html">Integration with EMF</a>
	</li>
	<li><a href="310_eclipse_support.html">Eclipse Support</a>
	</li>
	<!--li><a href="320_idea_support.html">IntelliJ IDEA Support</a>
[...]

PR #1724, opened 2020-04-03
[#1095] Remove MWE2 documentation from the Xtext repository.

miklossy:
- The MWE2 documentation has already been transfered to the MWE2 GitHub
repository with the PR https://github.com/eclipse/mwe/pull/140


szarnekow:
Generally speaking it's bad practice to simply delete resources without establishing a permanent redirect. What is the reasoning behind this?

miklossy:
Thanks for the hint @szarnekow ! The intention behind this was to clean up the Xtext documentation as much as possible. I do not prefer a permanent redirect since it would pollute the Xtext documentation folder. I would even remove the https://github.com/eclipse/xtext/blob/website-published/xtext-website/documentation/304_ide_concepts.html and https://github.com/eclipse/xtext/blob/website-published/xtext-website/documentation/310_web_integration.html pages to further clean up the Xtext documentation repository.

However, we should mention that in the release notes. People who will try to access to moved webpages can get in touch with us for further assistance.

@cdietrich What is your opinion about this?

cdietrich:
i dont care. i would have kept it in the first place. i also dont know how to do a redirect

cdietrich:
but it does not matter anyway as the old code whas just an html sitting around and referenced nowhere.
we did also do zero redirects for documentation moves and removals in the past

cdietrich:
maybe we have place a html with a link to the new MD

Issue #1689, opened 2020-02-21
[2020-06] create release notes


cdietrich:
https://github.com/eclipse/xtext/blob/website-master/xtend-website/_posts/releasenotes/2020-06-02-version-2-22-0.md
https://github.com/eclipse/xtext/blob/website-master/xtext-website/_posts/releasenotes/2020-06-02-version-2-22-0.md

miklossy:
I went through all the closed issues and pull request and think that the following topics maybe worth mentioning in the release notes:

**Xtext:**
Global unique name validation: https://github.com/eclipse/xtext-core/issues/1466Responsible: @szarnekow

**Xtend:**
https://github.com/eclipse/xtext-xtend/issues/1014 Responsible: @cdietrich

cdietrich:
hmm i dont want to make fools of ourselves

miklossy:
No problem, fine for me if we leave that out.

cdietrich:
done

PR #1763, opened 2020-05-04
[#1689] Add "Maven Tycho update to 1.7.0" to the release notes.

miklossy:


cdietrich:
this is imho an irrelevant change

miklossy:
Since it was quite a lot of work, I would be happy to see that in the release notes.

kthoms:
I think it is worth mentioning with a bullet point in the updates list

PR #1810, opened 2020-06-26
Add Blang, this time with email address in sync with EF

alexandrebouchard:


alexandrebouchard:
Will try one more time with the sign off message in commit.

PR #1868, opened 2020-11-02
added hint on quickfix code actions to rel notes

cdietrich:
added hint on quickfix code actions to rel notes


miklossy (changes requested):
Can we add links to the interface / class ?

cdietrich:
> Can we add links to the interface / class ?

i dont think this is a good idea if it will change sooner or later anyway.

miklossy (approved):
Nice!

PR #2161, opened 2023-03-28
Restore p2 contents without tests

### LINKED ISSUE #2159: Some test bundles are published in the p2 site
Due to the merging of all the split repos' p2 repositories.
I'll fix that ASAP.

LorenzoBettini:
Closes #2159


org.eclipse.xtext.p2repository/category.xml
```diff
@@ -1,336 +1,5 @@
...
-	<bundle id="org.eclipse.xtext.xtext.ui.examples" version="0.0.0">
-		<category name="Xtext" />
-	</bundle>
-	<bundle id="org.eclipse.xtext.xtext.ui.examples.source" version="0.0.0">
-		<category name="Xtext" />
-	</bundle>
-	<bundle id="org.eclipse.xtext.xtext.ui.graph" version="0.0.0">
-		<category name="Xtext" />
-	</bundle>
-	<bundle id="org.eclipse.xtext.xtext.ui.graph.source" version="0.0.0">
```
cdietrich, line 330: is all contained in sdk feature i assume?

org.eclipse.xtext.p2repository/category.xml
```diff
@@ -1,336 +1,5 @@
...
-		<category name="Xtext" />
-	</bundle>
-
    <feature id="org.eclipse.xtext.sdk" version="0.0.0">
```
cdietrich, line 3: is the 0.0.0 intentional. before we use x.x.xqualifer here too

cdietrich (approved):
looks good to me. lets see what build says

org.eclipse.xtext.p2repository/category.xml
LorenzoBettini: @cdietrich yes. That's the situation that was in umbrella so if it was included before, it is included now

org.eclipse.xtext.p2repository/category.xml
LorenzoBettini: Yes, it is intentional: specifying the version for our bundles/features doesn't make much sense and would force us to keep them consistent.

Issue #2141, opened 2023-02-27
prepare r202303target platforms and update latest to 202306/4.28 I-builds


cdietrich:
prepare r202303 (target platforms and update latest to 202306/4.28 I-builds (done)

- wizard 
- xtext-eclipse 
- xtext-xtend 
- oomph

cdietrich:
TODO: wizard and oomph are still outstanding

LorenzoBettini:
> TODO: wizard and oomph are still outstanding

what do you mean?

Oomph has been updated to consume the nightly repo.

The wizard should not require any change, or am I missing something?

cdietrich:
@LorenzoBettini  oomoh misses a 2023-06 target platform with 4.28 I-builds

cdietrich:
TODO: unclear if we can keep running builds with j11 or need a switch in TargetPlatformProject

cdietrich:
last outstanding: oomph

cdietrich:
Oomph is done too
Todo: change default once m1 is out

PR #2162, opened 2023-03-29
[#2141] add 2023-06 to Xtext.setup

cdietrich:
[#2141] add 2023-06 to Xtext.setup

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -67,6 +67,9 @@
     <annotation
         source="http://www.eclipse.org/oomph/setup/GlobalVariable"/>
     <choice
+        value="2023-06"
+        label="Eclipse 2023-06 - 4.28"/>
+    <choice
         value="2023-03"
         label="Eclipse 2023-03 - 4.27"/>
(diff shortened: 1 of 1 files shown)


releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -616,7 +619,40 @@
...
+        <repository
+            url="${p2.emf}"/>
+        <repository
+            url="${p2.gef-classic}"/>
+        <repository
+            url="${p2.xpand}"/>
+        <repository
+            url="${p2.buildship}"/>
+        <repository
+            url="${p2.webtools}/2023-03"/>
```
LorenzoBettini, line 655: Is that intentional? Because there's no 2023-06 for that?

LorenzoBettini (approved):
Looks good and it's working locally.
One additional note: the default TP is still 2023-03 in Oomph, should we change it to 2023-06?

releng/org.eclipse.xtext.contributor/Xtext.setup
cdietrich: i wanted to wait for m1 out to change it

PR #2165, opened 2023-03-30
rely on the getters like 'getScopeProvider()' instead of the corresponding fields in 'IdeContentProposalProvider'

sailingKieler:
this way the employed services can be replaced in a subclass


LorenzoBettini (approved):

dhuebner (approved):
Nice!

sailingKieler:
@LorenzoBettini thanks for merging, didn't find the time last week.

PR #2168, opened 2023-03-31
Rename releng projects in Domainmodel example

### LINKED ISSUE #2166: Rename releng projects in Domainmodel example
Currently, the projects are called `feature`, `tp`, and `repository` so in the Eclipse workspace it is not clear where they belong.
I'd rename them according to the standard `org.eclipse.xtext.example.domainmodel.feature`, etc.

LorenzoBettini:
Note that I renamed only the projects' names, not the directories, so from the Maven point of view nothing changes. Moreover, we only zip the releng project with its contents in the examples, so the ant script for zipping the examples doesn't need to be changed

Closes #2166


cdietrich (approved):
Lgtm. We will see if it breaks Xtext reference projects integ test. Can you also check if the wizard zipping or packaging or config in the Xtext.Xtext.ui.examples project need adaptation

cdietrich:
Add I overlooked you already checked the 2nd part

LorenzoBettini:
@cdietrich thanks! I'll wait for the build to finish and then I'll merge

cdietrich:
no issues in integ tests

PR #2170, opened 2023-04-03
fix jetty inconsistency

cdietrich:
fix jetty inconsistency


LorenzoBettini (commented):
Maybe we could use a property in the POM and refer to the property in the dependencies.
This should improve consistency.

cdietrich:
the inconsistency is not between poms or within pom
but the bump from gradle got missing

LorenzoBettini:
> the inconsistency is not between poms or within pom but the bump from gradle got missing

Yes, I understand, but for future updates, changing only the value of the property will be easier. That's what I was proposing.

cdietrich:
feel free to provide a 2nd pr

LorenzoBettini:
@cdietrich if you don't mind I added a commit on top of this PR

PR #2172, opened 2023-04-03
fix orbit inconsistency

cdietrich:
fix orbit inconsistency


LorenzoBettini (approved):
I also tested it on my machine

PR #2185, opened 2023-04-04
redo junit 5.9.2 update from gradle dev-bom

cdietrich:
redo junit 5.9.2 update from gradle dev-bom

@LorenzoBettini the vintage engine does no longer be in use. do i see that correct?


cdietrich:
@LorenzoBettini  what do you think?

LorenzoBettini:
@cdietrich sorry I forgot to answer: what do you mean that the vintage engine is no longer be in use?

cdietrich:
@LorenzoBettini  before the vintage engine was needed to run junit 4 and 5 test
is this now solved differently?

LorenzoBettini:
> @LorenzoBettini before the vintage engine was needed to run junit 4 and 5 test is this now solved differently?

I didn't know things had changed... however, there was no vintage dependency in the BOM even before, right?

cdietrich:
yes and no the 5.9.2 was not working without it in gradle

cdietrich:
@LorenzoBettini it was in the bom. this is why i stumbled on it.
but if it is  not needed we can continue without it

cdietrich:
@LorenzoBettini pushed. will wait for another build to run

LorenzoBettini (approved):

Issue #2174, opened 2023-04-03
Update to 2023-03 platform in maven/bom


cdietrich:
Update to 2023-03 platform in maven/bom
see also https://github.com/eclipse/xtext/issues/2134

to my knowledge this is the last Java 11 version

cdietrich:
move to 2023-09

LorenzoBettini:
Hadn't this already been approved for merging?

cdietrich:
yes but i did not find enhough time to test with all my maven and gradle test project in time for m3

Issue #2158, opened 2023-03-27
Cleanup Maven plugins ITs


LorenzoBettini:
@cdietrich @szarnekow this line https://github.com/eclipse/xtext/blob/main/org.eclipse.xtend.maven.plugin/src/test/resources/it/pom.xml#L13 scares me a lot!

It means that we run the IT projects using the currently released version of Xtext, NOT the currently developed version. We might release a new version of xtend-maven-plugin that does not work with the same released version of Xtext.

I'd apply the same strategy as in https://github.com/xtext/xtext-monorepo/pull/17 so that the integration tests are using the same artifacts of the current sources, OK?

cdietrich:
@LorenzoBettini this is new to me in xtext-xtend. we do all the "local repo foo"
to run the tests against the plugin we build right before,
picking up the stuff built by the gradle plugin.

```
		<profile>
			<id>useJenkinsSnapshots</id>
			<repositories>
				<repository>
					<id>local-repo</id>
					<url>${gradleMavenRepo}</url>
				</repository>
```

```
<plugin>
				<artifactId>maven-antrun-plugin</artifactId>
				<executions>
					<execution>
						<id>create-jar</id>
						<phase>pre-integration-test</phase>
						<configuration>
							<skip>${maven.test.skip}</skip>
							<target>
								<property name="src.jar"
									value="${project.build.directory}/${project.build.finalName}.jar" />
								<property name="it.jar.name" value="${project.build.finalName}-it" />
								<property name="it.jar.directory"
									value="${project.build.directory}/${it.jar.name}" />
								<unzip src="${src.jar}" dest="${it.jar.directory}" />
								<replaceregexp file="${it.jar.directory}/META-INF/maven/plugin.xml"
									match="&lt;version&gt;(.*)&lt;/version&gt;" replace="&lt;version&gt;IT-SNAPSHOT&lt;/version&gt;"
									byline="false" />
								<jar destfile="${project.build.directory}/${it.jar.name}.jar"
									basedir="${it.jar.directory}" />
							</target>
						</configuration>
						<goals>
							<goal>run</goal>
						</goals>
					</execution>
					<execution>
						<id>touch-jar</id>
						<phase>pre-integration-test</phase>
						<configuration>
							<target>
								<property name="it.jar.name" value="${project.build.finalName}-it" />
								<touch file="${project.build.directory}/${it.jar.name}.jar" />
							</target>
						</configuration>
						<goals>
							<goal>run</goal>
[...]

PR #2187, opened 2023-04-04
[#2174] bump platform maven artifacts to 2023-03

cdietrich:
[#2174] bump platform maven artifacts to 2023-03

### DIFF
--- a/org.eclipse.xtend.core/pom.xml
+++ b/org.eclipse.xtend.core/pom.xml
@@ -35,15 +35,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.eclipse.jdt</groupId>
+			<artifactId>ecj</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.jdt</groupId>
 			<artifactId>org.eclipse.jdt.core</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.eclipse.jdt</groupId>
-			<artifactId>org.eclipse.jdt.compiler.apt</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.eclipse.jdt</groupId>
-			<artifactId>org.eclipse.jdt.compiler.tool</artifactId>
 		</dependency>
 	</dependencies>
 
--- a/org.eclipse.xtext.builder.standalone/pom.xml
+++ b/org.eclipse.xtext.builder.standalone/pom.xml
@@ -32,15 +32,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.eclipse.jdt</groupId>
+			<artifactId>ecj</artifactId>
+		</dependency>
+		<dependency>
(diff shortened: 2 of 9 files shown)


cdietrich:
looks like this needs fix for
https://github.com/eclipse/xtext/issues/2158
or something the like as the bom does not seems to be used correctly

cdietrich:
unfortunately the log does not seems to have enhough information  on this

cdietrich:
checking locally the maven plugin tests indeed seem to take the dev bom from sonatype snapshots and not the local build one

LorenzoBettini:
@cdietrich on my branch https://github.com/eclipse/xtext/tree/lb_2158_cleanup-maven-plugins-ITs I've cleaned up the Xtend maven plugin ITs. I now create a new branch merging my branch with yours and see whether together the tests pass.

However, if you bump the versions in the BOM, this means that the minimal version of Eclipse deps is 2023-03 while the minimal target platform is 2022-03, is that right?

cdietrich:
No In tycho we stay with old version minimal as we did this diverging the last 7 years

LorenzoBettini:
> No In tycho we stay with old version minimal as we did this diverging the last 7 years

What looks strange to me is that the minimal version in the BOM is higher than the minimum target platform. I seemed to recall it was the other way around or that the minimum target platform corresponded to the minimal version in the BOM.

cdietrich:
Minimal target was oxygen until recently although it has no java 11 support but we had

LorenzoBettini:
@cdietrich merging the two branches https://github.com/eclipse/xtext/tree/lb_2158_cd_2174 and running Xtend Maven ITs work on my machine. I pushed the branch, let's see if also the CI is happy

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/node/TracedAccessorsProcessor.java
```diff
@@ -63,6 +63,7 @@ public void doTransform(MutableClassDeclaration annotatedClass, @Extension Trans
 					if (TracedAccessorsProcessor.TYPES_WITH_GOOD_TO_STRING
 							.contains(rt.getType().getSimpleName().toLowerCase())) {
 						annotatedClass.addMethod(tracerName(getter), it -> {
+							context.setPrimarySourceElement(it, context.getPrimarySourceElement(annotatedClass));
```
szarnekow, line 66: Is this intentionally part of this merge request?

cdietrich: No

LorenzoBettini (approved):

LorenzoBettini:
@cdietrich I see you rebased this branch, so I'm removing the temporary experiment https://github.com/eclipse/xtext/tree/lb_2158_cd_2174

Issue #2189, opened 2023-04-05
offending plugin.xml for L/org.eclipse.xtext.common.types.tests/plugin.xml


cdietrich:
offending plugin.xml for L/org.eclipse.xtext.common.types.tests/plugin.xml

```
org.xml.sax.SAXException: Done
	at org.eclipse.emf.ecore.plugin.EcorePlugin$1Handler.startElement(EcorePlugin.java:383)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.startElement(AbstractSAXParser.java:518)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLNSDocumentScannerImpl.scanStartElement(XMLNSDocumentScannerImpl.java:374)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLNSDocumentScannerImpl$NSContentDriver.scanRootElementHook(XMLNSDocumentScannerImpl.java:613)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl$FragmentContentDriver.next(XMLDocumentFragmentScannerImpl.java:3079)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLDocumentScannerImpl$PrologDriver.next(XMLDocumentScannerImpl.java:836)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLDocumentScannerImpl.next(XMLDocumentScannerImpl.java:605)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLNSDocumentScannerImpl.next(XMLNSDocumentScannerImpl.java:112)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanDocument(XMLDocumentFragmentScannerImpl.java:542)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:889)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:825)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.XMLParser.parse(XMLParser.java:141)
	at java.xml/com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.parse(AbstractSAXParser.java:1224)
[...]

PR #2190, opened 2023-04-05
[#2189] workaround for missing manifest

cdietrich:
[#2189] workaround for missing manifest

### DIFF
--- a/org.eclipse.xtext.common.types.tests/plugin.xml
+++ b/org.eclipse.xtext.common.types.tests/plugin.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
 <?eclipse version="3.0"?>
-<plugin>
+<plugin
+   id="org.eclipse.xtext.common.types.tests"
+>
 	<extension point="org.eclipse.emf.ecore.generated_package">
 		<package 
 			uri = "http://www.eclipse.org/2010/xtext/ui/common/types/tests/ContentAssist"


cdietrich:
@LorenzoBettini  is missing MANIFEST intentional?

cdietrich:
alternative solution might be to configure workflow not to reproduce plugin.xml and no manifest
or create manifest

cdietrich:
i also cannot run workflow
in /Users/cdietrich/xtext-main2/git/xtext/org.eclipse.xtext.common.types.tests
as mwe launch is not on classpath

@LorenzoBettini  what is the pattern to solve that?

szarnekow:
Would we need a build-time dependency for that?

cdietrich:
alternative approach

cdietrich:
see 2nd pr. as it is a test project i think it does not matter.
but i did not test any other workflow if they run

LorenzoBettini:
> @LorenzoBettini is missing MANIFEST intentional?

Yes: that is a JAR packaging type project, with no PDE nature, whose tests are meant to be executed with maven surefire, NOT with tycho surefire.

LorenzoBettini:
> i also cannot run workflow in /Users/cdietrich/xtext-main2/git/xtext/org.eclipse.xtext.common.types.tests as mwe launch is not on classpath
> 
> @LorenzoBettini what is the pattern to solve that? do we have this pattern somewhere else?

I think you figured it out here https://github.com/eclipse/xtext/pull/2191/files#diff-3b952e6aec6d75eff0a82a180180594f5f02f9b36a0312532364296e52080ba2R46 right?

As I said above, this is a pure Maven JAR project.

PR #2191, opened 2023-04-05
[#2189] workaround for missing manifest

cdietrich:
[#2189] workaround for missing manifest


cdietrich:
could replace https://github.com/eclipse/xtext/pull/2190

LorenzoBettini:
@cdietrich if this removes the log spammed with exceptions and tests are still green, I'd prefer this solution

org.eclipse.xtext.common.types.tests/pom.xml
```diff
@@ -43,6 +43,10 @@
 			<groupId>org.eclipse.emf</groupId>
 			<artifactId>org.eclipse.emf.mwe2.language</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.emf</groupId>
+			<artifactId>org.eclipse.emf.mwe2.launch</artifactId>
```
szarnekow, line 48: Should this be in scope compile?

cdietrich: as it is a test project i am not sure (is there a compileonly?)
this is why i asked @LorenzoBettini for a hint/pattern

szarnekow: TIL there is no compileonly in mvn scopes.

szarnekow (approved):
LGTM

org.eclipse.xtext.common.types.tests/pom.xml
LorenzoBettini: I think that's because there's no good reason to have a dependency only available at compile time ;) at least that's always been my understanding

PR #2205, opened 2023-04-08
Release of Milestone M0

### LINKED ISSUE #2152: Create M0 milestone
with publish to maven central
and std location of update site
+ adapt ref projects to use it

LorenzoBettini:
Closes #2152 

This also updates the use of xtend-maven-plugin version 2.31.0.M0


jenkins/milestone-deploy/Jenkinsfile
```diff
@@ -0,0 +1,61 @@
...
+      ''')
+  }
+
+  environment {
+    DOWNLOAD_AREA = '/home/data/httpd/download.eclipse.org/modeling/tmf/xtext'
+    REPOSITORY_PATH="${DOWNLOAD_AREA}/updates/milestones"
+  }
+
+  tools {
+    maven "apache-maven-3.8.6"
```
cdietrich, line 23: is it intentional not to have the latest 3.8 version here?

jenkins/milestone-deploy/Jenkinsfile
```diff
@@ -0,0 +1,61 @@
...
+    jdk "temurin-jdk17-latest"
```
cdietrich, line 24: do we intentionally not run with java 11 (and 202203 tp)

org.eclipse.xtext.full.releng/pom.xml
```diff
@@ -1,15 +1,16 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-	<parent>
-		<groupId>org.eclipse.xtext</groupId>
-		<artifactId>xtext-parent</artifactId>
-		<version>2.31.0-SNAPSHOT</version>
-	</parent>
+	<groupId>org.eclipse.xtext</groupId>
```
cdietrich, line 3: what is the background behind this change?

org.eclipse.xtext.maven.releng/pom.xml
```diff
@@ -1,15 +1,16 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-	<parent>
-		<groupId>org.eclipse.xtext</groupId>
-		<artifactId>xtext-parent</artifactId>
-		<version>2.31.0-SNAPSHOT</version>
-	</parent>
+	<groupId>org.eclipse.xtext</groupId>
```
cdietrich, line 3: what is the background behind this change?

org.eclipse.xtext.p2repository/pom.xml
```diff
@@ -13,6 +13,8 @@
 	<properties>
 		<root-dir>${basedir}/..</root-dir>
 		<zipFinalName>${project.artifactId}-${qualifiedVersion}</zipFinalName>
+		<!-- Used for tycho-eclipserun-plugin -->
+		<eclipse-version>2023-03</eclipse-version>
```
cdietrich, line 17: is this why we need j17?

jenkins/milestone-deploy/Jenkinsfile
LorenzoBettini: as I said in other PRs, that's what I found in the original Jenkinsfile, including the one we use for standard CI. I didn't dare to touch that for the moment. If we want to go with 3.8, maybe we should do that first in the standard Jenkinsfile so that we won't have bad surprise upon release.

jenkins/milestone-deploy/Jenkinsfile
LorenzoBettini: I thought that it would be better to build against the same TP that is part of simrel.
Do you think it's better to use the minimal TP also when releasing?

org.eclipse.xtext.full.releng/pom.xml
LorenzoBettini: The releng projects act only as aggregators: we don't publish them, and there's no need to make them inherit from the parent POM. This allows updating versions without problems due to possible cycles since they are independent of the versions of our actual POMs.

org.eclipse.xtext.maven.releng/pom.xml
LorenzoBettini: same as above

org.eclipse.xtext.p2repository/pom.xml
LorenzoBettini: That version is used only to collect the p2 bundles and features to use the p2 composite ant tasks through `tycho-eclipserun-plugin` (see a few lines below). I preferred to use the latest version of the platform, but it might also work with previous versions.

jenkins/milestone-deploy/Jenkinsfile
cdietrich: i assume this is again a thig of developer discipline. @szarnekow what do you think

szarnekow: I'm ok with it for M0. I'd like to review the involved shell-scripts together with the Jenkinsfile in more details though.

szarnekow (approved):

LorenzoBettini:
I forgot to mention that this script scripts/bump-minor-version.sh is not involved in the release process: it's a utility script for bumping the minor version after a final release.

@szarnekow if something is not clear in the release process, please let me know and I'll document it further

cdietrich:
@szarnekow @LorenzoBettini  what is the state of this pr?

LorenzoBettini:
@cdietrich sorry I hadn't seen Sebastian's approval. We can proceed.

PR #2206, opened 2023-04-11
[Setup] Simplify WorkingSets task and exclude unrelated projects

HannesWell:


LorenzoBettini:
@HannesWell, could you please summarize what the result would be?

HannesWell:
> @HannesWell, could you please summarize what the result would be?

Of course. Basically the result should be the same, if you only have xtext in your project. In case you also have other projects not related to xtext they should not end up in any of the xtext working sets (that the reason for the extra location rule). Instead they either go into the `Other Projects` working set or their own one if defined.

cdietrich:
Ha i was wondering why the other projects ws was not working anymore

LorenzoBettini:
@HannesWell I tested it locally, but it does not seem to work (you see my two external projects did not end in "Other Projects"):

image

LorenzoBettini:
@HannesWell , I've updated your setup to handle also the euler example and to handle `xtext-parent`, note that for the latter it's slightly more complex because it's at the root of the Git repository so it must be handled separately

image

I'll create a new PR built on top of yours

HannesWell:
> @HannesWell I tested it locally, but it does not seem to work (you see my two external projects did not end in "Other Projects"):

Can you check in the the preferences if the dynamic working sets are created properly and looks like

grafik


> @HannesWell , I've updated your setup to handle also the euler example and to handle `xtext-parent`, note that for the latter it's slightly more complex because it's at the root of the Git repository so it must be handled separately

OK, great.

> I'll create a new PR built on top of yours

You can also click below the headline of this PR on my fork's branch and edit the file in the GH web-interface and commit it from there to this PRs branch. Then you don't have to create an own PR.

If you want I can also include your extra commit to this PR.

HannesWell:
> @HannesWell I haven't thought about that and I've already created another #2217
> 
> your initial commit is there anyway ;)

Personally I just like to keep changes in one PR if reasonably possible so that the discussion is not distributed over multiple PRs and in this case have less merge commits :) Therefore I occasionally use that as reviewer as well, if non trivial changes are needed and its harder to describe to the creator what to do.

So how should we proceed, should I cherry-pick your changes into this PR or do you want to continue with #2217 or do you want to merge this first, then rebase #2217 and then merge that?

LorenzoBettini:
@HannesWell honestly I prefer to have all the commits in the history ;)

I guess the discussion already continued on https://github.com/eclipse/xtext/pull/2217 so I'd proceed there and close this one.

HannesWell:
> @HannesWell honestly I prefer to have all the commits in the history ;)

Yes every commit should definitely be attributed to the original author, but besides that I personally don't think that the process of a change does not have to be preserved in the commit history. So rebasing, amending, squashing commits etc. is IMHO fine to have a clean history where each commit is a self contained change (ideally minimal). And personally I don't like merge commits as they make the history harder to understand. For example in platform merge commits are forbidden and that works IMHO quite well and gives a nice and linear history.
> 
> I guess the discussion already continued on #2217 so I'd proceed there and close this one.

OK, then I close this one.

Issue #2197, opened 2023-04-06
Wizard: Update to Junit 5.9.2


cdietrich:
Wizard: Update to Junit 5.9.2

laeubi:
You are probably using a JUnit5 classpath container in your project, then the import is superfluous and might lead to some confusion here when you mix older tycho versions with newer eclipse releases.

cdietrich:
thx that seems to help

LorenzoBettini:
IIRC, the JUnit5 classpath container is required to run tests from Eclipse.
I also seem to remember that also the import package was needed... at least some time ago, maybe things have changed now.

cdietrich:
meh. asusme we have to add vintage enginge for eclipse :(

cdietrich:
runnings works anyway but you have tlo select junit 4 for junit 4 tests without viontage engine

laeubi:
As said, if you mix old stuff with new stuff then there is some chance that something gets mixed up.
We now have a page describing all setups and a demos, if anything is missing it would be good to add an example and/or test: https://tycho.eclipseprojects.io/doc/master/TestingBundles.html

cdietrich:
possible workaround for testing project: create a separate testing.tests project instead of having tests in the bundle

LorenzoBettini:
> possible workaround for testing project: create a separate testing.tests project instead of having tests in the bundle

but here I read: https://tycho.eclipseprojects.io/doc/master/TestingBundles.html

> You use the eclipse-test-plugin packaging, and with those your plugin must only contain test-classes and they will be executed automatically as part of the integration-test phase of your build. This aproach is not recommended for new designs.

However, is it working now?
Did you keep the JUnit5 classpath container and removed the imported bundles?

cdietrich:
i have to remove the container to get tyhco work
i need to select junit4 in the run as dialog as i had to do in gradle without vintage enging
junit5 tests seems to work ootb without container

LorenzoBettini:
> i have to remove the container to get tyhco work i need to select junit4 in the run as dialog as i had to do in gradle without vintage enging junit5 tests seems to work ootb without container
> 
> removing the import package is not an option as it wont work in shipped code

mh... sad to hear that... which project are we talking about?
Could you please save two launch configurations in the project? One for running JUnit4 tests and one for JUnit5 tests. Otherwise, it might not be clear how to run tests from Eclipse.

laeubi:
> but here I read

[...]

PR #2208, opened 2023-04-12
[WIP] use junit 5.9.2 in targets

cdietrich:
use junit 5.9.2 in targets

### DIFF
--- a/org.eclipse.xtext.tests/.classpath
+++ b/org.eclipse.xtext.tests/.classpath
@@ -6,7 +6,6 @@
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5"/>
 	<classpathentry exported="true" kind="lib" path="lib/simple.jar"/>
 	<classpathentry kind="src" output="target/test-classes" path="src">
 		<attributes>
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -56,7 +56,7 @@ Apache Commons Lang (2.4)
 * License: Apache License, 2.0
 
 
-Apache Log4j (1.2.17)
+Apache Log4j (1.2.24)
 
 * License: Apache License 2.0
 
@@ -64,7 +64,7 @@ args4j (2.33)
 
 * License: MIT License
 
-ASM (9.4)
+ASM (9.5)
 
 * License: BSD-3-Clause
 
@@ -86,7 +86,7 @@ Google Guice / Inject Core API (5.0.1)
 
 * License: Apache License, 2.0
 
-Gradle Wrapper (7.4)
+Gradle Wrapper (8.x)
 
 * License: Apache License, 2.0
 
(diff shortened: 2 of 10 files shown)


org.eclipse.xtext.tests/.classpath
```diff
@@ -6,7 +6,6 @@
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5"/>
```
cdietrich, line 9: with the container tycho fails to resolve
without you have to select junit4 for the junit4 tests in run as dialogs. did not find a nice solution yet
see discussion in https://github.com/eclipse/xtext/issues/2197

cdietrich: it is also about running single tests. currently in maintainance mode this is 50% of my usecases.
maybe there is also another option to make the vintage enginge visble to jdt.

cdietrich: i dont know. 
the problem is we mix tests and a plugin.
and we need the imports for the plugin

LorenzoBettini: What do you mean by "we need the imports for the plugin"?

cdietrich: xtext.testing has tests and is a real plugin

LorenzoBettini: Yes, I meant that: I seem to understand from the discussion that the JUnit container in the classpath should be enough (without import packages in the MANIFEST) to make everyone happy (running tests from Eclipse and Tycho).

cdietrich: will it work if you use xtext.testing in mydsl.tests then?i assume no

cdietrich: it is a plugin that is used in other plugins

cdietrich: yes this is what i mean. you have to selct junit 4 in the dialog.

i have doubts a plugin can have code in osgi that is not on classpath. but i am not an osgi expert.
am also not sure how to properly test this.

HannesWell:
The class org.eclipse.xtend.ide.buildpath.Junit5LibClasspathAdder also maintains a list of junit bundles that probably has to be adapted too.
I wonder if it would be possible to make that more dynamic and use the BSN of those bundles that contain the corresponding Junit classes?

cdietrich:
> Junit5LibClasspathAdder

this is a good question. it all is a  mess and was even more with older eclipse and xtend versions.
wonder if we can use package imports there.
or a mix of both

cdietrich:
@LorenzoBettini @HannesWell  any idea why this is not working

Bildschirmfoto 2023-04-13 um 12 50 16

LorenzoBettini:
@cdietrich I can reproduce it, but I have no idea why that's not working... even because that bundle is in the target platform...

cdietrich:
happens also in fresh eclipse.
created https://github.com/eclipse-pde/eclipse.pde/issues/574 to clarify

cdietrich:
will close this one and create a rebased new on 2morrow

PR #2211, opened 2023-04-12
Build only external PRs

LorenzoBettini:
This is the only solution I've found to build PRs only from external forks.
This is not directly supported by GitHub Actions, so we have to use this "trick".
The job for our internal PRs is started, but all the steps are skipped.

### DIFF
--- a/.github/workflows/external-pr.yml
+++ b/.github/workflows/external-pr.yml
@@ -1,9 +1,16 @@
-name: Build
+name: Build External PR
 
 on:
-  push:
+  pull_request:
     paths-ignore:
       - '**.md'
+
+env:
+  # Base for the PR
+  BASE_REPO: ${{ github.event.pull_request.base.repo.clone_url }}
+
+  # Repository of PR
+  PR_HEAD_REPO: ${{ github.event.pull_request.head.repo.clone_url }}
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


cdietrich:
hmmm then i have to think about this again.

LorenzoBettini:
if that's what you wanted then I guess there's no other solution.
This solution is not bad after all: for our PRs the jobs will end pretty soon (but of course we still need to wait for them to start).

.github/workflows/external-pr.yml
```diff
@@ -0,0 +1,56 @@
...
+          distribution: 'temurin'
+        if: ${{ env.BASE_REPO != env.PR_HEAD_REPO }}
+
+      - name: 'Cache Maven packages'
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml', '**/*.target') }}
+          restore-keys: ${{ runner.os }}-m2
+        if: ${{ env.BASE_REPO != env.PR_HEAD_REPO }}
```
cdietrich, line 41: wonder if this can be moved up to build
https://blog.mergify.com/running-github-actions-only-on-certain-pull-requests/

LorenzoBettini: Unfortunately not, because of this https://github.com/actions/runner/issues/1189

LorenzoBettini:
@cdietrich I'm merging this. In the future I'll experiment with some alternative approaches.

PR #2213, opened 2023-04-12
[Setup] Reference xtext target-files in Xtext's Oomph Setup directly

HannesWell:
As suggested in https://github.com/eclipse/xtext/pull/2207 this PR aims to simplify Xtext's Oomph setup by using the xtext-target files used in the Tycho/Maven build also in the Target-Platform of the development workspace.

This avoids the need to replicate and maintain all different target platforms as pure Oomph Targlets.
Use one TargletTask for each xtext-target file (the targlet-task contains the target-file as 'composedTarget') and use a corresponding filter to activate that task if the 'eclipse.target.platform' variable has a corresponding value.

Only for the o.e.cbi.p2repo.analyzers dependencies a always active TargletTask with corresponding Targlet is created, because they are not contained in the target-definition specified by the target-files. That targlet is merged with the one activated for the selected Eclipse target. Without the p2repo-analyzers dependencies or if they where contained in the target-files Xtext could use a simple TargetPlatformTask (if it is not desired to merge the TP with the Targlets defined by other Projects in the workspace).

I checked for all supported TPs that the right target-file is selected and tested the workspace with the latest (aka 2023-06), the 2023-03 and oldest 2022-03 target and all of them resulted in an error free workspace.

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -88,77 +88,14 @@
   </setupTask>
   <setupTask
       xsi:type="setup:VariableTask"
-      name="p2.xtext"
-      value="https://download.eclipse.org/modeling/tmf/xtext/updates/nightly"
-      label="Xtext p2 Respository URL"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      name="p2.orbit"
-      value="https://download.eclipse.org/oomph/simrel-orbit/2023-06"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      name="p2.mwe2"
-      value="https://download.eclipse.org/modeling/emft/mwe/updates/nightly/"
-      label=""/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      name="p2.emf"
-      value="https://download.eclipse.org/modeling/emf/emf/builds/release/latest"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      name="p2.lsp4j"
-      value="https://download.eclipse.org/lsp4j/updates/releases/0.20.1"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
       type="URI"
-      name="p2.xpand"
-      value="https://download.eclipse.org/modeling/m2t/xpand/updates/nightly/"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
       name="p2.antlr-gen"
       value="https://download.itemis.com/updates/releases/2.1.1"/>
   <setupTask
       xsi:type="setup:VariableTask"
       type="URI"
-      name="p2.draw2d"
-      value="https://download.eclipse.org/tools/gef/updates/legacy/releases"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      type="URI"
-      name="p2.buildship"
-      value="https://download.eclipse.org/buildship/updates/e411/releases/3.x"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      name="p2.m2e"
-      value="https://download.eclipse.org/technology/m2e/releases"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      name="p2.webtools"
-      value="https://download.eclipse.org/webtools/repository/">
-    <description>required by m2e</description>
-  </setupTask>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      type="URI"
-      name="p2.swtbot"
-      value="https://download.eclipse.org/technology/swtbot/releases/latest"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      type="URI"
       name="p2.cbi.analyzer"
       value="https://download.eclipse.org/cbi/updates/analyzers/4.7"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
-      name="p2.egit"
-      value="https://download.eclipse.org/egit/updates-nightly"/>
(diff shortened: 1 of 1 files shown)


cdietrich:
need to try this out

cdietrich:
am getting
```
        ERROR: org.eclipse.equinox.p2.engine code=0 session context was:(profile=TARGET_DEFINITION:resource:/xtext-parent/xtext-latest.target, phase=org.eclipse.equinox.internal.p2.engine.phases.Collect, operand=, action=).
        ERROR: org.eclipse.equinox.p2.artifact.repository code=0 No repository found containing: osgi.bundle,de.itemis.xtext.antlr,2.1.1.v201405091103
        ERROR: org.eclipse.equinox.p2.artifact.repository code=0 No repository found containing: org.eclipse.update.feature,de.itemis.xtext.antlr.feature,2.1.1.v201405091103
        ERROR: org.eclipse.equinox.p2.artifact.repository code=0 No repository found containing: osgi.bundle,org.antlr.generator,3.2.0.v201405091103
```
for latest.
this is an optional dep that should not be resolved

cdietrich:
for some reason we had the itemis update site in the targlet before
@szarnekow @kthoms do you know/remember why?
comfort?

HannesWell:
> for latest.
> this is an optional dep that should not be resolved

For me it worked well, but I have only tested it in my existing xtext workspace. Will try it again in a clean one.



> for some reason we had the itemis update site in the targlet before

To me it also looked like there were some inconsistencies between the target files and the workspace TP defined by the Oomph targlets. But because the target-files are used in the CI I assumed they should work and did not made a detailed comparison.

szarnekow:
>  because of the antlr generator so that it's not get downloaded if we run mwe2 workflows.

Yes, exactly. It's for convenience and dev experience.

cdietrich:
maybe we can add it to the cbi section or create a sibling to it?

HannesWell:
Can you tell me which IUs to install from the items update-site?
Looking into the repo I find the following ones, but AFAICT no one of them was previously in the workspace TP/Targlet created by the Oomph setup:

grafik

cdietrich:
de.itemis.xtext.antlr it requires the 2nd plugin

HannesWell:
> de.itemis.xtext.antlr it requires the 2nd plugin

Thanks. Added the `de.itemis.xtext.antlr.feature.feature.group`, since if installed with a feature the experience in the Installation details is usually better.

LorenzoBettini:
I'm not sure I understand the last change (that's why I don't like push-forces in PRs ;) ;)

isn't the `de.itemis.xtext.antlr.feature.feature.group` already in our target platforms?
Why does it need a special treatment Oomph setup?

cdietrich:
@LorenzoBettini  no its not. and it also should not.
but it was in old oomph

LorenzoBettini:
> @LorenzoBettini no its not. and it also should not. but it was in old oomph
> 
> (we dont generate workflows in build)

But it is in the target platform: https://github.com/eclipse/xtext/blob/main/xtext-latest.target#L42

cdietrich:
hmm this is also new to me.
so dont care. adding the requirement would be fine then.

LorenzoBettini:
> hmm this is also new to me. so dont care. adding the requirement would be fine then.

I don't understand what you mean: if it's in the target platform file it's useless to add it in Oomph, or am I missing something?

However, I tested this PR locally and it works for me! :)

One small note we should be aware of: by using `.target` files we rely on PDE, and this means that we'll have the

`.metadata/.plugins/org.eclipse.pde.core/.bundle_pool/`

in the workspace, containing features and bundles, while with standard Oomph targlets that directory was not populated and the Oomph bundle pool was completely reused.

I'm fine with that. Even though some disk space is used, I prefer to have only single target files to maintain.

LorenzoBettini (approved):

cdietrich (approved):

LorenzoBettini:
let's hear from @szarnekow

HannesWell:
> I'm not sure I understand the last change (that's why I don't like push-forces in PRs ;) ;)
>

If you click `Compare` next to the force-push notification then you can see the diff between the two commits.

grafik

After a rebase this also includes the changes from the commit one rebased on, but in case different files are affect you can see quite quickly the change within the PR.
In general I understand that concern and agree that it is not ideal, but in my opnionen git histories with (many) merge commits are also hard to understand. And assuming that the history is read more often afterwards than during PR review I find the choice of not using merges better.

If you want to make it easier to compare one could simply first rebase and force-push and the do additional changes or the other way round. This way you get two compares.

> > @LorenzoBettini no its not. and it also should not. but it was in old oomph
> > (we dont generate workflows in build)
> 
> But it is in the target platform: https://github.com/eclipse/xtext/blob/main/xtext-latest.target#L42

That's right and as far as I can tell not even one itemis plugin or feature was in the Oomph-Setup TP.
Looking at the git blame Christian added it to the older target-files in 554c1c737ec65f12a21c7c6d9c508c54384dc28f.

Any way, since it is now in the Oomph-setup created TP of the workspace, should it then be removed from the TP (in a subsequent PR)?

szarnekow (approved):

HannesWell:
> One small note we should be aware of: by using `.target` files we rely on PDE, and this means that we'll have the
> 
> `.metadata/.plugins/org.eclipse.pde.core/.bundle_pool/`
> 
> in the workspace, containing features and bundles, while with standard Oomph targlets that directory was not populated and the Oomph bundle pool was completely reused.

Just created https://github.com/eclipse-pde/eclipse.pde/issues/577 to track any future effort in the area.

PR #2217, opened 2023-04-13
[Setup] Simplify WorkingSets task and exclude unrelated projects

LorenzoBettini:
Replaces https://github.com/eclipse/xtext/pull/2206

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -829,6 +829,9 @@
             <operand
                 xsi:type="predicates:NamePredicate"
                 pattern=".*tutorial*"/>
+            <operand
+                xsi:type="predicates:NamePredicate"
+                pattern=".*euler"/>
           </operand>
         </predicate>
       </workingSet>
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -785,24 +785,25 @@
         xsi:type="setup.workingsets:WorkingSetTask">
       <workingSet
           name="Features"
-          id="">
+          id="ws-features">
         <predicate
             xsi:type="predicates:AndPredicate">
           <operand
-              xsi:type="predicates:LocationPredicate"
-              pattern="${git.clone.xtext.location|path}/.*"/>
+              xsi:type="predicates:RepositoryPredicate"
+              project="org.eclipse.xtext"/>
           <operand
               xsi:type="predicates:NaturePredicate"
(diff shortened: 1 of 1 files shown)


LorenzoBettini:
I intentionally cancelled the GitHub Actions jobs since they're useless here

cdietrich:
what do i need to do to get other ws working?

LorenzoBettini:
> what do i need to do to get other ws working?

sorry, I don't understand what you mean...

cdietrich:
i did setup on your branch
i created a new project
its not added to other projects working set

LorenzoBettini:
you have to re-execute the working set task after you imported new projects

cdietrich:
hm still have the feeling this works differently on old master
there they automatically end in the others

LorenzoBettini:
> hm still have the feeling this works differently on old master there they automatically end in the others

@merks has something changed in that respect concerning Oomph?
We have working sets for our projects, so external projects should end-up in "Other Projects". This works if the other projects are already imported before running the task. If we import another project (not related to Xtext), this does not end-up automatically in "Other Projects": we have to run the working set task again.

LorenzoBettini:
@merks I hadn't noticed we have several working set tasks! I thought it was one with nested specifications!
My bad: this comes from the repo merge.
Concerning "same repository as", I can safely use that because the task is run after project imports, right?
I can simply choose a project that I know will always be there.

merks:
Yes, the working sets are applied to actual IProject/IResource objects after these things are imported.   And yes, any representative project that will definitely be imported will work.

LorenzoBettini:
> Yes, the working sets are applied to actual IProject/IResource objects after these things are imported. And yes, any representative project that will definitely be imported will work.

Thanks! I'll update the setup locally and check that it works as expected; then, I'll update the PR.

LorenzoBettini:
However, we don't have several "Working Sets" task: we have one as in your screenshot @merks

LorenzoBettini:
OK, with the latest commit it seems to work:

image

I've tried importing another project from another git repository and it gets into "Other Projects" automatically. I tried creating another project and it gets into "Other Projects" automatically.

@cdietrich beware: if you create a new project and a working set is selected, the new project will go into that working set. I think that's expected, right @merks ?

merks:
Certainly the IDE allows you to manually add things to any working set.  But when the dynamic working sets are updated, the rules are reapplied.  One can apply them via the preferences:

image

HannesWell (approved):
Looks good and worked well for me.

This will likely conflict with https://github.com/eclipse/xtext/pull/2213.
@LorenzoBettini do you want to merge this first?

LorenzoBettini:
@HannesWell I've merged it.

PR #2219, opened 2023-04-15
Remove 'de.itemis.xtext.antlr.feature' from target files

HannesWell:
As discussed in https://github.com/eclipse/xtext/pull/2213#issuecomment-1506846616, the de.itemis.xtext.antlr.feature should not be in the target-files.
It is only used for developer convenience in the xText development workspace and already added to that by the Xtext Oomph-setup.

### DIFF
--- a/org.eclipse.xtext.generator/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.generator/META-INF/MANIFEST.MF
@@ -59,7 +59,6 @@ Require-Bundle: org.eclipse.xtext.xtext.generator;visibility:=reexport,
  org.eclipse.xtend.typesystem.emf;bundle-version="2.2.0";visibility:=reexport,
  org.eclipse.emf.codegen.ecore;bundle-version="2.29.0";resolution:=optional;visibility:=reexport;x-installation:=greedy,
  org.eclipse.xtext;resolution:=optional;x-installation:=greedy,
- de.itemis.xtext.antlr;bundle-version="2.0.0";resolution:=optional;visibility:=reexport,
  org.eclipse.emf.ecore;bundle-version="2.26.0",
  org.eclipse.emf.common;bundle-version="2.24.0",
  org.antlr.runtime;bundle-version="[3.2.0,3.2.1)",
@@ -72,5 +71,6 @@ Require-Bundle: org.eclipse.xtext.xtext.generator;visibility:=reexport,
 Bundle-ActivationPolicy: lazy
 Import-Package: org.apache.commons.logging;version="1.0.4";resolution:=optional;x-installation:=greedy,
  org.apache.log4j;version="1.2.24"
+DynamicImport-Package: de.itemis.xtext.antlr.toolrunner
 Automatic-Module-Name: org.eclipse.xtext.generator
 Eclipse-SourceReferences: eclipseSourceReferences
--- a/org.eclipse.xtext.xtext.generator/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.xtext.generator/META-INF/MANIFEST.MF
@@ -17,7 +17,6 @@ Require-Bundle: org.eclipse.xtext;x-installation:=greedy,
  org.eclipse.emf.mwe2.lib;bundle-version="2.14.0";resolution:=optional,
  org.eclipse.equinox.common;bundle-version="3.16.0",
  org.antlr.runtime;bundle-version="[3.2.0,3.2.1)",
- de.itemis.xtext.antlr;bundle-version="2.0.0";resolution:=optional;visibility:=reexport,
  org.eclipse.jdt.core;bundle-version="3.29.0";resolution:=optional
 Import-Package: org.apache.log4j;version="1.2.24"
 Export-Package: org.eclipse.xtext.xtext.generator,
@@ -62,5 +61,6 @@ Export-Package: org.eclipse.xtext.xtext.generator,
  org.eclipse.xtext.xtext.generator.validation,
  org.eclipse.xtext.xtext.generator.web,
  org.eclipse.xtext.xtext.generator.xbase
+DynamicImport-Package: de.itemis.xtext.antlr.toolrunner
 Automatic-Module-Name: org.eclipse.xtext.xtext.generator
 Eclipse-SourceReferences: eclipseSourceReferences
--- a/xtext-latest.target
+++ b/xtext-latest.target
@@ -38,10 +38,6 @@
 			<unit id="org.eclipse.lsp4j.sdk.feature.group" version="0.0.0"/>
 		</location>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<repository location="https://download.itemis.com/updates/releases/2.1.1"/>
-			<unit id="de.itemis.xtext.antlr.feature.feature.group" version="0.0.0"/>
-		</location>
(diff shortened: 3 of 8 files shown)


cdietrich:
it looks like its needed anyway cause tycho insists on resolving optional dependcies

HannesWell:
> it looks like its needed anyway cause tycho insists on resolving optional dependcies

That's unexpected. Maybe with the new resolver in Tycho 3/4 this is handled more gracefully.
I suggest to keep this open until Tycho 3 is used for xtext and then try it again?

LorenzoBettini:
@HannesWell, if you have time, could you please check locally whether with Tycho 3.0.4 there's no error in TP resolution? I guess `validate` phase should be enough.

HannesWell:
> @HannesWell, if you have time, could you please check locally whether with Tycho 3.0.4 there's no error in TP resolution? I guess `validate` phase should be enough.

Tried it with Tycho 3.0.4 locally and the same error occurred.

@laeubi do you have any idea, why we get the resolution error below with that require-bundle entry (with and without the reexport)?
`Require-Bundle: de.itemis.xtext.antlr;bundle-version="2.0.0";resolution:=optional;visibility:=reexport`
I would expect Tycho to ignore optional dependencies if absent.

```
 [INFO] Resolving target definition file:/home/jenkins/agent/workspace/xtext_PR-2219/org.eclipse.xtext.xtext.generator/../xtext-r202203.target for environments=[macosx/cocoa/x86_64, win32/win32/x86_64, linux/gtk/x86_64], include source mode=honor, execution environment=StandardEEResolutionHints [executionEnvironment=OSGi profile 'JavaSE-11' { source level: 11, target level: 11}], remote p2 repository options=org.eclipse.tycho.p2.remote.RemoteAgent@1055d261...
 [INFO] Resolving dependencies of MavenProject: org.eclipse.xtext:org.eclipse.xtext.xtext.generator:2.31.0-SNAPSHOT @ /home/jenkins/agent/workspace/xtext_PR-2219/org.eclipse.xtext.xtext.generator/pom.xml
 [INFO] {osgi.os=macosx, osgi.ws=cocoa, org.eclipse.update.install.features=true, osgi.arch=x86_64}
 [ERROR] Cannot resolve project dependencies:
 [ERROR]   Software being installed: org.eclipse.xtext.xtext.generator 2.31.0.qualifier
 [ERROR]   Missing requirement: org.eclipse.xtext.xtext.generator 2.31.0.qualifier requires 'osgi.bundle; de.itemis.xtext.antlr 2.0.0' but it could not be found
 [ERROR] 

```

laeubi:
> I would expect Tycho to ignore optional dependencies if absent.

If you want to ignore them simply remove them ;-)

Otherwise there are no "optional dependencies" in terms of a build-tool, because in contrast to the runtime where you probably do not execute all code pathes (or can handle the absence of a service / class / resource / ...) at compile time you need to compile everything.

Still I have created already:
- https://github.com/eclipse-tycho/tycho/issues/2309

cdietrich:
=> we have to keep them

HannesWell:
> Otherwise there are no "optional dependencies" in terms of a build-tool, because in contrast to the runtime where you probably do not execute all code pathes (or can handle the absence of a service / class / resource / ...) at compile time you need to compile everything.

Indeed, makes sense in general.
But in this case the classes are only accessed reflectively and therefore don't have to be present at compile-time either:
https://github.com/eclipse/xtext/blob/93acb221e6fabbc2d09e687365bfbbe4bdf29ce5/org.eclipse.xtext.generator/src/org/eclipse/xtext/generator/parser/antlr/AntlrToolFacade.java#L29

Wouldn't it make sense to use `DynamicImport-Package` in this case?
@cdietrich how can I test if that still works at runtime.

laeubi:
> But in this case the classes are only accessed reflectively and therefore don't have to be present at compile-time either:

The one should use `Dynamic-ImportPacakge` instead ...

cdietrich:
i dont know the 
Dynamic-ImportPacakge
do you
@szarnekow

szarnekow:
I'm aware of the concept of Dynamic-ImportPackage but have never used it. Conceptionally it should do the trick.

laeubi:
- https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.dynamicimportpackage
- http://keheliya.blogspot.com/2013/02/use-of-dynamicimport-package-in-osgi.html

cdietrich:
=> would need to be tested if it works.

HannesWell:
> => would need to be tested if it works.

Yes. Can you tell me an elegant way to do that? Then I can test it this evening.

In general, if this should be worked out in a clean way there could be a general `WhatEverTool` interface in the generator Plugin which is implemented by the itemes-implemention for example as declarative service. Then at the place where now the class is looked up, the OSGi framework could be asked for that service.
If that should work outside of OSGi, the Service-Loader mechanism of Java could be used.
But both would require a bit more rework on both sites (which I'm currently not eager to do^^).

cdietrich:
the dependeny is in xtext.xtext.generator.
so i expect if the itemis plugin is in tp
and you create a new xtext project wizard 
and then run/debug the workflow file,
it should not download antlr but load the antlr.generator from classpath instead

HannesWell:
I tested it with the example dsl xtext project from the xtext dev workspace and there I see the following print-out:
```
581  INFO  GenModelHelper     - Registered GenModel 'http://www.eclipse.org/xtext/common/JavaVMTypes' from 'platform:/resource/org.eclipse.xtext.common.types/model/JavaVMTypes.genmodel'
707  INFO  AntlrToolFacade    - Downloading file from 'https://download.itemis.com/antlr-generator-3.2.0-patch.jar' ...
1291 INFO  AntlrToolFacade    - Finished downloading.
1318 INFO  XtextGenerator     - Generating org.xtext.example.mydsl.MyDsl
1380 INFO  EMFGeneratorFragment2 - Generating EMF model code
```
So is this kind of download expected or not?

Nevertheless I thought about it again and assume the mwe2 workflow execution isn't an OSGi runtime, is it? Therefore the DynamicImport-Package header is actually meaningless there. But the project's dependencies are probably added to the classpath of the mwe2 runtime, aren't they? But for that the regular Import-Package header can be considered and DynamicImport-Package probably isn't.
When I revert to the old Import-Package header I don't get the download printout, even if the downloaded jar is absent.
So I have the impression that the Import-Package is necessary here and therefore also the IUs in the TP.

As far as I know there is know other trick to pull in the itemis-plugin into a launch, unless one specifies a launch config for it and adds it.

cdietrich:
the xtext.generator and xtext.xtext.generator have a dep to the itemis thing which has a dependency to antlr.generator. so when the itemis thing is there the antlrtoolfacade should be able to find and use it.

cdietrich:
any update here? for me it works without download when the itemis plugin is present in tp.

HannesWell:
> any update here? for me it works without download when the itemis plugin is present in tp.

Unfortunately I cannot report anything successful. I also tried to use p2.inf to tell Tycho that the package requirement is not greedy but that seems not to be considered during initialization.

So unless there isn't or at least will be soon another trick to tell Tycho to ignore the Package during resolution, I think this attempt failed und the PR can be abandoned.

cdietrich:
ok thx for your effort anyway

HannesWell:
Should we then remove the items feature from the Oomph targlet? Since it is already in the target it is not nessesary to have it in the setup again.
If you agree I can create a PR.

cdietrich:
yes please create a pr. but i will have to test this once the pr is open if it still is working

HannesWell:
> yes please create a pr. but i will have to test this once the pr is open if it still is working

Of course, please see https://github.com/eclipse/xtext/pull/2641.

PR #2221, opened 2023-04-15
purexbase.eclipse.tests in the build

### LINKED ISSUE #2220: purexbase.eclipse.tests not executed during the build
I did not notice that during repo merge, because it wasn't executed in the original `xtext-eclipse` repository either.

Was that intentional @szarnekow ?

I can create a PR to enable that.
There are only two tests, and the second one is ignored with a TODO.

LorenzoBettini:
Closes #2220


cdietrich (approved):
approved given the ci build succeeds

Issue #2225, opened 2023-04-16
Setup doesn't work anymore


merks:
You have to be more careful with the modified modular targlet task because you are include a target platform resource specified by ID but that only works if that *.target is already in the workspace (and of course the clone must exist too):

![image](https://user-images.githubusercontent.com/208716/232292111-a4834c7d-e0d6-4484-971a-f73262abd05d.png)

Personally I would have gone for approach to generate the target from the *.target from targlet...

Easiest is to move all the targlet tasks after the import projects task.

Getting a clean build of out this whole deal seems challenging at best:
```
Updating duplicate resource: '/xtext-parent/org.eclipse.xtend.examples/projects/xtend-examples/xtend-gen/example6/.Movies.java._trace'.
ERROR: org.eclipse.core.resources code=75 Errors occurred during the build.
  at org.eclipse.core.internal.resources.Workspace.buildInternal(Workspace.java:541)
  at org.eclipse.core.internal.resources.Workspace.build(Workspace.java:420)
  at org.eclipse.oomph.setup.projects.impl.ProjectsBuildTaskImpl.perform(ProjectsBuildTaskImpl.java:525)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.doPerformNeededSetupTasks(SetupTaskPerformer.java:3864)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer$WorkspaceUtil$1.run(SetupTaskPerformer.java:5200)
  at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2382)
  at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2407)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer$WorkspaceUtil.performNeededSetupTasks(SetupTaskPerformer.java:5193)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.performNeededSetupTasks(SetupTaskPerformer.java:3798)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.performTriggeredSetupTasks(SetupTaskPerformer.java:3773)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.perform(SetupTaskPerformer.java:3651)
  at org.eclipse.oomph.setup.ui.wizards.ProgressPage$9.run(ProgressPage.java:592)
[...]

PR #2226, opened 2023-04-16
Ensure that targlets are performed after project imports

merks:
https://github.com/eclipse/xtext/issues/2225


cdietrich (approved):

cdietrich:
thx for the fix @merks  cc @HannesWell

PR #2569, opened 2023-04-17
macOS failure in GitHub Actions

### LINKED ISSUE #2229: macOS failure in GitHub Actions
the build on GitHub Actions in macOS started to fail like that consistently:

```
Downloaded from central: https://repo.maven.apache.org/maven2/org/eclipse/emf/org.eclipse.emf.mwe2.launch/2.14.0/org.eclipse.emf.mwe2.launch-2.14.0.pom
...
Error:  Failed to execute goal on project org.eclipse.xtext.common.types.tests: Could not resolve dependencies for project org.eclipse.xtext:org.eclipse.xtext.common.types.tests:jar:2.31.0-SNAPSHOT: Failed to collect dependencies at org.eclipse.emf:org.eclipse.emf.mwe2.language:jar:2.14.0: Failed to read artifact descriptor for org.eclipse.emf:org.eclipse.emf.mwe2.language:jar:2.14.0: The following artifacts could not be resolved: org.eclipse.emf:org.eclipse.emf.mwe2.language:pom:2.14.0 (absent): Could not transfer artifact org.eclipse.emf:org.eclipse.emf.mwe2.language:pom:2.14.0 from/to central (https://repo.maven.apache.org/maven2): Connection reset -> [Help 1]
```

These are my suspects:

- macOS uses Maven 3.9.1, which might have introduced some breakage
- The really strict version in mwe2.launch https://github.com/eclipse/mwe/blob/master/plugins/org.eclipse.emf.mwe2.launch/pom.xml#L21 "[version]" is even stricter than "version"

Concerning the second point, https://books.sonatype.com/mvnref-book/reference/pom-relationships-sect-project-dependencies.html#pom-relationships-sect-version-ranges seems to discourage that syntax:

> When declaring a "normal" version such as 3.8.2 for Junit, internally this is represented as "allow anything, but prefer 3.8.2." This means that when a conflict is detected, Maven is allowed to use the conflict algorithms to choose the best version. If you specify [3.8.2], it means that only 3.8.2 will be used and nothing else. If somewhere else there is a dependency that specifies [3.8.1], you would get a build failure telling you of the conflict. We point this out to make you aware of the option, but use it sparingly and only when really needed. The preferred way to resolve this is via dependencyManagement.

LorenzoBettini:
Closes #2229 

I think that listing transitive dependencies of mwe2 is a better approach, but I still think mwe2 POMs should be fixed under that respect

### DIFF
--- a/org.eclipse.xtext.common.types.tests/pom.xml
+++ b/org.eclipse.xtext.common.types.tests/pom.xml
@@ -41,11 +41,15 @@
 		</dependency>
 		<dependency>
 			<groupId>org.eclipse.emf</groupId>
-			<artifactId>org.eclipse.emf.mwe2.language</artifactId>
+			<artifactId>org.eclipse.emf.mwe2.runtime</artifactId>
 		</dependency>
 		<dependency>
 			<groupId>org.eclipse.emf</groupId>
 			<artifactId>org.eclipse.emf.mwe2.launch</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.emf</groupId>
+			<artifactId>org.eclipse.emf.mwe2.language</artifactId>
 		</dependency>
 		<dependency>
 			<groupId>org.eclipse.platform</groupId>


cdietrich (approved):
given tests are green

PR #2570, opened 2023-04-17
Tycho 3 and parameterized Jenkins job

LorenzoBettini:
@cdietrich @HannesWell I had a few spare time and I wanted to try my idea to deal with Tycho 3 on Java 17 and Tycho 2.7.5 on Java 11.

The parameterization of the Jenkinsfile is pretty simple, IMHO. Whether it works it's still under testing.

I've only tested it on my fork and GitHub actions with a dedicated job for Java 11, overriding `tycho-version` and it works like a charm.

This is a temporary workaround to switch to Tycho 3 until the toolchain stuff is ready.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -62,7 +62,7 @@ pipeline {
       }
       steps {
         xvnc(useXauthority: true) {
-          sh "./full-build.sh --tp=${selectedTargetPlatform()}"
+          sh "./full-build.sh --tp=${selectedTargetPlatform()} ${javaVersionBasedProperties()}"
         }
       }// END steps
     } // END stage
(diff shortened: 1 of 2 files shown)


cdietrich (approved):

LorenzoBettini:
@HannesWell as I said, it's a temporary solution. We need Tycho 3 for building with more recent target platforms, to benefit from its fixed bugs and its alignment with the new target platforms (see all the problems we had when switching to jupiter 5.9.2). For the moment, we don't need features that are only in Tycho 3.

Note that we can still build with Java 11 as long as we revert to Tycho 2.7.5, which this PR does.

The function in the Jenkinsfile can also be used in your work about toolchains to create the arguments for switching to the toolchain. If not, we'll remove that.

I'm merging this, but I'm NOT closing the corresponding issue: we'll close that when the toolchain is up and running.

HannesWell:
> @HannesWell as I said, it's a temporary solution. We need Tycho 3 for building with more recent target platforms, to benefit from its fixed bugs and its alignment with the new target platforms (see all the problems we had when switching to jupiter 5.9.2). For the moment, we don't need features that are only in Tycho 3.

OK, then it makes sense to have this right now. Thanks for the elaboration.

PR #2618, opened 2023-04-20
[Build] Extend GH workflow matrix for jdkVersion and targetPlatform

HannesWell:
This PR extends the GH workflow matrix for Xtext with an axis for the `jdkVersion` and `targetPlatform` version. So that all selected combinations are always build.
In order to limit the number of combinations only the oldest and latest supported Eclipse TP are build.

The incompatible combination of Java-11 and Eclipse 2023-06 (aka latest), that [matrix cell is excluded](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations).

Because the `coactions/setup-xvfb` can only execute a single command I had to adjust the `strict-release-jdk` a bit so that the the jdk can be configured via an extra property. I didn't found a solution to use the same approach as in the Jenkins file.

I tried 
```
      - name: Build and test
        uses: coactions/setup-xvfb@v1
        with: 
          run: "(if [[ ${{ matrix.jdkVersion }} == 11 ]]; then toolchainArgs='--toolchains ../releng/toolchains.xml -Pstrict-release-jdk';fi && \
            mvn clean verify -PuseJenkinsSnapshots $toolchainArgs -Dtarget-platform-classifier=xtext-${{ matrix.targetPlatform }})"
```
But then setup-xvfb only tried to run the command if and failed.
Personally I also find the approach used now simpler to read.

`--toolchains releng/toolchains.xml -Pstrict-release-jdk -Dtoolchain.jdk=17` is actually a NoOp and not specifying has the same result since JAVA_HOME points to the same JDK as jdk-17 from the toolchain.
Since I found it simpler to read I applied the same approach in the Jenkins workflow, but if you prefer the previous way I'll revert that part. 

In order to make the job labels not too long I also adjusted the `os` axis values.

You can see the results in my fork at:
https://github.com/HannesWell/eclipse.xtext/actions/runs/4757063052

From https://github.com/eclipse/xtext/pull/2223#issuecomment-1516145445
> It would be nice to have such a matrix in this repository, but due to the limitations of the usage of GitHub Actions shared by all Eclipse repositories, it wouldn't be feasible. Maybe just for Ubuntu and not macOS... maybe just nightly. We can investigate that in the future.

AFAIK the workflow resource quota is per organization, so if Xtext would have its own GH-organisation more resources would probably be available.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -11,11 +11,17 @@ on:
 jobs:
   build:
     if: github.event_name != 'pull_request' || github.event.pull_request.base.repo.clone_url != github.event.pull_request.head.repo.clone_url
+    name: ${{ matrix.os }}, Java-${{ matrix.jdkVersion }}, Eclipse ${{ matrix.targetPlatform }}
     strategy:
       matrix:
-        os: ['ubuntu-latest', 'macos-latest' ] # 'windows-latest' too flaky
+        os: ['ubuntu', 'macos' ] # 'windows-latest' too flaky
+        jdkVersion: ['11', '17']
+        targetPlatform: ['r202203', 'latest'] # Only test latest and earliest platform
+        exclude: # Since Eclipse 2023-06 Java17 is required
+          - jdkVersion: 11
+            targetPlatform: latest
       fail-fast: false
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}-latest
 
     steps:
       - name: 'Check out repository'
@@ -24,7 +30,9 @@ jobs:
       - name: 'Set up Java'
         uses: actions/setup-java@v3
         with:
-          java-version: 17
+          java-version: |
+            11
+            17
           distribution: 'temurin'
 
       - name: 'Cache Maven packages'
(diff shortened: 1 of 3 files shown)


LorenzoBettini:
@HannesWell , as I suggested in the other PR, I don't think we want to change the maven.yml; what might help would be **another** workflow that only runs periodically (even nightly might be too much, maybe weekly and/or on demand, AKA workflow dispatch, or both).

The combinations currently in the matrix of this PR maybe are not that important because we already have those configurations in Jenkins, so I wouldn't repeat the effort. Something more along the lines of the experiments you made for the other PR (https://github.com/HannesWell/eclipse.xtext/actions/runs/4746875852), where you were testing also with the other target platforms, that, IIRC, are never tested automatically in Jenkins. I would stick with Linux only, which is way faster. I don't think it's worthwhile to test with macOS with so many configurations. (this would also simplify the maven command since it could be simply be always prefixed with xvfb-run)

The `Jenkinsfile` should not be changed either: really, I wouldn't want to always run with toolchains, even only because locally I wouldn't want to run with a toolchain unless strictly required. So, I prefer to have a run with no toolchain at all.

I also guess it was decided not to switch to another organization when we switched to the monorepo. In that respect, if I need to do GitHub Actions CI heavily, I do it on my fork (this was also suggested on the cross dev mailing list). 

However, this is just my opinion, we should hear from @cdietrich and @szarnekow

HannesWell:
I don't expect this to be completed and therefore close this.
If still of interest, it can be the base for a follow-up change.

PR #2619, opened 2023-04-20
Simplify JDK_VERSION parameter and make 17 the default value

HannesWell:
Initially suggested in https://github.com/eclipse/xtext/pull/2223, but reverted to make that PR simpler to review, I now want to suggest again in a smaller scope to simplify the `JDK_VERSION` parameter of the Xtext's Jenkins pipeline.

Furthermore I assume that, although Xtext is compiled against Java-11 testing with Java-17 is often more interesting and therefore suggest to use 17 as default (and therefore as first). 
Or do you want to leave the Java-17 testing to the GH-workflows?

In the same sense it could also be considered to make the latest TP the default/first.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,15 +8,13 @@ pipeline {
   parameters {
     choice(name: 'TARGET_PLATFORM', choices: ['r202203', 'r202206', 'r202209', 'r202212', 'r202303', 'latest'], description: 'Which Target Platform should be used?')
     // see https://wiki.eclipse.org/Jenkins#JDK
-    choice(name: 'JDK_VERSION', description: 'Which JDK should be used?', choices: [
-       'temurin-jdk11-latest', 'temurin-jdk17-latest'
-    ])
+    choice(name: 'JDK_VERSION', choices: [ '11', '17' ], description: 'Which JDK version should be used?')
   }
 
   triggers {
     parameterizedCron(env.BRANCH_NAME == 'main' ? '''
-      H H(0-1) * * * %TARGET_PLATFORM=r202203;JDK_VERSION=temurin-jdk17-latest
-      H H(3-4) * * * %TARGET_PLATFORM=latest;JDK_VERSION=temurin-jdk17-latest
+      H H(0-1) * * * %TARGET_PLATFORM=r202203;JDK_VERSION=17
+      H H(3-4) * * * %TARGET_PLATFORM=latest;JDK_VERSION=17
       ''' : '')
   }
 
@@ -121,7 +119,7 @@ pipeline {
 
 /** return the Java version as Integer (8, 11, ...) */
 def javaVersion() {
-  return Integer.parseInt(params.JDK_VERSION.replaceAll(".*-jdk(\\d+).*", "\$1"))
+  return Integer.parseInt(params.JDK_VERSION)
 }
 
 /** returns true when this build was triggered by an upstream build */


LorenzoBettini:
@HannesWell IIRC, @cdietrich and @szarnekow really want to also test with JDK 11 at runtime, not only compile. And that should be the default.

We have two nightly triggers that also test with JDK 17, and if we want to do it on the fly for a branch, then we simply "Run with parameters". That's what they told me when I did the porting. Moreover, GitHub Actions already builds with JDK 17 only.

To summarize, I think that the `Jenkinsfile` should not be touched in that respect ;)

cdietrich:
yes we want to make sure we work fine with J11 and r202203.

HannesWell:
Acknowledge. Restored the previous order of 11, 17.

Jenkinsfile
```diff
@@ -8,15 +8,13 @@ pipeline {
   parameters {
     choice(name: 'TARGET_PLATFORM', choices: ['r202203', 'r202206', 'r202209', 'r202212', 'r202303', 'latest'], description: 'Which Target Platform should be used?')
     // see https://wiki.eclipse.org/Jenkins#JDK
-    choice(name: 'JDK_VERSION', description: 'Which JDK should be used?', choices: [
-       'temurin-jdk11-latest', 'temurin-jdk17-latest'
-    ])
+    choice(name: 'JDK_VERSION', choices: [ '11', '17' ], description: 'Which JDK version should be used?')
```
cdietrich, line 11: just glancing at the pr its unclear how this maps to the toolchains

cdietrich: so we use toolchains only for 11?

cdietrich:
@LorenzoBettini do the nightly/milestone deploy scripts/jenkinsfiles need adaptations too?

Jenkinsfile
cdietrich: tested locally. seems to work when both env vars are set

LorenzoBettini:
> @LorenzoBettini do the nightly/milestone deploy scripts/jenkinsfiles need adaptations too?

No, they don't: in those cases we use Java 17.

cdietrich:
looks good from my side @LorenzoBettini what do you think?

Jenkinsfile
LorenzoBettini: What do you mean by tested locally? This PR does not change the environment variables (those were already changed in the https://github.com/eclipse/xtext/pull/2223 so they were already working). This PR changes the job's parameters names for JDK. Unfortunately, the Jenkins PR does not check that because for security reasons, it still takes the Jenkinsfile from the internal repository not from the external one. We should replay the PR by manually copying the Jenkinsfile. In this case, I think we can avoid that and merge this. Then, I think the first time we'll have to make sure that the jobs for the `main` branch are aware of the changed parameters. We saw in the past, that Jenkins does not react well when you change parameters' names or values' orders ;)

cdietrich: this is why i did
https://ci.eclipse.org/xtext/job/xtext/job/HannesWell-jenkinsParams/

LorenzoBettini:
> looks good from my side @LorenzoBettini what do you think?

Yes, I'm fine (please also see my answer to your latest comment above about testing locally).

Jenkinsfile
LorenzoBettini: OK! So you tested the parameters, that's great!
You only mentioned the environment variables, which are independent of the parameters (they are always both set anyway and always), that's why I wanted to be sure.

cdietrich: @HannesWell can you confirm this?

cdietrich:
```
06:52:09  [INFO] --- tycho-surefire-plugin:3.0.4:test (default-test) @ org.eclipse.xtext.purexbase.tests ---
06:52:09  [INFO] Could not find a java toolchain of type jdk, using java from JAVA_HOME instead (/opt/tools/java/temurin/jdk-17/latest/bin/java)
06:52:09  [INFO] Executing Test Runtime with timeout 0, logs (if any) will be placed at: /home/jenkins/agent/workspace/xtext_HannesWell-jenkinsParams/org.eclipse.xtext.purexbase.tests/target/work/data/.metadata/.log
```

LorenzoBettini: @cdietrich I once again confirm this.

As I said, that was already tested in the https://github.com/eclipse/xtext/pull/2223 .
This PR only changes the parameter names, it does not affect the toolchain strategy that has already been implemented in https://github.com/eclipse/xtext/pull/2223.

Well, initially it did, but then @HannesWell reverted that part.

cdietrich (approved):

cdietrich:
thx @HannesWell

LorenzoBettini:
@cdietrich as expected, the first build job on `main` for this merge failed because of changed parameters. I manually restarted the build with the new ones. From now on, we should be fine.

Issue #2627, opened 2023-04-25
New Build: Xtend Classpath container in Runtime eclipse


cdietrich:
If i start a a runtime eclipse
and create a java project
and then add the xtend classpath container
(e.g. create xtend file and use quickfix or use the java buildpath config)
then the xtend classpath container is not working

(there seems to be inconsistency with class dirs, build properties, whatever)

@LorenzoBettini  @szarnekow do you see this problem too
any idea what could be the problem

cdietrich:
i assume the bin/bin guessing in 
org.eclipse.xtend.ide.buildpath.XtendClasspathContainer
will need adaptation

LorenzoBettini:
I hadn't thought about such a use case and I thought we had a test even for such situations ;)

cdietrich:
have no clue how this could be tested automatically

LorenzoBettini:
I don't think we can because I think we use workarounds instead of using general solutions based on the classloader. But I wouldn't know how to do that.

PR #2628, opened 2023-04-25
[#2627] fix XtendClasspathContainer to work with the new maven based project setup

cdietrich:
[#2627] fix XtendClasspathContainer to work with the new maven based project setup

### DIFF
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/buildpath/XtendClasspathContainer.java
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/buildpath/XtendClasspathContainer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2012, 2023 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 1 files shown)


org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/buildpath/XtendClasspathContainer.java
```diff
@@ -87,7 +87,16 @@ private IPath bundlePath(Bundle bundle) {
...
+		URL binFolderURL = FileLocator.find(bundle, new Path("target/classes"), null); //$NON-NLS-1$
+		if (binFolderURL != null) {
+			try {
+				URL binFolderFileURL = FileLocator.toFileURL(binFolderURL);
+				return new Path(binFolderFileURL.getPath()).makeAbsolute();
+			} catch (IOException e) {
+				LOG.error("Can't resolve path '" + bundle.getSymbolicName() + "'", e); //$NON-NLS-1$ //$NON-NLS-2$
+			}
+		}
+		binFolderURL = FileLocator.find(bundle, new Path("bin/main"), null); //$NON-NLS-1$
```
cdietrich, line 99: should we remove the two bin options?

LorenzoBettini: bin/main is history so I'd keep only the new one

LorenzoBettini (approved):
I seem to understand that this is only for runtime Eclipse, so if you tested it and it works then we can merge this. I'd only remove the old bin main folder as I commented inline.

LorenzoBettini:
Argh... I hadn't noted there's also an attempt with "bin"! That one shouldn't be there either, because .classpath should already be set to generate to target/classes, not to bin.

cdietrich:
ok, i have also removed that one ;)

PR #2635, opened 2023-04-26
 bootstrap against 2.31.0.M1 & bump node and npm 

cdietrich:


LorenzoBettini (approved):
Given the build is green.
Does the bumping of node and npm help concerning downloading?

cdietrich:
It seems to solve the 500 error that were there today

cdietrich:
i dont know. i tested both variants in browser. one was working the other not

PR #2637, opened 2023-04-26
simplify the junit5 dep in target

cdietrich:
simplify the junit5 dep in target


LorenzoBettini (approved):
Personally, I never specify versions in the tp (version 0.0.0), so I'm perfectly fine with this PR. We lose some control and reproducibility but that allows for less maintenance burden.

PR #2641, opened 2023-04-28
Remove 'de.itemis.xtext.antlr.feature' from Setup targlet

HannesWell:
After the attempt to remove the itemis feature from Xtext's target files in https://github.com/eclipse/xtext/pull/2219 failed, this PR removes the itemis feature from Xtext's Oomph setup to avoid having it defined one more time than necessary.


cdietrich (approved):

cdietrich:
thx @HannesWell

Issue #2644, opened 2023-05-01
Dont build on Github Actions for tags


cdietrich:
Dont build on Github Actions for tags

@LorenzoBettini  @HannesWell  any idea how to prevent the actions to run on tags?

HannesWell:
According to the docs, the push event is fired when a branch or tag is pushed:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

We can probably use the `tags-ignore` filter to not run on tags at all.
I'll try it out and provide a PR if it works.

PR #2646, opened 2023-05-01
Don't run 'Build' GH workflow when tags are pushed

HannesWell:
As desribed in the [GH workflow docs for the push event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push), if one specifies the `branches` filter all non-matching refs are ignored by the workflow.
Consequently when running on any branch, all tags are excluded.

Tested it in my fork and it worked as expected.

Fixes https://github.com/eclipse/xtext/issues/2644

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -2,6 +2,7 @@ name: Build
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


cdietrich:
thx @HannesWell

Issue #2176, opened 2023-03-24
Require-Bundle javax.inject


jukzi:
Instead of using "Require-Bundle" could you please use "Import-Package" for javax.inject

We currently try to use a Non-OSGi java where libraries should not conflict but
bundle javax.inject conflicts with jakarta.inject which is normally used for that package. As far as i know org.eclipse.xtext.util is the only lib that explicitly requests javax.inject.

https://github.com/eclipse/xtext-core/blob/0ca980145ddb04780e6e898a023458782e734977/org.eclipse.xtext.util/META-INF/MANIFEST.MF#L48

HannesWell:
> @jukzi @szarnekow can package import be reexported? so this would break a ton of xtext plugin and all xtext languages out there. e.g. mwe

Not as far as I know.

> A bundle can export it's imported packages, even though the UI does not allow it. The OSGI spec has rules how that situation will be resolved, though. I'm not terribly sure about that approach though, since it's certainly prone to confusion.

It is more the other way round, a bundle can import the packages it exports. In general a Bundle can only export what is contained in the bundle.


> so this would break a ton of xtext plugin and all xtext languages out there. e.g. mwe

As far as I can tell there will be no way around that, if that issue should be addressed. This is one of the main the reason why reexported require-bundle is strongly discouraged and require bundle is also discouraged, although it might seem to be more convenient at first sight.

HannesWell:
> best would be to not reexport but package import in all depended bundles. yes it's not compatible but straight forward.

I suggest when Xtext has migrated to `jakarata.inject/annotation` annotations to remove the reexported bundles and then only import `jakarta.inject/annotation`.

For platform I started the migration, first only to the javax-annotations provided by jakarta and eventually to jakarta-annotations: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1056
If time permits I will finish this for the 2023-06 release.

And when Xtext requires that release IMHO it would be a good point to make the switch.

HannesWell:
> we still need javax.inject for 2023-06 as i dont think we get this updated so soon on our with the capacity i see.

If you are willing to potentially break down stream projects it would be sufficient to change the Require-bundle javax.inject to Import-Package `javax.inject`. That can be done quickly.
What could be tried out if it helps to make the re-exported javax.inject required bundle optional. Maybe this would make the re-export available if the bundle is present but would not force it's presence.

[...]

PR #2654, opened 2023-05-05
[#2176] use guice namespace annotations to prepare switch from javax.inject to jakarta.inject in guice

cdietrich:
[#2176] use guice namespace annotations to prepare switch from javax.inject to jakarta.inject in guice


szarnekow (approved):
Interesting that we had so many usages of javax.inject.

Thank you for cleaning this up.

cdietrich:
yes i saw this multiple times the last days. it says connection reset.
maybe something wrong with maven central connection on github action side

LorenzoBettini:
I still think it's the Maven version: in GitHub Actions macOS it's 3.9.1
and I still believe the versions [number] in mwe2 poms are wrong ;)

PR #2667, opened 2023-05-12
Avoid repeated calls to getBinaryStorageURI and getURIConverter

rubenporras:


cdietrich:
https://ci.eclipse.org/xtext/job/xtext/job/PR-2667/1/console
or did 2nd push fix that?

rubenporras:
I think so, but I need to wait for the build. I do not have a workspace setup at the moment, so for this tiny PR, I am doing the changes directly in GitHub.

Issue #2677, opened 2023-05-23
Problem with malformed bundles during build or opening DSL files


rosczja:
When opening a DSL file or building a project that contains DSL files an IllegalArgumentException is thrown whenever a jar archive is on the classpath of the project, claiming to be an OSGi bundle, but it has an empty BundleSymbolicName.

This happens in the class org.eclipse.xtext.ui.resource.JavaProjectResourceSetInitializer on line 86 where the prefix of an URL gets replaced. The issue can be traced back to the class  org.eclipse.xtext.ui.resource.Storage2UriMapperJavaImpl line 347 where the uriPrefix is set to the symbolic name of the bundle if it is not null.

cdietrich:
am not sure if i can follow you.
you mean there is a symbolic name (!= null) but it is empty?

rosczja:
Yes exactly. In the Manifest you will find "BundleSymbolicName=" which seems to translate to an empty string.

cdietrich:
can you provide a pr?

rosczja:
Sure. Already working on it.

rosczja:
Cannot create an account on Eclipse Foundation to sign CLA.

rosczja:
Because they disabled web account registration due to spam. At least that is what the website says.

cdietrich:
can you give me a link?

cdietrich:
@rosczja there is already an account with the id
jrosczak
is this you?

rosczja:
I finally managed to get it to work, my email was not public, so Eclipse Foundation could not match my account. Sorry for the chaos.

PR #2678, opened 2023-05-23
[#2677] Handle empty bundle symbolic name

rosczja:


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/SourceAttachmentPackageFragmentRootWalker.java
```diff
@@ -296,7 +296,7 @@ protected T handle(IJarEntryResource jarEntry, TraversalState state) {
 	 */
 	private String getBundleSymbolicName(Manifest manifest) {
 		String name = manifest.getMainAttributes().getValue("Bundle-SymbolicName");
-		if (name != null) {
+		if (name != null && !name.trim().isEmpty()) {
```
LorenzoBettini, line 299: I think that instead of trimming and check if empty you could use https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#isBlank()

rosczja: I wanted to keep it compatible to Java 8, but if Java 11 is the baseline, I can of course change it.

PR #2679, opened 2023-05-23
Handle empty bundle symbolic name

rosczja:
See #2677


cdietrich:
wonder what about similar places like

org.eclipse.xtext.builder.standalone.incremental.ResourceURICollector.registerBundle(File)
org.eclipse.xtext.builder.standalone.StandaloneBuilder.registerBundle(File)

szarnekow:
@rosczja Could you please add the same guard to both classes that @cdietrich mentioned?

rosczja:
Added the same check into the two files @cdietrich mentioned

szarnekow (approved):
Thank you

cdietrich:
thx @rosczja

Issue #2651, opened 2023-05-02
Remove outdated scripts and Jenkinsfiles


cdietrich:
Once we have the first release with new process done we should
remove outdated scripts and Jenkinsfiles

LorenzoBettini:
@cdietrich can we close this one?

cdietrich:
i believe we still have a ton of dead code / scripts

PR #2684, opened 2023-05-24
[#2651] remove old jenkinsfiles

cdietrich:


LorenzoBettini (approved):
If you think we don't have anything to take inspiration from these old files I'm OK to remove them.

cdietrich:
yes and not. but the old repos still exist.

LorenzoBettini:
> yes and not. but the old repos still exist.

You're right!

Issue #2184, opened 2023-04-04
Create new Dev Docs


cdietrich:
Create new Dev Docs

- [ ] how to regenate languages
- [ ] how to regenerate docs from website
- [ ] how to build milestones and release
- [ ] how to do version bumps

cdietrich:
move the rest to 2.32

PR #2685, opened 2023-05-25
[#2184] document milestone and release process

cdietrich:
[#2184] document milestone and release process


Builds.md
```diff
@@ -0,0 +1,88 @@
+
```
cdietrich, line 1: guess we need some beta testing this with the upcoming releases and milestones

PR #2700, opened 2023-06-13
Add check if WorkspaceFolders are supported or not

jnt0r:


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
```diff
@@ -232,18 +231,23 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
...
 				workspaceManager.initialize(workspaceFolders, this::publishDiagnostics, CancelIndicator.NullImpl);
 			} else {
+				URI baseDir = getBaseDir(params);
 				workspaceManager.initialize(baseDir, this::publishDiagnostics, CancelIndicator.NullImpl);
 			}
 			return result;
 		}, (cancelIndicator, it) -> it).thenApply(it -> initializeResult = it);
 	}
 
+	private boolean isWorkspaceFoldersSupported(InitializeParams params) {
```
cdietrich, line 247: should we make this protected?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
```diff
@@ -173,14 +173,6 @@ public void initialize(List<WorkspaceFolder> workspaceFolders, Procedure2<? supe
 		refreshWorkspaceConfig(cancelIndicator);
 	}
 
-	/**
-	 * @return whether this workspace manager supports multiple workspace root folders.
-	 * @since 2.21
-	 */
-	public boolean isSupportsWorkspaceFolders() {
```
cdietrich, line 180: wonder if we should keep this and also in the check, @szarnekow what do you think? do we want to opt out from workspace folders on server side=

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
jnt0r: Could do that. Then we should use the class field initializeParams instead of requiring them as a method param.

jnt0r:
@szarnekow what do you think about this?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
szarnekow: Yes, please

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
szarnekow: I've seen language server implementations that explicitly wanted to disable these, yes.

jnt0r: How would you opt out from this? By simply overriding this method?

jnt0r: Or by some sort of configuration?

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/WorkspaceFoldersTest.java
```diff
@@ -13,8 +13,13 @@
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
+import org.eclipse.lsp4j.ClientCapabilities;
+import org.eclipse.lsp4j.DidChangeNotebookDocumentParams;
```
szarnekow, line 17: Is this import unused?

org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/WorkspaceFoldersTest.java
```diff
@@ -47,16 +52,35 @@ public class WorkspaceFoldersTest extends AbstractTestLangLanguageServerTest {
...
+			WorkspaceClientCapabilities workspaceClientCapabilities = new WorkspaceClientCapabilities();
+			workspaceClientCapabilities.setWorkspaceFolders(true);
+			clientCapabilities.setWorkspace(workspaceClientCapabilities);
+			it.setCapabilities(clientCapabilities);
 			it.setWorkspaceFolders(
 					Lists.newArrayList(new WorkspaceFolder(uriExtensions.toUriString(rootFolder1.toURI()), "root1"),
 							new WorkspaceFolder(uriExtensions.toUriString(rootFolder2.toURI()), "root2")));
 		});
+		Assert.assertEquals(true, capabilities.getCapabilities().getWorkspace().getWorkspaceFolders().getSupported());
+		;
```
szarnekow, line 83: Unnecessary empty statement.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java
```diff
@@ -232,18 +231,25 @@ public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
...
 				workspaceManager.initialize(workspaceFolders, this::publishDiagnostics, CancelIndicator.NullImpl);
 			} else {
+				URI baseDir = getBaseDir(params);
 				workspaceManager.initialize(baseDir, this::publishDiagnostics, CancelIndicator.NullImpl);
 			}
 			return result;
 		}, (cancelIndicator, it) -> it).thenApply(it -> initializeResult = it);
 	}
 
+	protected boolean isWorkspaceFoldersSupported() {
```
szarnekow, line 247: This is in fact about the client capabilities. Can this method be renamed to `clientSupportsWorkspaceFolders`?

org.eclipse.xtext.testing/src/org/eclipse/xtext/testing/AbstractLanguageServerTest.xtend
```diff
@@ -217,13 +219,26 @@ abstract class AbstractLanguageServerTest implements Endpoint {
...
+	}
+	
+	protected def InitializeResult initialize((InitializeParams)=>void initializer, boolean callInitialized, boolean useRootPath) {
 		val params = new InitializeParams => [
 			processId = 1
 			workspaceFolders = #[
 				new WorkspaceFolder(root.toURI.normalize.toUriString, '')
 			]
 		]
 		initializer?.apply(params)
```
szarnekow, line 232: The initializer should be called after the evaluation of `useRootPath` - otherwise the new default behavior might override previously configured params.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/WorkspaceManager.java
szarnekow: I think the negotiation between client and server is done by means of the InitializeResult. If the server doesn't support folders, it would set the boolean value to false (not to true, respectively).

jnt0r:
I applied your changes @szarnekow 
Should I commit the generated files as well?

szarnekow:
> Should I commit the generated files as well?

You're referring to the Java files that originate from modified Xtend files?

jnt0r:
Yes, I mean that files.

cdietrich:
can you please also have a look at the failing tests

jnt0r:
I don't know why it closed this PR. I tried to fix the Builds and had some conflicts with the encoding, so I force pushed to my fork branch to undo these changes.

szarnekow:
I'd love to reopen it but the tip of your main branch appears to be part of this' repositories main branch already.

jnt0r:
I opened a new pr with the same changes.

szarnekow:
Thank you.

PR #2711, opened 2023-06-21
Remove reference to deprecated XbaseScopeProvider

steve-hickman-epistimis:
Sample code was out of date. This should correct it.


cdietrich:
The code for this is in Xtext-website repo

PR #2714, opened 2023-06-23
package also args4j to update site

cdietrich:
package also args4j to update site


LorenzoBettini:
I don't remember: who's using this? I seem to remember this is used only by a Maven only project?

cdietrich:
it is used by wizard. i had problem running stuff without it from eclipse.
thus i want to add it

LorenzoBettini:
OK!

PR #2723, opened 2023-06-30
[#2721] use maven gson instead of orbit gson

cdietrich:
[#2721] use maven gson instead of orbit gson


LorenzoBettini (approved):
If the build succeeds, I'm fine with this PR

szarnekow (approved):
Thank you for taking care of the orbit breakage.

Issue #2726, opened 2023-06-30
Update to Guava 32.1


cdietrich:
Update to Guava 32.1

cdietrich:
@merks where is org.apiguardian gone?

merks:
This?

https://download.eclipse.org/oomph/simrel-orbit/nightly/N202306301409/archive/download.eclipse.org/oomph/simrel-orbit/nightly/N202306301409/index/org.apiguardian.api_1.1.2.html

cdietrich:
no

<unit id="org.apiguardian" version="0.0.0"/>

merks:
Where was it before? Not that Orbit itself has been actively deleting things.

cdietrich:
:(

PR #2728, opened 2023-07-01
Update to latest 32.x guava

jonahgraham:
See https://github.com/eclipse-oomph/oomph.incubator/commit/d706d426fda8a02c0577f7b80450ac2625a827f2

I think this is part of #2726


cdietrich:
There is more places to adapt see https://github.com/eclipse/xtext/pull/2727

Issue #2715, opened 2023-06-25
LSP SemanticTokens offset incorrect in some cases


jnt0r:
The offset in SemanticTokens response is slightly off in some cases. I debugged this and found out, that the wrong offset is calculated deeper in xtext. The TextRegion is already filled with the wrong offset, which is then used to calculate the line and character in the SemanticTokensService. I don't know and yet didn't find where the TextRegion is created and filled with the offset.

It looks like line ending characters (\n) are counted in the offset as well:

![grafik](https://github.com/eclipse/xtext/assets/38696668/abda86f9-2f62-4d9f-a675-9df87e3fc18a)

![grafik](https://github.com/eclipse/xtext/assets/38696668/7c4234e2-81bf-4754-96ee-508bcdbb19b9)

![grafik](https://github.com/eclipse/xtext/assets/38696668/27eadc16-e812-4e6c-855f-0d070dba3be7)

cdietrich:
Xtext offsets are for the complete document and this include whitespace 

so the question is if they are correctly translated to line and column for lsp 

a sample grammar and unit test might be helpful

jnt0r:
I try to take a look into this in the next days.

jnt0r:
I think adjusting the way of calculating offsets is very complicated, as it is used widely in xtext and lays deeply in the core. Therefore, I am currently taking a look at the PositionReader that is returning an LSP4j Position based on an InputStream and an offset. I try to adjust it, so it ignores newlines.

jnt0r:
Update: The Problem lays in the underlying LineNumberReader which is used in the PositionReader. The LineNumberReader skips \r characters and does not count them as a character when reading.

![grafik](https://github.com/eclipse/xtext/assets/38696668/8a7b9e5f-1bc6-43d0-b258-3aab60434cde)

When the offset is calculated, it includes \r characters. Therefore, we get a wrong position, as these characters are skipped when resolving the position to this offset. 

As the LineNumberReader is part of the java.io package, I try to work around this behavior in the PositionReader by detecting if carriage return characters are used and then subtracting the line number from the calculated column. Still not 100% happy with my solution, so will think about a bit more.

szarnekow:
Can you provide an isolated test case that illustrates what you are observing? 
Something like 
```
var reader = new PositionReader("some string literal");
reader.read(..)
...
assertEquals(.. , reader.getPosiiton())
```
?

jnt0r:
[...]

PR #2734, opened 2023-07-06
Fix PositionReader wrong position on files with carriage returns

jnt0r:
This fixes the issue #2715 in the PositionReader on wrong position returned on files containing carriage returns.

As I said, I'm still not 100% sure this is the best solution. I'm open for input and new ideas.

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/util/PositionReaderTest.java
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/util/PositionReaderTest.java
@@ -1,0 +1,64 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2023 TypeFox GmbH (http://www.typefox.io) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ide.tests.util;
+
+import static java.lang.Math.*;
+
+import java.io.IOException;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


joaodinissf:
Hi jnt0r! I also ran into issue #2715 this week, and happened to look into this problem independently. I just saw that you had reported the same issue and prepared a PR for it.

The fix I prepared changes the behavior of the `read()` function so that it doesn't skip '\r' characters. I think this gets at the root cause of the issue. I have created a PR with it, #2735.

I believe that this is a more robust solution than the one you implemented in this change, because of what you mention in [this comment](https://github.com/eclipse/xtext/issues/2715#issuecomment-1614956679). What do you think?

szarnekow:
I prefer getting rid of the LineNumberReader altogether, too.
@joaodinissf can you please add tests to your PR with Linux/ Windows and mixed line endings?

joaodinissf:
Will do, @szarnekow.

szarnekow:
Bonus points if you can make it work with the anachronistic `\r` only (not `\r\n`), too.

joaodinissf:
@szarnekow, since @jnt0r already had some work preparing this contribution, how do you feel about the following plan:

1. We merge #2735 with the fix.
2. We merge #2734 with the tests that @jnt0r prepared.
3. In a separate PR, I can add a few more tests and make any necessary adjustments to correctly handle the bonus issue that you correctly [identified](https://github.com/eclipse/xtext/pull/2734#issuecomment-1625129094).

That way, both @jnt0r's and my contributions and efforts are preserved. How does that sound?

szarnekow:
Sounds very good to me.

jnt0r:
@joaodinissf I will take a look at your PR. Getting rid of the LineNumberReader sounds like a more appropriate solution that I didn't want to do :)

jnt0r:
@joaodinissf your solution looks good to me. Also tested it with the test case I provided. I will close this PR as your solution is more appropriate.

PR #2735, opened 2023-07-07
[#2715] Fix PositionReader's handling of files with CRLF line endings

joaodinissf:
The previous implementation of PositionReader extended LineNumberReader. Unfortunately, the implementation of PositionReader.read() compresses line endings. In other words, both "\n" and "\r\n" contribute one character to the document offset count. This behavior is undesired, and can lead to the production of incorrect line and column offsets.

This commit changes PositionReader to extend BufferedReader instead. Now, calling read() always reads exactly one character. We also implement logic specific to PositionReader to correctly handle both "\r\n" and "\n" line endings.

The PositionReader class now also keeps track of both the line and column values internally. The mark() and reset() functions have been updated accordingly, to correctly fulfill the contract.


szarnekow (approved):
The change looks good to me.

PR #2738, opened 2023-07-11
Restore old exception handling for

cdietrich:
Restore old exception handling for

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014, 2020 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2014, 2023 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -65,8 +65,8 @@ public class TaskMarkerContributor implements IMarkerContributor {
 	public void deleteMarkers(IFile file, IProgressMonitor monitor) {
 		try {
 			file.deleteMarkers(TaskMarkerTypeProvider.XTEXT_TASK_TYPE, true, IResource.DEPTH_ZERO);
-		} catch (Throwable e) {
-			throw new RuntimeException(e);
+		} catch (CoreException e) {
+			LOG.error(e.getMessage(), e);
 		}
 	}
 }


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.java
```diff
@@ -65,8 +66,8 @@ protected void createTaskMarkers(IFile file, List<Task> tasks, IProgressMonitor
 	public void deleteMarkers(IFile file, IProgressMonitor monitor) {
 		try {
 			file.deleteMarkers(TaskMarkerTypeProvider.XTEXT_TASK_TYPE, true, IResource.DEPTH_ZERO);
-		} catch (Throwable e) {
-			throw new RuntimeException(e);
+		} catch (CoreException e) {
+			Exceptions.sneakyThrow(e);
```
cdietrich, line 70: excedption handling is still bogus in 
org.eclipse.xtext.builder.builderState.MarkerUpdaterImpl.processDelta(Delta, ResourceSet, IProgressMonitor)
case where delete markers is dirctly called. wonder if we should catch and log there too

cdietrich: wonder how to run in the delta.getNew() == case

cdietrich: https://ci.eclipse.org/xtext/job/xtext/job/cd_restore_deleteMarkers_b/1/ 
seems to be ran by some tests

cdietrich: Bildschirmfoto 2023-07-11 um 17 19 25
exception will fall though, be caught and rethrown here
and finally handled here
Bildschirmfoto 2023-07-11 um 17 21 04
this makes limited sense to me. @szarnekow what do you think?

cdietrich: other code like
org.eclipse.xtext.ui.validation.MarkerEraser.deleteValidationMarkers(IFile, CheckMode, IProgressMonitor)
org.eclipse.xtext.ui.validation.MarkerEraser.deleteValidationMarkers(IFile, CheckMode, IProgressMonitor)
directly catch and log
=> propose to change to

```
@Override
	public void deleteMarkers(IFile file, IProgressMonitor monitor) {
		try {
			file.deleteMarkers(TaskMarkerTypeProvider.XTEXT_TASK_TYPE, true, IResource.DEPTH_ZERO);
		} catch (CoreException e) {
			LOG.error(e.getMessage(), e);
		}
	}
	```

szarnekow: I agree. We shouldn't fail miserably because some marker wasn't created

cdietrich: thx. updated the pr

PR #2741, opened 2023-07-13
update to guava 32.1.1

cdietrich:
update to guava 32.1.1


cdietrich:
replaced by https://github.com/eclipse/xtext/pull/2751

PR #2747, opened 2023-07-17
fix gradle 8.2 deprecations

cdietrich:
fix gradle 8.2 deprecations


cdietrich:
also fixes https://github.com/eclipse/xtext/issues/2748

Issue #2750, opened 2023-07-18
SemanticTokensService does not corretly implement LSP specification


jnt0r:
The SemanticTokensService implementation does not correctly implement the LSP specification in two points.

1. It only supports the three TokenTypes: KEYWORD_ID, DEFAULT_ID, NUMBER_ID. In the specification there are way more TokenTypes.
2. When converting the TokenType from String to index there is always an offset of 1 added which is not correct.
```
private int getTokenType(final String[] ids) {
    for (String id : ids) {
	Integer index = tokenTypes.indexOf(id);
	if (index != -1) {
		return index + 1;
	}
    }
    return 0;
}
...
Integer positionTokenType = getTokenType(lightweightPosition.getIds());
Integer positionTokenModifiers = getTokenModifiers(lightweightPosition.getIds());

if (positionTokenType != 0 || positionTokenModifiers != 0) {
    int deltaOffset = lightweightPosition.getOffset() - lastOffset;
    ...
```
The specification says: 
Token types are looked up by index, so a tokenType value of 1 means tokenTypes[1]. 
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens

I will open a PR to fix these issues, but I wanted to address them as a problem first.

PR #2753, opened 2023-07-19
Adjust SemanticTokensService to LSP specification

jnt0r:
TokenType is 0 based and represents the index in the list of supported token types.

0 is valid for TokenType and TokenModifiers, so we do not need to check for 0 here.

I added all supported TokenTypes and TokenModifiers from the specification to the List of possible types and modifiers.

### DIFF
--- a/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/syntaxcoloring/SemanticHighlightingCalculator.java
+++ b/org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/syntaxcoloring/SemanticHighlightingCalculator.java
@@ -9,11 +9,15 @@
 package org.eclipse.xtext.ide.tests.testlanguage.syntaxcoloring;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.lsp4j.SemanticTokenModifiers;
+import org.eclipse.lsp4j.SemanticTokenTypes;
 import org.eclipse.xtext.Keyword;
+import org.eclipse.xtext.TerminalRule;
 import org.eclipse.xtext.ide.editor.syntaxcoloring.DefaultSemanticHighlightingCalculator;
 import org.eclipse.xtext.ide.editor.syntaxcoloring.HighlightingStyles;
 import org.eclipse.xtext.ide.editor.syntaxcoloring.IHighlightedPositionAcceptor;
 import org.eclipse.xtext.ide.tests.testlanguage.testLanguage.TestLanguagePackage;
+import org.eclipse.xtext.ide.tests.testlanguage.testLanguage.TypeDeclaration;
 import org.eclipse.xtext.nodemodel.ICompositeNode;
 import org.eclipse.xtext.nodemodel.INode;
 import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
@@ -26,7 +30,7 @@ public class SemanticHighlightingCalculator extends DefaultSemanticHighlightingC
 	@Override
 	protected boolean highlightElement(EObject object, IHighlightedPositionAcceptor acceptor,
 			CancelIndicator cancelIndicator) {
-		if (object.eClass() == TestLanguagePackage.Literals.TYPE_DECLARATION) {
+		if (object instanceof TypeDeclaration) {
 			ICompositeNode node = NodeModelUtils.getNode(object);
 			for (INode childNode: node.getChildren()) {
 				if (childNode.getGrammarElement() instanceof Keyword) {
@@ -35,6 +39,9 @@ public class SemanticHighlightingCalculator extends DefaultSemanticHighlightingC
 						acceptor.addPosition(childNode.getOffset(), childNode.getLength(), HighlightingStyles.KEYWORD_ID);
 					}
 				}
+				NodeModelUtils.findNodesForFeature(object, childNode.getGrammarElement().eClass().getEStructuralFeature("name")).forEach(n -> {
+					acceptor.addPosition(n.getOffset(), n.getLength(), HighlightingStyles.TYPE_ID, HighlightingStyles.DEFINITION_ID);
+				});
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


jnt0r:
fixes #2750

jnt0r:
One thing not working yet is the "multilineTokenSupport". VS Code for example does not support multiline tokens but this implementation uses multiline tokens by default.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semantictokens/SemanticTokensService.java
```diff
@@ -54,13 +55,42 @@ protected SemanticTokensService() {
 	}
 
 	protected void addTokenTypes() {
-		tokenTypes.add(HighlightingStyles.KEYWORD_ID);
-		tokenTypes.add(HighlightingStyles.DEFAULT_ID);
-		tokenTypes.add(HighlightingStyles.NUMBER_ID);
+		tokenTypes.add(SemanticTokenTypes.Namespace);
```
szarnekow, line 58: Thank you for completing the list.
I've to say I don't quite understand this protected method - given that the private field tokenTypes is not accessible, I don't see how a client might override it. Can this be private?

szarnekow: Ok, I missed to getter during the review. Forget the comment.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semantictokens/SemanticTokensService.java
```diff
@@ -105,20 +135,20 @@ public SemanticTokens semanticTokensFull(final Document document, final XtextRes
...
-					data.add(deltaLine); // delta line
-					lastLine = position.getLine();
-					// delta start relative to previous token if on the same line or to 0
-					data.add(deltaLine == 0 ? deltaOffset : position.getCharacter());
-					lastOffset = lightweightPosition.getOffset();
-					data.add(lightweightPosition.getLength()); // length
-					data.add(positionTokenType); // token type
-					data.add(positionTokenModifiers); // token modifiers
-				}
+				int deltaOffset = lightweightPosition.getOffset() - lastOffset;
```
szarnekow, line 138: After this change, I wonder how you'd indicate a position with an ID that is specifically not compatible to the LSP token types? E.g. another UI might support more token types.

I think that's why `0` was used as a return type for an ID that's semantic but not supported by the LSP protocol.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semantictokens/SemanticTokensService.java
```diff
@@ -105,20 +135,20 @@ public SemanticTokens semanticTokensFull(final Document document, final XtextRes
...
+				
+				int deltaLine = position.getLine() - lastLine;
+				lastLine = position.getLine();
+				lastOffset = lightweightPosition.getOffset();
+				
+				data.add(deltaLine); // delta line
+				// delta start relative to previous token if on the same line or to 0
+				data.add(deltaLine == 0 ? deltaOffset : position.getCharacter());
+				data.add(lightweightPosition.getLength()); // length
+				data.add(positionTokenType); // token type
```
szarnekow, line 150: I think here is where we missed a `positionTokenType - 1` to compensate for the chosen communication channel `0 == not contained in list`

org.eclipse.xtext.ide.tests/testlang-src/org/eclipse/xtext/ide/tests/testlanguage/syntaxcoloring/SemanticHighlightingCalculator.java
```diff
@@ -9,32 +9,39 @@
...
 			CancelIndicator cancelIndicator) {
-		if (object.eClass() == TestLanguagePackage.Literals.TYPE_DECLARATION) {
+		if (object instanceof TypeDeclaration) {
 			ICompositeNode node = NodeModelUtils.getNode(object);
 			for (INode childNode: node.getChildren()) {
 				if (childNode.getGrammarElement() instanceof Keyword) {
 					String nodeText = childNode.getText();
 					if (nodeText.equals("extends") || nodeText.equals("type")) {
-						acceptor.addPosition(childNode.getOffset(), childNode.getLength(), HighlightingStyles.KEYWORD_ID);
+						acceptor.addPosition(childNode.getOffset(), childNode.getLength(), SemanticTokenTypes.Keyword);
```
szarnekow, line 39: The highlighting calculator is meant to be agnositic to the UI frontend, being it Eclipse or LSP. Of course an implementor might chose to use SemanticTokenTypes directly, but the underlying idea is to decouple these from each other. This change here renders the approach pointless. I'm not convinced, that it is the right solution.

cdietrich:
@jnt0r @szarnekow what is the state of this pr?

jnt0r:
I currently don't have much time. Will have to take a look at it in a week.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semantictokens/SemanticTokensService.java
jnt0r: the problem is, that 0 is a valid tokenTypeIndex in LSP. I now changed it to return -1 if no valid TokenType is found. So the TokenType with index 0 can be used and still we can detect unsupported types.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/semantictokens/SemanticTokensService.java
jnt0r: to use 0 as an indicator for not supported tokenTypes we would need to add a placeholder tokenType in the list of supported types. Otherwise we would not be able to use the first token Type in the list of supported types. That's why I changed the indicator to -1

cdietrich:
@jnt0r can you please rebase and sequash commits into 1

jnt0r:
@cdietrich I don't quite understand why the build failed. Looks like some build setup and no code issue...
Can you take a look?

cdietrich:
As I said you need to rebase

jnt0r:
Okay. Thought I already did it. Seems to work now.

swissiety:
Thank you all for working on this problem - my research project will definitely benefit from it. I hope this can be merged soon.

szarnekow (approved):
Thank you!

jnt0r:
What's the state of this?

jnt0r:
@szarnekow can this be merged?

jnt0r:
ping @cdietrich @szarnekow

szarnekow:
Thanks again. I merged the changes @jnt0r

PR #2762, opened 2023-07-25
adapt to latest orbit changes

cdietrich:
adapt to latest orbit changes


LorenzoBettini:
I was wondering whether for junit 4 (pretty stable and no major version change) we could simply use version 0.0.0. (in my projects I've always done like that)

Issue #2318, opened 2022-08-03
Xtend Active annotation that remove Declaration raise BatchLinkableResourceStorageWritable messages


dvojtise:
I have a project  where I move some declaration (FieldDeclaration and MethodDeclaration) from one class to another using xtend active annotations

This was working fine with xtend 2.25  but since I moved to xtend 2.27  I get a lot of errors in the Error log such as:

```
org.eclipse.xtext.xbase.resource.BatchLinkableResourceStorageWritable  - Object (JvmParameterizedTypeReference:  (type uri: platform:/resource/test_remove_with_annotation/src/test_remove_with_annotation/TestClass.xtend#|3)) is not contained in any resource
```

I've extracted a subset of the active annotation that illustrate the issue (on Field only  but the same occurs when deleting Method)
Annotation code:
```
@Retention(RetentionPolicy::RUNTIME)
@Active(typeof(MyAnnotationProcessor)) 
@Target(ElementType.TYPE)
annotation  MyAnnotation {
	
}

class MyAnnotationProcessor extends AbstractClassProcessor {

	/**
	 * Phase 2: Transform class' fields and methods
	 */
	override doTransform(List<? extends MutableClassDeclaration> classes, extension TransformationContext context) {
	    val	List<MutableFieldDeclaration> toRemove = newArrayList
		for (clazz : classes) {
			// remove fields starting with "_"
			for (f : clazz.declaredFields) {
				if(f.simpleName.startsWith("_")) {
					toRemove.add(f)
				}
			}
		}
		toRemove.forEach[ f | f.remove]
	}
}
```
The code generated in `xend-gen` is Ok (cf. screenshot

Annotation project (run in the first Eclipse):
[annotation_projet.zip](https://github.com/eclipse/xtext-xtend/files/9253044/annotation_projet.zip)

Annotated project (run in the runtime workbench):
[annotated_projet.zip](https://github.com/eclipse/xtext-xtend/files/9253049/annotated_projet.zip)

![image](https://user-images.githubusercontent.com/661468/182652692-89fad3aa-364f-43d6-a486-ee8734ab2f98.png)

dhuebner:
@cdietrich 
> side effect of https://github.com/eclipse/xtext-extras/issues/695 ?!?

In 695 it was just about to prevent NPE which was caused by the missing clean ups when removing something from the JvmModel. In my case it was removing the default constructor, if I recall it correctly. Overall I think, yes it the same problem.

PR #2776, opened 2023-08-09
[#2318] Clean AllAssociation when removing field

dvojtise:
This PR contributes to https://github.com/eclipse/xtext/issues/2318

- adds a test highlighting the issue  (the test `testresourceStorageWrite` fails if the fix below isn't applied)
- applies the suggested correction in `org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend`

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend
@@ -1320,6 +1320,8 @@ class JvmFieldDeclarationImpl extends JvmMemberDeclarationImpl<JvmField> impleme
 	}
 	
 	override remove() {
+		compilationUnit.jvmModelAssociator.removeAllAssociation(delegate.type)
+		compilationUnit.jvmModelAssociator.removeAllAssociation(delegate)
 		compilationUnit.jvmModelAssociator.removeLogicalChildAssociation(delegate)
 		super.remove()
 	}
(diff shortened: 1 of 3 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend
```diff
@@ -1320,6 +1320,8 @@ class JvmFieldDeclarationImpl extends JvmMemberDeclarationImpl<JvmField> impleme
 	}
 	
 	override remove() {
+		compilationUnit.jvmModelAssociator.removeAllAssociation(delegate.type)
```
szarnekow, line 1323: There might be other `removevAllAssociations(..)` missing in this class, no?

cdietrich:
can you also have a look at the failing test?

org.eclipse.xtend.core/src/org/eclipse/xtend/core/macro/declaration/JvmBasedDeclarations.xtend
dvojtise: yes probably on other kind of remove 
but since it make silently crash other remove (see the new failed test. after the first remove. the other ones aren't processed)

I'm not sure anymore that it should be done that way

I must admit I don't master xtend internal structures and I'm not efficient in finding a solution for cleaning these structures on all remove situations

szarnekow: I'll give it a stab later this week.

PR #2810, opened 2023-09-28
remove xpand based Xtext2Ecore Postprocessor (Xpand removal part 1)

cdietrich:
see #2811


org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/XtextUiModule.java
```diff
@@ -141,12 +141,6 @@ public void configureFilterTerminalRulesContribution(Binder binder) {
 
 	@Deprecated(forRemoval = true)
 	public void configureIXtext2EcorePostProcessor(Binder binder) {
```
szarnekow, line 143: The entire method can be removed, since it is in xtext.xtext.ui

org.eclipse.xtext/src/org/eclipse/xtext/XtextRuntimeModule.java
```diff
@@ -122,14 +122,8 @@ public Class<? extends IValueConverterService> bindIValueConverterService() {
 	/**
 	 * @since 2.9
 	 */
-	@SuppressWarnings({"removal","deprecation"})
+	@Deprecated
 	public void configureIXtext2EcorePostProcessor(Binder binder) {
```
szarnekow, line 126: Likewise.

szarnekow (approved):

org.eclipse.xtext/src/org/eclipse/xtext/XtextRuntimeModule.java
cdietrich: i wanted to keep these in a first step and do their removal in a 2nd stage

PR #2825, opened 2023-10-20
use import for classgraph dependency

cdietrich:
@szarnekow wdyt


cdietrich:
Should we also remove imports of bundle from features .

PR #2827, opened 2023-10-27
Refactor IncrementalBuilder and Indexer.

andrewL-avlq:
Refactor of the incremental builder and indexer to make it easier to write custom builders that calculate the affected resources separately from indexing the changed and deleted sources in the build request.


cdietrich:
@szarnekow what do you think?

PR #2832, opened 2023-10-30
Ignore NumberLiterals without CompositeNode in HighlightingCalculator

HannesWell:
We have a customized language where we inject extra objects into the AST that don't have a corresponding text node.
Consequently the node is null, which breaks the standard `XbaseHighlightingCalculator`. In all other cases null nodes are ignored too.


cdietrich (approved):

HannesWell:
Thanks Christian.

PR #2840, opened 2023-11-12
Update `XtextFormattingTokenSerializer` references in JavaDocs

wborn:
The `XtextFormattingTokenSerializer` referred to in JavaDocs no longer exists and it seems to have been refactored into the [`XtextFormatter`](https://github.com/eclipse/xtext/blob/main/org.eclipse.xtext/src/org/eclipse/xtext/xtext/XtextFormatter.java).


szarnekow (approved):
Thank you!

PR #2842, opened 2023-11-15
Collect outgoing siblings for UnorderedGroup

rubenporras:
CollectOutgoingByContainer will stop collecting the next siblings if the next to it is mandatory, because it is expected to appear in that order. Handle unordered groups as well to collect all of the outgoing siblings.


rubenporras:
For https://github.com/eclipse/xtext/issues/2843

szarnekow:
@rubenporras Can you please add a test for this?

rubenporras:
Do you have a suggestion how that could be tested? I are not very experienced with how the Xtext tests are structured and what can be easily done and what not.

szarnekow:
Sorry for the unreasonable delay. I just read the code here again, and I think it is reasonable. A test class to start with would be org.eclipse.xtext.parsetree.reconstr.TreeConstTest - are you still available to add a test case for unordered groups?

rubenporras:
Sorry also for the long delay. We have been very busy upgrading Xtext due to the release of the old Xtext generator fragments. I might have time to look at the test now.

PR #2851, opened 2023-11-21
Bump Xtext version to 2.34

szarnekow:


cdietrich:
looks like the same problem as in mwe: search and replace also catched others (in this case args4j)

PR #2875, opened 2023-12-18
Replace dependency to org.eclipse.osgi.services

HannesWell:
The bundle `o.e.osgi.services` is deprecated for removal as part of https://github.com/eclipse-equinox/equinox/issues/18 and already has been changed to be empty and only re-export the 'official' OSGi artifacts published to Maven-Central.

It should be replaced by imports of the desired org.osgi.service.* packages so that the runtime can choose a suitable provider independently of the bundle name.


HannesWell:
@cdietrich or @szarnekow could you please have a look?

szarnekow (approved):

szarnekow:
Thank you, @HannesWell

PR #2878, opened 2023-12-20
Update to Guava 33

cdietrich:
Closes #2876

PR #2882, opened 2023-12-21
Restore common types and Xbase generators

### LINKED ISSUE #2881: Restore common types and Xbase generators
Currently, they are in a non-source folder (dependencies are also to be updated), see also https://github.com/eclipse/xtext/issues/2880#issuecomment-1864919281

LorenzoBettini:
Closes #2881 

WARNING: to have something buildable in the CI, I based my branch on top of the update to the new Guava. Before merging, I can rebase on master once #2878 is merged.

For common types, I created a new MWE2 since there's no need to run it as a Java application as it was done before in `GenerateEMF.java`.

For Xbase, I simply restored `generator` as a source folder, and I kept the Java generator (that's too big and I didn't dare to port it as a MWE2), though there was an old reference to `xtext-eclipse` folder (a remainder of the split repo), which I fixed.

I also regenerated the Java classes: there are only a few white spaces in a few classes. I thought it was worthwhile to commit them.


LorenzoBettini:
@cdietrich I rebased the PR, which is now ready to review

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/GenerateCommonTypes.mwe2
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
```
szarnekow, line 1: I'd prefer if we wouldn't ship this workflow file.

org.eclipse.xtext.xbase/generator/org/eclipse/xtext/xbase/GenerateXbase.java
```diff
@@ -52,7 +52,7 @@ public static void main(String[] args) {
 		final String root = "..";
```
szarnekow, line 52: Is this file excluded from the binary jar?

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/GenerateCommonTypes.mwe2
szarnekow: Since MWE2 uses common.types, I'd prefer if we could regenerate without an MWE dependency. Otherwise changes to the types model would require MWE to run successfully before other changes to the types model can be applied.

LorenzoBettini:
@szarnekow so you would suggest going back to the Java file generator for common types?

Moreover, looking at your other comment on GenerateXbase, if that generator folder is set as a source folder then it will be part of the binary JAR; the same for Generate common types if we go back to the Java version.

If we don't want to put the Java generator file in the binary JAR, I can restore the previous situation where the generator Java files were not in a source folder. To run those files, we should temporarily set the "generator" folder as a source folder, run the generator, and then restore the "generator" folder as a NON-source folder. In such a case, maybe we could leave a README in those projects.

szarnekow:
Would a separate output folder for the src/generator as well as a bin.exclude (? is there such a thing?) help?

LorenzoBettini:
Let's see: the `build.properties` editor does not complain about `bin.excludes`. I'll update the PR and see whether the produced JAR are as expected.

In any case, I go back to the Java code for generating common types, right?

Note however that the `additional.bundles` do require MWE

LorenzoBettini:
@szarnekow I restored the Java file for common types (I renamed it to `GenerateCommonTypes.java`) and I excluded the `generator` folder with `bin.excludes` both in common.types and in xbase.

The Jenkins build is https://ci.eclipse.org/xtext/job/xtext/job/lb_2881/lastSuccessfulBuild/artifact/build/p2-repository/plugins/. I downloaded the JARs of xbase and common types and the `GenerateCommonTypes` and `GenerateXbase` are NOT there:

```
❯ jar -tf org.eclipse.xtext.xbase_2.34.0.v20231230-2015.jar | grep Generate
org/eclipse/xtext/xbase/annotations/XbaseWithAnnotationsStandaloneSetupGenerated.class
org/eclipse/xtext/xbase/XbaseStandaloneSetupGenerated.class
org/eclipse/xtext/xbase/XtypeStandaloneSetupGenerated.class
❯ jar -tf org.eclipse.xtext.common.types_2.34.0.v20231230-2015.jar | grep Generate
```

So, it worked! :)

szarnekow (approved):
Nice 👍

PR #2888, opened 2023-12-31
Some Xbase cleanup

LorenzoBettini:
The two commits' messages describe the proposed changes (it's a small cleanup)

PR #2889, opened 2024-01-02
don't retrieve the ImportManager by reflection

LorenzoBettini:
Since we already assume that the import manager is part of the appendable, which is a TreeAppendable, I think it's better to do the case and expose getImportManager in TreeAppendable than to call that method with two reflective calls.
This would also allow clients to provide a custom TreeAppendable; with the reflective call this latter use-case would not be supported

### DIFF
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/JvmModelGenerator.xtend
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/JvmModelGenerator.xtend
@@ -804,7 +804,7 @@ class JvmModelGenerator implements IGenerator {
 				generateDocumentation(adapter.documentation, emptyList, appendable, config)
 			}
 		}
-	} 
+	}
 	
 	def addJavaDocImports(EObject it, ITreeAppendable appendable,List<INode> documentationNodes) {
 		for(node : documentationNodes){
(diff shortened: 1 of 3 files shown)


LorenzoBettini:
@szarnekow I'll split https://github.com/eclipse/xtext/pull/2888 into separate PRs.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/JvmModelGenerator.xtend
```diff
@@ -839,12 +839,7 @@ class JvmModelGenerator implements IGenerator {
 	}
 
 	def getImportManager(ITreeAppendable appendable) {
```
szarnekow, line 841: Should the XbaseCompiler work with TreeAppendable instead of ITreeAppendable on its APIs?

LorenzoBettini: @szarnekow from what I see and understand, the XbaseCompiler does not need anything more than what the `ITreeAppendable` interface provides and doesn't seem to make further assumptions.
On the contrary, `JvmModelGenerator` completely relies on `TreeAppendable`.

szarnekow: Meh sadness. We cannot easily change signatures here. Casting seems the best we can do.

szarnekow (approved):

Issue #2886, opened 2023-12-30
XtendJvmModelInferrer and treatment of anonymous classes


LorenzoBettini:
The Xtend inferrer does this:

```java
	public void inferLocalClass(
			AnonymousClass anonymousClass,
			String localClassName,
			JvmFeature container) {
		final JvmGenericType inferredType = typesFactory.createJvmGenericType();
		inferredType.setSimpleName(localClassName);
		inferredType.setAnonymous(!hasAdditionalMembers(anonymousClass));
...
	}
	
	protected boolean hasAdditionalMembers(AnonymousClass anonymousClass) {
		for(XtendMember member: anonymousClass.getMembers()) {
			if(member instanceof XtendField ||	
				(member instanceof XtendFunction && !((XtendFunction) member).isOverride())) 
				return true;
		}
		return false;
	}
```

Thus, an anonymous Xtend class is NOT mapped to a `JvmGeneritcType.isAnonymous` if it contains a field or a non-overridden method.

Is there a specific reason?

While working on https://github.com/eclipse/xtext/issues/2880 I'd like to analyze also JvmGenericTypes corresponding to anonymous classes, but with the above behavior, I cannot correctly detect them: if `isAnonymous` returns `false` it can still correspond to an anonymous class. Some checks on anonymous classes must be different from standard classes or interfaces. First, an anonymous class's supertype can be a class or interface. For standard classes, I have to distinguish between the intended superclass and the intended implemented interfaces.

cc @szarnekow

szarnekow:
Even though a class is an anonymous class in Xtend, it’s not necessary possible to compile it to an anonymous Java class. In some cases , it’s necessary to compile to a named, local class.

LorenzoBettini:
@szarnekow but if I understand correctly, that's only a code generation strategy; I mean, that check could be delayed to code generation instead of at the JVM inferer level. I did a quick experiment (setting isAnonymous to true) and the failing test only concern code generation. Can it be?

szarnekow:
I might be confused. The JVM model is the input model for the code generator, no?

LorenzoBettini:
[...]

PR #2890, opened 2024-01-03
Xbase/Xtend refactoring

LorenzoBettini:
This provides some general refactoring for Xbase and Xtend and as a preparation for https://github.com/eclipse/xtext/issues/2886, for which I'll also provide a PR soon.

The main focus is on the treatment of anonymous classes, which was originally split into Xbase and Xtend. In this PR, anonymous classes are handled by the XbaseCompiler completely, which now provides a new method https://github.com/eclipse/xtext/compare/main...LorenzoBettini:xtext:lb_xbase_cleanup-1?expand=1#diff-3c98e6c5688a98e8fc686a884a0462f80d32c824ea233925c42b813a15345743R1148 `canCompileToJavaAnonymousClasses`. This relies on the `JvmGenericType.isAnonymous`. This was done before by `XbaseCompiler`, though with some duplicated code. Now, `XbaseCompiler` consistently calls this method to treat other situations that in the past were customized in `XtendCompiler`.

For example, `isVariableDeclarationRequired` is now properly handled in the `XbaseCompiler` also for constructor calls, which possibly refer to an anonymous class and does not need to be customized by `XtendCompiler`.

Similarly, `_toJavaStatement(final XConstructorCall expr...` does not need to be redefined by the `XtendCompiler`, which only needs to redefine `constructorCallToJavaExpression` to handle the generation of its anonymous classes.

I also took the chance to restrict the visibility of a few methods in `XtendCompiler`, which are used only internally.

The build is green.

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
@@ -576,20 +576,7 @@ public class XtendCompiler extends XbaseCompiler {
 		} else
 			return super.internalCanCompileToJavaExpression(expression, appendable);
 	}
-	
-	@Override
-	protected boolean isVariableDeclarationRequired(XExpression expr, ITreeAppendable b, boolean recursive) {
-		if (expr instanceof XConstructorCall) {
-			XConstructorCall constructorCall = (XConstructorCall) expr;
-			if (constructorCall.isAnonymousClassConstructorCall()) {
-				EObject container = expr.eContainer();
-				return isVariableDeclarationRequired((XExpression) container, b, recursive) &&
-					!canCompileToJavaAnonymousClass(constructorCall);
-			}
-		}
-		return super.isVariableDeclarationRequired(expr, b, recursive);
-	}
-	
+
 	@Override
 	/* @Nullable */
 	protected String getReferenceName(XExpression expr, ITreeAppendable b) {
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
@@ -357,7 +357,7 @@ public class XtendCompiler extends XbaseCompiler {
 	}
 	
 	@Override
-	public void doInternalToJavaStatement(XExpression obj, ITreeAppendable appendable, boolean isReferenced) {
+	protected void doInternalToJavaStatement(XExpression obj, ITreeAppendable appendable, boolean isReferenced) {
 		prependLocalTypesIfFieldInitializer(obj, appendable, isReferenced);
 		if(obj instanceof AnonymousClass) 
 			_toJavaStatement((AnonymousClass)obj, appendable, isReferenced);
@@ -374,7 +374,7 @@ public class XtendCompiler extends XbaseCompiler {
 		}
 	}
 	
-	public void _toJavaStatement(RichString richString, ITreeAppendable b, boolean isReferenced) {
+	private void _toJavaStatement(RichString richString, ITreeAppendable b, boolean isReferenced) {
 		LightweightTypeReference actualType = getLightweightType(richString);
 		b = b.trace(richString);
 		if (actualType.isType(StringConcatenationClient.class)) {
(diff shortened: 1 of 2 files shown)


org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
```diff
@@ -375,7 +374,7 @@ private void prependLocalTypesIfFieldInitializer(XExpression obj, ITreeAppendabl
 		}
 	}
 	
-	public void _toJavaStatement(RichString richString, ITreeAppendable b, boolean isReferenced) {
+	private void _toJavaStatement(RichString richString, ITreeAppendable b, boolean isReferenced) {
```
szarnekow, line 377: I'd prefer this to remain at least protected since I'm aware of languages that inherit from Xtend.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
```diff
@@ -426,19 +425,27 @@ else if (obj instanceof RichString)
 			super.internalToConvertedExpression(obj, appendable);
 	}
 	
-	protected void _toJavaExpression(AnonymousClass anonymousClass, ITreeAppendable b) {
+	private void _toJavaExpression(AnonymousClass anonymousClass, ITreeAppendable b) {
```
szarnekow, line 428: Same

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
```diff
@@ -1106,7 +1105,9 @@ protected void constructorCallToJavaExpression(final XConstructorCall expr, ITre
 		}
 		ITreeAppendable typeAppendable = appendableWithNewKeyword.trace(expr, XbasePackage.Literals.XCONSTRUCTOR_CALL__CONSTRUCTOR, 0);
 		appendConstructedTypeName(expr, typeAppendable);
-		if (hasTypeArguments || (expr.isAnonymousClassConstructorCall() && !explicitTypeArguments.isEmpty() && ((JvmGenericType) constructor.getDeclaringType()).isAnonymous())) {
+		if (hasTypeArguments ||
```
szarnekow, line 1108: Nit: The formatting is a bit weird here.

szarnekow (approved):

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
LorenzoBettini: OK, I'll update that. I thought Xtend was not meant to be API ;)

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
LorenzoBettini: I thought splitting into separate lines could be more readable, but I can revert and put everything on a single line, as you prefer.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
szarnekow: True, yet no reason to do unwarranted changes that can be easily avoided.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/XbaseCompiler.java
szarnekow: No strong preference.  Please keep as you prefer if it’s intentional

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompiler.java
LorenzoBettini: OK, I reverted to protected methods here https://github.com/eclipse/xtext/pull/2890/commits/7ac3f496fd906b5cdbeef5932ff0d17388d405b6

PR #2891, opened 2024-01-05
Moved Xtend core performance tests in xtend.performance.tests project

LorenzoBettini:
In `xtend.core.tests` there are two performance test `PerformanceTest` and `XtendIncrementalBuilderPerformanceTest` that are "a bit in the way" when running tests in `xtend.core.tests`: there's no easy way to exclude them and the launch configurations were meant for the old gradle age, so they don't help anymore.

This PR:

- move those tests in the `xtend.performance.tests` project; in that project there are already a few performance tests; those are for the UI, but the above mentioned tests are still related to building so it does not harm to put them in the performance.tests project; I also renamed the name in the POM without the "IDE"
- update the launch configurations in `xtend.core.tests`: there are only two of them, one for the Java files and one for the Xtend files.
- The `XtendIncrementalBuilderPerformanceTest` used to be in a package that ends with `build` so its directory it's excluded by the main `.gitignore` file; it took me a while to understand what was going on. To avoid problems in the future, I renamed its package to `building`.
- removed `RuntimePerformanceSuite.java` since it didn't seem to make sense


szarnekow:
@LorenzoBettini Would Junit tags be option instead of splitting the project? See org.junit.jupiter.api.Tag

LorenzoBettini:
@szarnekow from my understanding JUnit 4 does not provide tags: it provides categories that require proper suite classes. In fact, Eclipse allows for tag selection/exclusion only for JUnit 5.
Thus, for JUnit 4 this would not help.
I'd also avoid switching to JUnit 5 which gave a few headaches in the past.
What do you think?

szarnekow:
Sadness. Ok, I'm fine with a separate test project or with a separate source folder for. the performance tests. Whatever you prefer.

LorenzoBettini:
@szarnekow I had also thought about a separate source folder in `xtend.core.tests` but that would work seamlessly for Java files: the two performance tests are Xtend files, so they are generated into `xtend-gen` together with the other ones, so the launch configuration could not distinguish them.

LorenzoBettini:
I spotted other performance tests also in xtend.ide.tests; I'd like to move them as well to the project performance.tests

szarnekow:
> I'd like to move them as well

Please do

PR #2892, opened 2024-01-07
Harden bundle constraints between Xtext bundles

szarnekow:
Also:
- Export versioned packages throughout the code base
- Some fixes to MergableManifest2
- Added a Java based script to bump the Xtext version honoring the Manifest semantics without regex foo
- Aligned version of xtext.logging fragment with version number of Xtext rather than the version number of the host bundle


cdietrich (approved):

org.eclipse.xtext.logging/pom.xml
```diff
@@ -9,13 +9,6 @@
 	</parent>
 	<packaging>eclipse-plugin</packaging>
 	<artifactId>org.eclipse.xtext.logging</artifactId>
-	<version>1.2.24-SNAPSHOT</version>
```
LorenzoBettini, line 12: @szarnekow I think that's required, otherwise, the Maven version will not match with the MANIFEST version `1.2.24.qualifier`, and since this is a fragment, it should have the same version of the log4j bundle.

org.eclipse.xtext.tests/testdata/wizard-expectations/lsMavenTychoApp/lsMavenTychoApp.parent/pom.xml
```diff
@@ -79,7 +79,7 @@
 							<requirement>
 								<type>eclipse-plugin</type>
 								<id>org.eclipse.xtext.logging</id>
-								<versionRange>1.2.24</versionRange>
+								<versionRange>${xtextVersion}</versionRange>
```
LorenzoBettini, line 82: see my comment above about the xtext.logging version

LorenzoBettini (commented):
Sorry for being late ;)

I think the updates to the `xtext.logging` version are wrong: that is a fragment that has to match the version of the log4j bundle. At least, that's what @cdietrich told me during the porting to the monorepo.

Moreover, I don't remember the rationale for this explicit versioning. You told me something about future Xtext OSGI enhancements; it's related to that, isn't it?

szarnekow:
> Sorry for being late ;)

No worries.

The rationale is that Xtext bundles strictly depend on each other and therefore the OSGI metadata should reflect that. So far we didn't provide this information as part of the Manifest which could have led to wirings in an OSGI container that are not working. 

Exporting packages with proper versions will help users who try to be good OSGI citizens and use import-package rather than require-bundle.

org.eclipse.xtext.logging/pom.xml
LorenzoBettini: @szarnekow as I said in the general comments, it used to be like that in the old days before repo merging and it had to be updated consistently with the version of log4j: @cdietrich told me that a fragment must have the same number of the hosting bundle; maybe it was something we didn't know about or it was like that in the past. I'm happy we can get rid of that inconsistent version :)

cdietrich:
This is also new to me

szarnekow:
> This is also new to me

You mean the versioning of fragments?

cdietrich:
yes we always had the host bundle version there.

szarnekow:
> yes we always had the host bundle version there. and i dont know why

I think back then we didn't know better.

Issue #2351, opened 2022-09-22
Compile error with valid xbase / xtend code


szarnekow:
```xtend
package p

import java.math.BigDecimal

class C {
	def Object m(int i) {
		val r = switch(i) {
			case 1: 1.0 // double
			case 2: 1.0bd // BigDecimal
			default: weird("")
		}
		return r	
	}
	
	private def <T> T weird(Object object) {
		return null;
	}
}
```

The type of `r` is resolved to `double` which is not correct given that `BigDecimal` is not a `double`. The bug depends on the presence of the method call to `weird` which returns basically a blank type parameter which in turn confuses the computation of the least upper bound.

szarnekow:
Fixed by #2894

PR #2894, opened 2024-01-08
[#2351] Fixed type common super type computation when type params are involved

szarnekow:


org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/conformance/TypeConformanceComputer.java
```diff
@@ -252,6 +253,36 @@ protected LightweightTypeReference getCommonSuperType(List<LightweightTypeRefere
 			if (allPrimitiveVoid(types)) {
 				return types.get(0);
 			}
+			
+			// Compute the common super type without looking at any of the UnboundTypeReferences that do
+			// not carry any hints yet.
+			List<LightweightTypeReference> unconstrainedUnboundTypes = null;
+			for(LightweightTypeReference type: types) {
```
LorenzoBettini, line 260: Maybe just nitpicking but why not create the array list here and in 271 check whether it's empty?
This could simplify the rest of the code?

LorenzoBettini (approved):

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/typesystem/conformance/TypeConformanceComputer.java
szarnekow: I was thinking that it's better to avoid the allocation of the array list, but given that recent Java versions don't create the inner array immediately, it can be simplified, true.

szarnekow: Adjusted.

LorenzoBettini (approved):

PR #2897, opened 2024-01-08
Removed cyclic dependency to mwe2.launch

szarnekow:


szarnekow:
FYI @cdietrich

org.eclipse.xtend.core/build.properties
```diff
@@ -13,6 +13,5 @@ exclude.. = **/GenerateXtend.mwe2
 src.includes = about.html
 src.excludes = generator/,\
                generator-xtend-gen/
-additional.bundles = org.eclipse.emf.mwe2.launch,\
```
cdietrich, line 16: did you try if run workflow works? for me it does not.
grafik

szarnekow: Of course, my bad
We need to port the mwe2 file (back) to Java

cdietrich: or we keep launch here

szarnekow: Yes, I just pushed that. I misread the error message. Multitasking is hard.

cdietrich (approved):

LorenzoBettini (approved):

PR #2898, opened 2024-01-08
Xtend refactoring handling of anonymous class

LorenzoBettini:
Another proposal for https://github.com/eclipse/xtext/issues/2886


LorenzoBettini:
This is another working proposal for this issue that does not touch the inferred model.

the changes are rather minimal:

- the inferrer always sets the inferred type for AnonymousClass as anonymous
- a new `XtendCompilerHelper` performs the check the inferred used to do, i.e., the anonymous class does not contain fields or methods that are not overridden
- the `XtendGenerator` uses this helper to know whether it has to create nested local classes instead of anonymous classes; previously, it only checked JvmGenericType.isAnonymous.
- the `XtendCompiler` redefines the new `XbaseCompiler.canCompileToJavaAnonymousClass(XConstructorCall constructorCall)` introduced in https://github.com/eclipse/xtext/pull/2890 and delegates to `XtendCompilerHelper`
- a custom `TreeAppendable`, I called that `AnonymousClassAwareTreeAppendable,` but it could be called `XtendTreeAppendable`, uses a custom `LightweightTypeReferenceSerializer` to decide what to do with JvmGenericType that are "anonymous", again, using the XtendCompiler. This is useful to intercept all the serialization of types related to anonymous classes correctly. While the previous customizations catch anonymous classes in "new" expressions, this one catches the occurrences of related types in other parts, e.g., in local variables in case anonymous classes are compiled into nested local classes.

This way, during the Xtend compilation process, anonymous classes are correctly generated as Java anonymous classes or nested local classes. All the tests are green. All the generated code is as before. No test has been touched apart from this one https://github.com/eclipse/xtext/pull/2898/files#diff-8d2c98464642a8554f494c6f4c1de6f865cbec0d6cb5b288412da41e6067cba1L222 of course.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompilerHelper.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.xtext.util.IResourceScopeCache;
+import org.eclipse.xtext.util.Tuples;
+import org.eclipse.xtext.xbase.jvmmodel.IJvmModelAssociations;
+
+import com.google.inject.Inject;
+
+/**
+ * @author Lorenzo Bettini - Initial contribution and API
+ */
+public class XtendCompilerHelper {
```
szarnekow, line 26: Since this can only answer whether a JvmType can be compiled to an anonymous class, maybe a more restrictive name should be chosen?
`AnonymousClassCompilerHelper`?

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendGenerator.xtend
```diff
@@ -320,7 +330,9 @@ class XtendGenerator extends JvmModelGenerator implements IGenerator2 {
 			}
 			if (declaringType.local && it instanceof JvmOperation) {
 				val declarator = declaringType as JvmGenericType
-				if (!declarator.anonymous) {
+				if (!declarator.anonymous ||
```
szarnekow, line 333: I'm not sure whether I like the fact that the helper doesn't check the declarator itself but this code here has to do both.

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/JvmModelGenerator.xtend
```diff
@@ -1002,11 +1002,20 @@ class JvmModelGenerator implements IGenerator {
...
-		val appendable = new TreeAppendable(importManager, cachingConverter, locationProvider, jvmModelAssociations, context, "  ", "\n")
+		val appendable = createAppendable(importManager, cachingConverter, locationProvider, jvmModelAssociations, context, "  ", "\n")
 		appendable.state.generatorConfig = config
 		return appendable
 	}
 	
+	/**
+	 * @since 2.34
+	 */
+	def TreeAppendable createAppendable(ImportManager importManager, ITraceURIConverter converter,
```
szarnekow, line 1013: Could this be protected?

szarnekow (commented):
Overall, I like this change. Just a few small comments / questions.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompilerHelper.java
LorenzoBettini: Indeed that was my second choice, so I'll rename it :)

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendGenerator.xtend
LorenzoBettini: Indeed I had put a Javadoc comment on the helper https://github.com/eclipse/xtext/pull/2898/files/bc766b0068465148679855a36c8c9c88e5c62065#diff-c6ea2a981e27e70dc2cd53ff982ee051ac65096607807462e10196d625518f88R35
to avoid a check there, but if you prefer we can check `isAnonymous` directly in the helper, if that's what you mean.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendCompilerHelper.java
LorenzoBettini: By the way, we already have https://github.com/eclipse/xtext/blob/88acf1555988cd81e1756abc0d6d2c3f1f1efa11/org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/AnonymousClassUtil.java#L39
should I put the methods of `XtendCompilerHelper` directly there?

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendGenerator.xtend
```diff
@@ -164,9 +171,12 @@ class XtendGenerator extends JvmModelGenerator implements IGenerator2 {
 	}
 	
 	def compileLocalTypeStubs(JvmFeature feature, ITreeAppendable appendable, GeneratorConfig config) {
-		feature.localClasses.filter[ !anonymous ].forEach[
-			appendable.newLine
+		feature.localClasses.forEach[
+			if (compilerHelper.canCompileToJavaAnonymousClass(it)) {
```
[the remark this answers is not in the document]
LorenzoBettini: At least in the current implementation anonymous classes are local classes

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/compiler/JvmModelGenerator.xtend
LorenzoBettini: Right! I think so.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/compiler/XtendGenerator.xtend
LorenzoBettini: @szarnekow I've just pushed https://github.com/eclipse/xtext/pull/2898/commits/6af742c8bceb96792ea6d06325536ed8b0a31209 

that check for anonymous is redundant

szarnekow (approved):

LorenzoBettini:
Closing and re-opening to trigger CI rebuild

Issue #2349, opened 2019-11-05
Provide a JvmGenericTypeValidator


LorenzoBettini:
I'd like to propose an idea I've been thinking of for a while.

When implementing an Xbase language you have to repeat several typical checks, like no duplicate fields, no duplicate methods (according to Java overriding and overloading rules), cyclic hierarchy, etc., like we recently did in https://github.com/eclipse/xtext-eclipse/issues/1205 and https://github.com/eclipse/xtext-eclipse/issues/1205 . I seem to understand that the best way of doing that is by performing checks directly on the inferred JvmModel, like we did in the issues above and like Xtend itself does. After all, it does not make sense to have an inferred JvmModel which then produces incorrect Java code...

So why not providing some kind of `JvmGenericTypeValidator` that does all these checks automatically on the inferred JvmModel and then generating issues the original elements in the program (by means of `IJvmModelAssociations`)? With that respect the starting point could be the `XtendValidator` itself, from where we can extract most of the logic of typical JvmModel validations in the reusable `JvmGenericTypeValidator` (and then `XtendValidator` could be refactored accordingly). The `JvmGenericTypeValidator` could follow the same strategy of `JvmTypeReferencesValidator` which is automatically enabled (by relying on `getEPackages`).

All Xbase languages would then benefit from such automatic validations.

@szarnekow do you think it is feasible, makes sense, could be useful?

LorenzoBettini:
@szarnekow I started to work on this; I was planning to factor many parts of `XtendValidator` into an Xbase `JvmGenericTypeValidator`. Then, I would use `XtendValidatorTest` to check whether errors are still generated as expected, of course, after adapting a few things.

That's the very first experimental commit on my fork: https://github.com/LorenzoBettini/xtext/commit/c490251f7e46f789648409d7cf93492719fd1bda

I set `JvmGenericTypeValidator` as a composite validator in `XtendValidator`.

However, by debugging, my validator never kicks in for inferred elements like `JvmGenericType`. Am I doing anything wrong?
How would such a validator be supposed to be configured so that it can inspect inferred types in a derived state resource?

LorenzoBettini:
@szarnekow, I pushed an updated commit

https://github.com/LorenzoBettini/xtext/commit/7c9717b8e0f23d4d6a93d4158ccc5a30c53717a2

tests are green.

It looks like a small hack because I need to intercept an EObject, which is the root of the Resource's contents, because, as you said, only the first element of the Resource is currently validated.

[...]

PR #2899, opened 2024-01-09
Add features to JvmGenericType to express the expected semantics of supertypes

### LINKED ISSUE #2880: Xbase proposal: Add features to express the expected semantics of supertypes
To properly and cleanly implement #2349, it would be nice to have in the types model proper types for a Jvm class and Jvm interface to distinguish between the intended extended class and implemented interfaces (for classes) and extended interfaces (for interfaces).
Actually, the `JvmGenericInterface` would not be strictly necessary since we can assume that all the set supertypes are meant to be interfaces.
On the contrary, `JvmGenericClass` would be necessary (see https://github.com/eclipse/xtext/issues/2349#issuecomment-1862713233).

My proposed extension is meant to be backward compatible. Speaking about `JvmGenericClass`, I planned to add two fields, `extendedClass` (singular) and `implementedInterfaces` (multiple), and let the inherited `superTypes` be consistent with them. Similarly, for the methods `getExtendedClass` and `getImplementedInterfaces.`

I could provide a PR against the main branch (and then use it in #2349).

@szarnekow, what do you think?

LorenzoBettini:
Closes #2880

I haven't introduced `JvmGenericClass` and `JvmGenericInterface`. Instead, I enriched `JvmGenericType`.
The new features are `classToExtend` and the (non-containment) lists `interfacesToImplement` and `interfacesToExtend`.
I chose new names not to mix them with the get methods like `getExtendedClass`. In fact, these features are meant for expressing the intentions (class to extend and interfaces to implement for classes and interfaces to extend for interfaces) to better perform validation (#2349).

The getter and setter for `classToExtend` and the getters for the lists are customized to keep the new features consistent with what `getSuperTypes` returns. This way, this extension will be backward compatible. It is not a complete synchronization, see the Javadoc https://github.com/eclipse/xtext/compare/main...LorenzoBettini:xtext:lb_2880-2?expand=1#diff-0243bd309c39347e67704b49bdd4b374c1a1a475fefb6beb2ceb1268c5f8a273R20
but it should be enough for the proposed extension.
In particular, adding something directly to `getSuperTypes` does not change the new features because there would be no way to know which one to update.
On the contrary, adding a type with the new features will update the list returned by `getSuperTypes`.
Removing that type from `getSuperTypes` would remove it from the original place as well.
I have added several tests in that respect in `JvmGenericTypeTest`.

I then applied these new features in the `XtendJvmModelInferrer` and the build is still green.

I added a few more tests to `JvmModelGeneratorTest` as well

Finally, I applied that to Domainmodel example and added a test (we were never testing with a superclass) 

Unfortunately, I could not add a check for proper usages of the new features: `classToExtend` and `interfacesToImplement` should not be used when `isInterface` is true; accordingly for `interfacesToExtend`.
Actually, I had added such checks but had to revert them because such a check would break code that scans all the features, e.g.,
org.eclipse.xtext.findReferences.ReferenceFinder.findLocalReferencesFromElement(Predicate<URI>,
EObject, Resource, Acceptor).


szarnekow:
I prefer the option with the adapter.

LorenzoBettini:
Superseeded by https://github.com/eclipse/xtext/pull/2924

PR #2911, opened 2024-01-15
[Xtend] put INCOMPATIBLE_RETURN_TYPE marker on field's type 

LorenzoBettini:
While working on #2349 I discovered that Xtend puts an INCOMPATIBLE_RETURN_TYPE marker on the whole field; this happens, e.g., when an active annotation like `Accessors` is put on a field in a subclass of another class; the `Accessors` creates a getter or setter that might conflict with an inherited getter or setter:

image

This PR, independently from #2349, allows the marker to be put on the incompatible type, just like it happens for overridden methods:

image

This PR also improves a few other Xtend tests checking also the position of the error marker. The relevant change in Xtend is this one: https://github.com/eclipse/xtext/compare/main...LorenzoBettini:lb_xtend_validation_tests?expand=1#diff-7eccd6f746520f0ceab7a1355467b7ca7009462eec0d4efa65b4e23749e40f95R886


szarnekow (approved):
Looks good to me.

PR #2913, opened 2024-01-18
Xtend: put an error marker on each incompatible exception in overridden method

### LINKED ISSUE #2912: [Xtend] misplaced error marker for incompatible exception throws
When a method overrides another method by declaring in incompatible exception throws declaration, Xtend puts a mark on the first exception of the throws clause. This is misleading, e.g., when there's more than one incompatible exception, and wrong, e.g., when the marked exception is NOT incompatible.

For example, take this Java class

```java
package test;

public class ExceptionThrowing {

	public void nullPointerException() throws NullPointerException {
	}
}
```

This is the error marker: you see that it's wrong, since NullPointerException is perfectly fine; in fact, the error description reports the effective incompatible exceptions:

20240118_10h02m54s_grim

I would suggest a different strategy:
- create a single error for each incompatible exception, like JDT does
- put the marker on the effectively incompatible exception

@szarnekow what do you think?

LorenzoBettini:
Closes #2912 

The PR also covers the suffix of the error message in case there are several supertypes (this was not covered before).

The markers now look like this:

image

and the two error messages say

```
The declared exception IOException is not compatible with throws clause in ExceptionThrowing.nullPointerException(), ExceptionThrowingInterface.nullPointerException() and ExceptionThrowingInterface2.nullPointerException()

The declared exception FileNotFoundException is not compatible with throws clause in ExceptionThrowing.nullPointerException(), ExceptionThrowingInterface.nullPointerException() and ExceptionThrowingInterface2.nullPointerException()
```

### DIFF
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/OverrideValidationTest.java
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/OverrideValidationTest.java
@@ -514,15 +514,18 @@ public class OverrideValidationTest extends AbstractXtendTestCase {
 	}
 
 	/**
-	 * Two incompatible exceptions; the marker is on the first one
+	 * Three incompatible exceptions from different supertypes; the marker is on the first one
 	 */
 	@Test public void testIncompatibleThrowsClause_06() throws Exception {
-		var source = "class Foo extends test.ExceptionThrowing { override nullPointerException() throws java.io.IOException, java.io.FileNotFoundException {} }";
+		var source = "class Foo extends test.ExceptionThrowing implements test.ExceptionThrowingInterface, test.ExceptionThrowingInterface2 {"
+				+ "override nullPointerException() throws NullPointerException, java.io.IOException, java.io.FileNotFoundException {} "
+				+ "}";
 		XtendClass xtendClass = clazz(source);
 		helper.assertError(xtendClass.getMembers().get(0), XTEND_FUNCTION, INCOMPATIBLE_THROWS_CLAUSE,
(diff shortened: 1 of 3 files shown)


org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/OverrideValidationTest.java
```diff
@@ -514,15 +514,24 @@ public void testClassMustBeAbstract_06() throws Exception {
...
+	 * Three incompatible exceptions from different supertypes;
+	 * the marker is set on the offending exception
 	 */
 	@Test public void testIncompatibleThrowsClause_06() throws Exception {
-		var source = "class Foo extends test.ExceptionThrowing { override nullPointerException() throws java.io.IOException, java.io.FileNotFoundException {} }";
+		var source = "class Foo extends test.ExceptionThrowing implements test.ExceptionThrowingInterface, test.ExceptionThrowingInterface2 {"
+				+ "override nullPointerException() throws NullPointerException, java.io.IOException, java.io.FileNotFoundException {} "
+				+ "}";
 		XtendClass xtendClass = clazz(source);
+		var expectedSuffix = " is not compatible with throws clause in "
```
szarnekow, line 525: Maybe and a tiny `the` like this: `is not compatible with the throws clause in`

szarnekow (approved):

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/validation/OverrideValidationTest.java
LorenzoBettini: OK, done!

PR #2915, opened 2024-01-19
Avoid busy and possibly infinite loop in JavaEditorExtension

### LINKED ISSUE #2895: Timeout on Tycho 4/5 builds
Tycho 4/5 builds regularly get killed due to hanging tests it seems
needs to be investigated

cc @LorenzoBettini @szarnekow 

https://ci.eclipse.org/xtext/job/xtext/job/cd_tycho40/
https://ci.eclipse.org/xtext/job/xtext/job/cd_tycho50/

LorenzoBettini:
Maybe due to a bug in Platform not correctly closing or saving editors.

This avoids the busy waiting loop and getting into an infinite loop.
The idea is to sleep a bit, giving the listener some time to create the event. Moreover, it gives up after a few retries.

The test becomes flaky when the but above appears, but at least it doesn't exhaust the CI CPU and does not get stuck.

Locally, I haven't seen significant performance changes during the execution.

Closes #2895

### DIFF
--- a/org.eclipse.xtend.ide.tests/src/org/eclipse/xtend/ide/tests/builder/JavaEditorExtension.xtend
+++ b/org.eclipse.xtend.ide.tests/src/org/eclipse/xtend/ide/tests/builder/JavaEditorExtension.xtend
@@ -99,13 +99,25 @@ class JavaEditorExtension {
 					}
 				}
 			], eventMask)
+
 		producer.apply
-		while (!changed.get) {
-			if (Display.getCurrent() !== null) {
-				while (Display.getDefault().readAndDispatch()) {
-					// process queued ui events
+
+		var display = Display.getCurrent()
+		assertNotNull(display)
+
+		val stop = new AtomicBoolean()
+		display.timerExec(10000, [stop.set(true)])
+
+		while (!changed.get && !stop.get) {
+			if (!display.readAndDispatch()) {
+				if (VERBOSE) {
+					println("Display sleep...")
 				}
+				display.sleep();
 			}
+		}
+		if (!changed.get) {
+			throw new AssertionError("No event has been received")
 		}
 		if (VERBOSE) {
 			println('''end waiting for an element changed event: «eventMask»''')
--- a/org.eclipse.xtend.ide.tests/xtend-gen/org/eclipse/xtend/ide/tests/builder/JavaEditorExtension.java
+++ b/org.eclipse.xtend.ide.tests/xtend-gen/org/eclipse/xtend/ide/tests/builder/JavaEditorExtension.java
@@ -134,13 +134,27 @@ public class JavaEditorExtension {
       };
       JavaCore.addElementChangedListener(_function, eventMask);
       producer.apply();
-      while ((!changed.get())) {
-        Display _current = Display.getCurrent();
-        boolean _tripleNotEquals = (_current != null);
-        if (_tripleNotEquals) {
-          while (Display.getDefault().readAndDispatch()) {
+      Display display = Display.getCurrent();
+      Assert.assertNotNull(display);
+      final AtomicBoolean stop = new AtomicBoolean();
+      final Runnable _function_1 = () -> {
+        stop.set(true);
+      };
+      display.timerExec(10000, _function_1);
+      while (((!changed.get()) && (!stop.get()))) {
+        boolean _readAndDispatch = display.readAndDispatch();
+        boolean _not = (!_readAndDispatch);
+        if (_not) {
+          if ((JavaEditorExtension.VERBOSE).booleanValue()) {
+            InputOutput.<String>println("Display sleep...");
           }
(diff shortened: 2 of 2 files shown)


org.eclipse.xtend.ide.tests/src/org/eclipse/xtend/ide/tests/builder/JavaEditorExtension.xtend
```diff
@@ -100,13 +100,22 @@ class JavaEditorExtension {
 				}
 			], eventMask)
 		producer.apply
-		while (!changed.get) {
+		var retries = 30
```
szarnekow, line 103: Can we refactor to remove the busy wait entirely?

```
Display display = Display.getCurrent();
assertNotNull(display);

AtomicBoolean stop = new AtomicBoolean();
display.timerExec(1000, ()->stop.set(true));

while (!changed.get && !stop.get && !display.readAndDispatch()) {
  display.sleep();
}
if (!changed.get) {
  throw new AssertionError(..)
}
```

LorenzoBettini: @szarnekow my proposal was also meant to remove the busy-wait since it sleeps a bit before retrying.

If I understand your proposal, that also removes the busy wait, but it only gives one second, right? Maybe that's too few?

However, yours is cleaner for sure :)

szarnekow: Yeah, we can give it 10 seconds or similar. In most cases it will be faster, I guess.

LorenzoBettini: @szarnekow you meant something like that right?

```xtend
		var display = Display.getCurrent()
		assertNotNull(display)

		val stop = new AtomicBoolean()
		display.timerExec(10000, [stop.set(true)])

		while (!changed.get && !stop.get && display.readAndDispatch()) {
			if (VERBOSE) {
				println("Display sleep...")
			}
			display.sleep()
		}
		if (!changed.get) {
			throw new AssertionError("No event has been received")
		}
```

Note that it is `display.readAndDispatch()` in the while condition, not as you said `!display.readAndDispatch()` otherwise the loop exits immediately and the test fails.

However, with this solution, I see tons of `Display sleep...` so it looks like it's busy waiting again...

szarnekow: Maybe:
```
		while (!changed.get && !stop.get) {
			if (!display.readAndDispatch()) {
				display.sleep();
			}
		}
```

LorenzoBettini: @szarnekow there are less "sleeping" messages, but still many in fast sequence.

The thread executing the code is not sleeping at all, I seem to understand

szarnekow: The thread executing the code is the display thread. That thread should not sleep since it still must process the incoming events.
It's the regular SWT event loop code that we use here.

szarnekow: Does it still turn on the fan of you machine when you run this test?

LorenzoBettini: @szarnekow I would say less than with the current code in `main` ;)

You're right about the Display thread should not sleep. My feeling was that by making it sleep the other threads could have more chances to do their job, but if these other threads are all related to the UI thread, then my proposal would make things worse.

I'll push force your proposal in my branch with verbose TRUE just to see in the CI how many messages it prints. Then I can remove verbosity and push force again so that we can merge if you're OK.

LorenzoBettini:
@szarnekow the number of "display sleep..." is big but I guess it's acceptable, see https://pipelinesghubeus7.actions.githubusercontent.com/xwd7D2dTH7KjkIQFqrWM6jPOzIKPN7cIWrsK5HktZyQ1MMY4jS/_apis/pipelines/1/runs/1433/signedlogcontent/5?urlExpires=2024-01-31T14%3A11%3A22.5089260Z&urlSigningMethod=HMACV1&urlSignature=krrosF%2FN7vVlfpD929Kelq42L4qaLjL2uqABeoORdS0%3D

I'm removing the verbose flag and push again, then, if you want, we can merge.

szarnekow (approved):

PR #2916, opened 2024-01-19
Allow customizing the NodeModelUtils and the way to obtain an INode

szarnekow:
Node lookup can be customized by implementing INodeReference on the EMF model. Node storage by means of specializing the NodeModelBuilder. Common operations on nodes by provding custom implementations of INode and overriding INode.utils().

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
@@ -24,7 +24,6 @@ import com.google.common.collect.ObjectArrays;
  * so it makes sense to remove the duplicate arrays.  This {@link GrammarElementsInterner} tries to reuse an interned version of such an array
  * wherever possible
  */
-
 public class GrammarElementsInterner {
 
 	private static class InternKey {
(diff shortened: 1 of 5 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
```diff
@@ -82,7 +81,7 @@ public boolean equals(Object o) {
 	 * this code will only function efficiently if repeated invocation uses the same grammar element
 	 * objects and arrays or else a new concatenated array will be created every time.
 	 */
-	EObject[] prependAndIntern(EObject grammarElement, Object grammarElements) {
+	protected EObject[] prependAndIntern(EObject grammarElement, Object grammarElements) {
```
LorenzoBettini, line 84: This requires a since 2.34 annotation, doesn't it?

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/impl/GrammarElementsInterner.java
```diff
@@ -93,7 +92,7 @@ EObject[] prependAndIntern(EObject grammarElement, Object grammarElements) {
 	 * this code will only function efficiently if repeated invocation uses the same grammar element
 	 * objects and arrays or else a new concatenated array will be created every time.
 	 */
-	EObject[] appendAndIntern(Object grammarElements, EObject grammarElement) {
+	protected EObject[] appendAndIntern(Object grammarElements, EObject grammarElement) {
```
LorenzoBettini, line 95: as above

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/util/NodeModelUtils.java
```diff
@@ -51,7 +53,7 @@
  * @author Sebastian Zarnekow - Initial contribution and API
  */
 public class NodeModelUtils extends InternalNodeModelUtils {
-
+
```
LorenzoBettini, line 56: nip: maybe this empty tab shouldn't be there.

LorenzoBettini (approved):
It looks good, there are only a few minor comments.
I'm not sure I understand the use-case. Maybe a test case?
Do you think it is worthwhile to document this new feature?

szarnekow:
I’ll provide follow up changes that use the new API for Xtext.xtext as an example

PR #2923, opened 2024-01-31
Use Maven wrapper consistently and the Maven enforcer plugin

### LINKED ISSUE #2922: Use Maven wrapper also in shell scripts, i.e., also in Jenkins
see https://github.com/eclipse/xtext/issues/2920#issuecomment-1918766178

LorenzoBettini:
Closes #2922

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,7 +25,6 @@ pipeline {
   }
 
   tools {
-     maven "apache-maven-3.8.6"
      jdk "temurin-jdk17-latest"
   }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,7 @@ pipeline {
   }
 
   tools {
+     maven "apache-maven-3.8.6"
      jdk "temurin-jdk17-latest"
   }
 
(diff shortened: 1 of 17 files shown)


LorenzoBettini:
IMPORTANT: I tried to get rid of the Maven specification in the Jenkinsfile (e.g., remove `maven "apache-maven-3.8.6"`, but the tests for our Maven plugins fail in Jenkins:

```
16:48:04  [ERROR] org.eclipse.xtext.maven.plugin.XtextGeneratorIT  Time elapsed: 0.269 s  <<< FAILURE!
16:48:04  java.lang.AssertionError: 
16:48:04  Maven home not found. Tried to call 'mvn'.
16:48:04  Consider to set the envVar 'M2_HOME' or System property 'maven.home'.
16:48:04  Current settings are: maven.home=null M2_HOME=null
```

I forgot about this known problem with the Maven wrapper when testing Maven plugins, which need the executable to be found in the system path. Referring to the path where the Maven wrapper installs Maven is cumbersome.

So I guess we must live with that. At least, I changed it to `maven "apache-maven-3.9.2"` in Jenkinsfiles to be consistent with the version of our Maven wrapper.

cdietrich (approved):

PR #2924, opened 2024-01-31
Add adapters to JvmTypeReference to express the expected semantics of supertypes

### LINKED ISSUE #2880: Xbase proposal: Add features to express the expected semantics of supertypes

LorenzoBettini:
Closes #2880 

This is an alternative to https://github.com/eclipse/xtext/pull/2899 that doesn't touch the Jvm types model: it introduces a new utility class to mark a JvmTypeReference as an expected class or interface, using adapters.

Currently, this information is never used, but it will be useful for https://github.com/eclipse/xtext/issues/2349

I also applied that to XtendJvmModelInferrer. I'm not too happy with the introduced API, which requires saving the cloned type reference in a local variable, use the new utility class, and then use the type reference.

To make the API better I can think of two possible solutions:
- make the setters in `JvmTypeReferenceUtil` "fluent" to return the reference
- add other methods to the JvmTypesBuilder of the shape `cloneWithProxiesExpectedAsClass`, `cloneWithProxiesExpectedAsInterface`


org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/XtendJvmModelInferrer.java
```diff
@@ -391,10 +392,14 @@ protected void initialize(XtendClass source, JvmGenericType inferredJvmType) {
 			if (typeRefToObject != null)
 				inferredJvmType.getSuperTypes().add(typeRefToObject);
 		} else {
-			inferredJvmType.getSuperTypes().add(jvmTypesBuilder.cloneWithProxies(extendsClause));
+			JvmTypeReference superClass = jvmTypesBuilder.cloneWithProxies(extendsClause);
```
szarnekow, line 395: API wise:
What do you think about 
```
jvmTypesBuilder.addImplementedInterface(inferredJvmType, superInterface);
jvmTypesBuilder.setSuperClass(inferredJvmType, superClass);
```
Internally that would do `cloneWithProxies` and add the adapter.

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/JvmTypeReferenceUtil.java
```diff
@@ -0,0 +1,44 @@
...
+	private JvmTypeReferenceUtil() {
+		// only static methods
+	}
+
+	public static void setExpectedAsClass(JvmTypeReference typeReference) {
+		typeReference.eAdapters().add(new ExpectedClassMarker());
+	}
+
+	public static boolean isExpectedAsClass(JvmTypeReference typeReference) {
+		return typeReference.eAdapters().stream().anyMatch(ExpectedClassMarker.class::isInstance);
```
szarnekow, line 34: Preferably let the adapter implement `isAdapterForType` and use `EcoreUtils.getAdapter(..)` It avoids some allocations and is more in line with the "usual" EMF way of doing this.

org.eclipse.xtend.core/src/org/eclipse/xtend/core/jvmmodel/XtendJvmModelInferrer.java
LorenzoBettini: I like it! :)
I would call the first one `addSuperInterface` though since for an interface `implemented` is misleading: in that case it should be `extended`. With `SuperIterface` we could cover both cases, right?

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/JvmTypeReferenceUtil.java
LorenzoBettini: Right! I took inspiration from `DisplayHelper`, maybe that should be changed as well.
I think we could also use singletons for the two adapters.

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/JvmTypeReferenceUtil.java
```diff
@@ -0,0 +1,56 @@
...
+public class JvmTypeReferenceUtil {
+
+	private static class ExpectedClassMarker extends AdapterImpl {
+		@Override
+		public boolean isAdapterForType(Object type) {
+			return type == ExpectedClassMarker.class;
+		}
+	}
+
+	private static class ExpectedInterfaceMarker extends AdapterImpl {
```
szarnekow, line 27: AdapterImpl has a field `Notifier target` that is maintained. I don't think we need that and it should be fine use an implementation similar to org.eclipse.xtext.xbase.validation.ReadAndWriteTracking.READMARKER

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/JvmTypeReferenceUtil.java
```diff
@@ -0,0 +1,56 @@
...
+package org.eclipse.xtext.common.types.util;
+
+import org.eclipse.emf.common.notify.impl.AdapterImpl;
+import org.eclipse.emf.ecore.util.EcoreUtil;
+import org.eclipse.xtext.common.types.JvmTypeReference;
+
+/**
+ * @author Lorenzo Bettini - Initial contribution and API
+ */
+public class JvmTypeReferenceUtil {
```
szarnekow, line 18: I don't really like `JvmTypeReferenceUtil` next to `TypeReferences`. Would you think these two classes could be merged (only keeping TypeReferences)?

szarnekow (approved):

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/JvmTypeReferenceUtil.java
LorenzoBettini: OK, so simply check the singleton is there.
I'll do that

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/JvmTypeReferenceUtil.java
LorenzoBettini: I can put the methods in `TypeReferences`, as instance methods in that case.
I haven't found a `TypeReferencesTest` so I'll create that and move also the tests of `JvmTypeReferenceUtilTest` there

LorenzoBettini: > I don't really like `JvmTypeReferenceUtil` next to `TypeReferences`. Would you think these two classes could be merged (only keeping TypeReferences)?

@szarnekow I checked `TypeReferences`, which is based on injection. The new util class doesn't need injection, that's why I created static methods. Does adding the static methods as instance methods in `TypeReferences` still make sense? Or as instance methods? For testing, I seem to understand the common.types.tests don't use injection. I could still create an instance of `TypeReferences` with `new` to test the new methods since I wouldn't use other methods from that class. Just to know whether it still looks fine.

szarnekow: You're right. I missed that TypeReferences has quite a few dependencies. The approach based on its own utility class JvmTypeReferenceUtil might be the better option indeed.

org.eclipse.xtext.common.types/src/org/eclipse/xtext/common/types/util/JvmTypeReferenceUtil.java
LorenzoBettini: I now use the same strategy of `org.eclipse.xtext.xbase.validation.ReadAndWriteTracking.READMARKER` in https://github.com/eclipse/xtext/pull/2924/commits/4b407b41bfab5d77935627d95dbb383aefdba417

org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel/src/org/eclipse/xtext/example/domainmodel/jvmmodel/DomainmodelJvmModelInferrer.xtend
```diff
@@ -32,8 +32,7 @@ class DomainmodelJvmModelInferrer extends AbstractModelInferrer {
 	def dispatch infer(Entity entity, extension IJvmDeclaredTypeAcceptor acceptor, boolean prelinkingPhase) {
 		accept(entity.toClass( entity.fullyQualifiedName )) [
 			documentation = entity.documentation
-			if (entity.superType !== null)
-				superTypes += entity.superType.cloneWithProxies
+			superClass = entity.superType
```
LorenzoBettini, line 35: @szarnekow just a small note: by using Xbase syntactic sugar, the use of the new API looks like above.
Is it better to be more explicit in this example and call `setSuperClass(entity.superType)`? We'd still use the implicit `it` anyway.

szarnekow: I like it.

LorenzoBettini: I'm a fan as well, I just wanted to make sure :)

OK, so when the build is green I'll merge

PR #2927, opened 2024-02-03
Add more flexibility to the way we find the ICompositeNode for an object

szarnekow:


org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/INodeReference.java
```diff
@@ -8,17 +8,24 @@
...
 import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
 
 /**
  * An EMF object might store the node model information differently from the default adapter
  * based approach. If it implements {@link INodeReference}, it will be queried by 
- * {@link NodeModelUtils#getNode(org.eclipse.emf.ecore.EObject)}.   
+ * {@link NodeModelUtils#getNode(org.eclipse.emf.ecore.EObject)}.
  * 
- * @since 2.34
+ * Alternatively, the {@link EObject} might contain an adapter that is for the {@link ICompositeNode ICompositeNode.class}
```
LorenzoBettini, line 20: nip: maybe "that is" is redundant?

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/util/NodeModelUtils.java
```diff
@@ -111,7 +112,14 @@ public static ICompositeNode getNode(/* @Nullable */ EObject object) {
...
 			return ((INodeReference) object).getNode();
 		}
-		return (ICompositeNode) EcoreUtil.getExistingAdapter(object, ICompositeNode.class);
+		Adapter adapter = EcoreUtil.getExistingAdapter(object, ICompositeNode.class);
+		if (adapter instanceof ICompositeNode) {
+			return (ICompositeNode) adapter;
+		}
+		if (adapter instanceof INodeReference) {
+			return ((INodeReference) adapter).getNode();
+		}
```
LorenzoBettini, line 121: Maybe I don't understand the details, but wouldn't this be equivalent to

```java
Adapter adapter = EcoreUtil.getExistingAdapter(object, ICompositeNode.class);
if (adapter instanceof INodeReference) {
	return ((INodeReference) adapter).getNode();
}
return (ICompositeNode) adapter;
```

LorenzoBettini (approved):
I only left 2 small comments.

org.eclipse.xtext/src/org/eclipse/xtext/nodemodel/util/NodeModelUtils.java
szarnekow: I wanted to have the instanceof check for the ICompositeNode first since it's far more likely to be true. Doesn't matter in practice so I'll apply your suggestion.

PR #2929, opened 2024-02-05
[Xtend] check for duplicate interface specification

### LINKED ISSUE #2928: [Xtend] No check for duplicate interface specification
While working on https://github.com/eclipse/xtext/issues/2349 I discovered that Xtend does not check for duplicate interface specification:

image

This leads to a wrong Java class generation:

image

And then the Java error is reported on the Xtend file

Screenshot-20240205134759-2560x1440

But maybe we could catch it before getting to the code generation.

@szarnekow I can fix it upstream and then port it to https://github.com/eclipse/xtext/issues/2349 , what do you think?

LorenzoBettini:
Closes #2928 

The errors are now as follows (I took inspiration from JDT, which shows the error starting from the first duplicate on):

image


szarnekow (approved):
LGTM

PR #2931, opened 2024-02-06
regenerate test languages

cdietrich:


szarnekow:
Gitlab doesn't like the diff. But there are no dragons as far as I can tell from the cmd line diff. Thank you for updating the copyright.

szarnekow (approved):

PR #2939, opened 2024-02-20
bootstrap against 2.34.0.M2

cdietrich:


cdietrich:
we ususally have this just as a workaround.
to satisfy tycho. in the past we never bumped these bootstraps
but we might reconsider

LorenzoBettini:
OK!

LorenzoBettini (approved):

PR #2942, opened 2024-02-24
update to mwe 2.17.0 release.

cdietrich:


LorenzoBettini:
@cdietrich sorry for being late.

I only noted that you did not increment the number in our target platforms.
That shouldn't be a huge problem though, since the update site has changed so I hope PDE will realize that are resolve the target platform again.

cdietrich:
yes forgot. can do in the evening

PR #2944, opened 2024-02-26
bump sequence numbers

cdietrich:


LorenzoBettini (approved):
Thanks!
On a side note: with a single (or two commits), I prefer to rebase instead of creating a merge commit. Do we have some rules about that, or is it just a matter of personal taste?

PR #2949, opened 2024-02-29
issue 2945 - add "writeStorageResources" to xtext-maven-plugin

florianpirchner:
Signed-off-by: Florian Pirchner <florian.pirchner@gmail.com>

### DIFF
--- a/org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java
+++ b/org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java
@@ -26,6 +26,9 @@ import org.eclipse.xtext.builder.standalone.LanguageAccessFactory;
 import org.eclipse.xtext.builder.standalone.StandaloneBuilder;
 import org.eclipse.xtext.builder.standalone.compiler.CompilerConfiguration;
 import org.eclipse.xtext.builder.standalone.compiler.IJavaCompiler;
+import org.eclipse.xtext.resource.IResourceDescription;
+import org.eclipse.xtext.resource.persistence.IResourceStorageFacade;
+import org.eclipse.xtext.resource.persistence.StorageAwareResource;
 import org.eclipse.xtext.util.Strings;
 import org.eclipse.xtext.xbase.lib.IterableExtensions;
 
@@ -87,10 +90,10 @@ public abstract class AbstractXtextGeneratorMojo extends AbstractXtextMojo {
 	private List<ProjectMapping> projectMappings;
 
 	@Parameter(property = "xtext.generator.skip", defaultValue = "false")
-	private Boolean skip;
+	private boolean skip;
 
 	@Parameter(defaultValue = "true")
-	private Boolean failOnValidationError;
+	private boolean failOnValidationError = true;
 
 	@Parameter(property = "maven.compiler.source", defaultValue = "11")
 	private String compilerSourceLevel;
@@ -99,10 +102,10 @@ public abstract class AbstractXtextGeneratorMojo extends AbstractXtextMojo {
 	private String compilerTargetLevel;
 
 	@Parameter(defaultValue = "false")
-	private Boolean compilerSkipAnnotationProcessing;
+	private boolean compilerSkipAnnotationProcessing;
 
 	@Parameter(defaultValue = "false")
-	private Boolean compilerPreserveInformationAboutFormalParameters;
+	private boolean compilerPreserveInformationAboutFormalParameters;
 
 	/**
 	 * RegEx expression to filter class path during model files look up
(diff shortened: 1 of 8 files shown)


org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java
```diff
@@ -137,6 +137,14 @@ public String getEncoding() {
 	@Parameter(defaultValue = "false")
 	private Boolean incrementalXtextBuild = Boolean.FALSE;
 
+	/**
+	 * Will use the ResourceStorage to write the semantic model,
+	 * {@link IResourceDescription} and optional the node model also to a binary
+	 * file. For details see {@link StorageAwareResource} and {@link ResourceStorageFacade}.
+	 */
```
cdietrich, line 144: can you please add a `@since 2.35`

florianpirchner:
@cdietrich somehow the squash did not work well.

need to configure out, how i did the last time

org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java
LorenzoBettini: Maybe the Javadoc needs some adjustments if I understand the intended meaning correctly. Shouldn't it be something like

```
Whether to use the ResourceStorage (is it a link to Java type?) to write the semantics model, the resource descriptions {@link IResourceDescription}, and optionally...
```
```

**Review by LorenzoBettini — CHANGES_REQUESTED** [review]
Thanks for the contribution, which looks great.
I only doubt the Javadoc; see the detailed comment.
It would be great if also @szarnekow could have a look.

**Comment by LorenzoBettini** [comment]
> @cdietrich somehow the squash did not work well.
> 
> need to configure out, how i did the last time

@florianpirchner maybe you wanted to do a soft or mix rebase on top of the starting branch, then stage and commit, then push force?

**Comment by florianpirchner** [comment]
@LorenzoBettini Thanks a lot for your hints. Thats exactly what i wanted to do.

**Comment by LorenzoBettini** [comment]
@florianpirchner have you seen my comment about the Javadoc?

**Comment by florianpirchner** [comment]
> @florianpirchner have you seen my comment about the Javadoc?

@LorenzoBettini Yes, thanks. Will create a new commit in about 2 hours.

**Review by LorenzoBettini — APPROVED** [review]

**Comment by LorenzoBettini** [comment]
@florianpirchner I'm going to close and reopen the PR to trigger the build against the new `main` where I merged a PR re-enabling Windows build on GitHub Actions. Just to make sure your contribution works there too.

Then, if @szarnekow agrees, we can merge

File: org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java (hunk 1 above)
Comment (szarnekow) [reply org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java:149]: Nit: I see that we use `Boolean` instead of `boolean` for some properties in this class. Should be fine to use the primitive `boolean` here, as it seems.

Comment (LorenzoBettini) [reply org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java:149]: @szarnekow should this be addressed in this PR?

Comment (florianpirchner) [reply org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java:149]: @LorenzoBettini 
Should i change my Boolean to the primitiv type?

Comment (LorenzoBettini) [reply org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java:149]: @florianpirchner That's what I asked @szarnekow . If you don't mind, you could do that in this PR. Maybe also the other `Boolean`s.

Comment (florianpirchner) [reply org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java:149]: @LorenzoBettini sure. i will do

File: org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java [hunk 2]
```diff
@@ -87,10 +90,10 @@ public String getEncoding() {
 	private List<ProjectMapping> projectMappings;
 
 	@Parameter(property = "xtext.generator.skip", defaultValue = "false")
-	private Boolean skip;
+	private boolean skip;
 
 	@Parameter(defaultValue = "true")
-	private Boolean failOnValidationError;
+	private boolean failOnValidationError;
```
LorenzoBettini, line 96: I think this should be `= true` (similar to another one below), just in case.

LorenzoBettini (changes requested):

org.eclipse.xtext.maven.plugin/src/main/java/org/eclipse/xtext/maven/AbstractXtextGeneratorMojo.java
florianpirchner: I changed it too.

LorenzoBettini (approved):

szarnekow (approved):
Thank you for the contribution and the cleanup. 👍

florianpirchner:
> Thank you for the contribution and the cleanup. 👍

You are welcome

LorenzoBettini:
@florianpirchner, the build on Jenkins fails because you haven't rebased on top of the `main` branch, where I recently committed a small change to the way we test our Maven plugins; On Jenkins we use the `full-build.sh` script, and if yours branch version is not updated it will fail. Could you please rebase?

LorenzoBettini:
@florianpirchner I'll first try to close and re-open the PR to check whether Jenkins will take the new `main`.
If it still fails, then you'll have to rebase (not reset) on top of `main`.

LorenzoBettini:
@florianpirchner no, I'm afraid the trick does not convince Jenkins ;)
if you could please rebase.

PR #2950, opened 2024-03-04
Re-enable Windows GitHub Actions and use Maven Wrapper also for Maven plugins ITs

LorenzoBettini:
This contributes two main features:

- The Windows GitHub Actions workflow is re-enabled. It's slower than Linux but faster than macOS (about 1:10 instead of 1:30). Note that two tests had to be adjusted to deal with Windows eols
- By retrieving the home of the installed Maven wrapper, the same Maven wrapper installation is used to run all our Maven plugins ITs, enforcing consistency and reproducibility and allowing us to rely only on the wrapper instead of the possibly installed Maven (in Jenkins, I modified the workflows accordingly).

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -14,7 +14,7 @@ jobs:
     if: github.event_name != 'pull_request' || github.event.pull_request.base.repo.clone_url != github.event.pull_request.head.repo.clone_url
     strategy:
       matrix:
-        os: ['ubuntu-latest', 'macos-latest' ] # 'windows-latest' too flaky
+        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
       fail-fast: false
     runs-on: ${{ matrix.os }}
 
@@ -35,10 +35,18 @@ jobs:
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml', '**/*.target') }}
           restore-keys: ${{ runner.os }}-m2
 
+      - name: Set Maven Home environment variable
+        run: echo "MAVEN_WRAPPER_HOME=$(./mvnw --version | grep "Maven home:" | cut -c 13-)" >> "$GITHUB_ENV"
+        if: runner.os != 'Windows'
+      - name: Set Maven Home environment variable on Windows
+        run: echo "MAVEN_WRAPPER_HOME=$(./mvnw --version | grep "Maven home:" | cut -c 13-)" | Out-File -FilePath $env:GITHUB_ENV -Append
+        # In Windows the syntax for setting environment variable is different
+        if: runner.os == 'Windows'
+
       - name: Build and test
         uses: coactions/setup-xvfb@v1
         with: 
-          run: ./mvnw clean verify -PuseJenkinsSnapshots -f org.eclipse.xtext.full.releng
+          run: ./mvnw clean verify "-Dmaven.home=${{ env.MAVEN_WRAPPER_HOME }}" -PuseJenkinsSnapshots -f org.eclipse.xtext.full.releng
 
   build-maven-artifacts:
     if: github.event_name != 'pull_request' || github.event.pull_request.base.repo.clone_url != github.event.pull_request.head.repo.clone_url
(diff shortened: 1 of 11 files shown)


.github/workflows/maven.yml
```diff
@@ -61,5 +69,8 @@ jobs:
           key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml', '**/*.target') }}
           restore-keys: ${{ runner.os }}-maven
 
+      - name: Set Maven Home environment variable
+        run: echo "MAVEN_WRAPPER_HOME=$(./mvnw --version | grep "Maven home:" | cut -c 13-)" >> "$GITHUB_ENV"
+
       - name: Build Maven artifacts
-        run: ./mvnw clean verify -PuseJenkinsSnapshots -f org.eclipse.xtext.maven.releng
+        run: ./mvnw clean verify "-Dmaven.home=${{ env.MAVEN_WRAPPER_HOME }}" -PuseJenkinsSnapshots -f org.eclipse.xtext.maven.releng
```
szarnekow, line 76: Nice trick

org.eclipse.xtext.maven.plugin/src/test/java/org/eclipse/xtext/maven/plugin/XtextGeneratorIT.java
```diff
@@ -49,6 +49,7 @@ public class XtextGeneratorIT {
 	static public void setUpOnce() throws IOException, VerificationException {
 		testDir = extractTestRoot();
 		if (fork) {
+			System.out.println("maven.home=" + System.getProperty("maven.home"));
```
szarnekow, line 52: Do you think this should stay here?

.github/workflows/maven.yml
LorenzoBettini: It took me some time to get to that O:)

org.eclipse.xtext.maven.plugin/src/test/java/org/eclipse/xtext/maven/plugin/XtextGeneratorIT.java
LorenzoBettini: Yes, that would be intentional: it will always show the value of maven.home, even when not set. This should help debug possible problems, even when running locally. Remember that once the verifier object is created, it will swallow System.out, System.err, so it's better to know this information before creating it.

szarnekow: Maybe make it more useful and print the value of "maven.home" here and after the dance that happens in the lines below

LorenzoBettini: @szarnekow as I said, after a `Verifier` object is created, System.out and System.err are not directed to the console anymore; anything printed after the next line will be swallowed.
What I could do is save `System.out` in advance, what do you think?
And then print the result: either the executable or the result of `findMaven`.

szarnekow: > What I could do is save System.out in advance, what do you think?

Yes, I'd probably save it as a field on the class to have it available for other logging, too. WDYT?

szarnekow: (static field, that is)

LorenzoBettini: @szarnekow done here: https://github.com/eclipse/xtext/pull/2950/commits/83ecc709e46aae6a1b2e4e17ac99d85986ec6a31

szarnekow (approved):
Nice nice

LorenzoBettini:
This is what's printed in GH:

```
IT projects will be executed from /home/runner/work/xtext/xtext/org.eclipse.xtext.maven.plugin/target/local-test-dir/it/generate
Local Maven Central Repository /home/runner/.m2/repository
Current maven.home=/home/runner/.m2/wrapper/dists/apache-maven-3.9.2-bin/3238cb54/apache-maven-3.9.2
Maven executable: /home/runner/.m2/wrapper/dists/apache-maven-3.9.2-bin/3238cb54/apache-maven-3.9.2/bin/mvn
```

On my local machine, without passing anything:

```
Current maven.home=null
Detected maven.home=/usr/local/
```

I'll wait for the build to finish and then I'll merge.

PR #2951, opened 2024-03-05
Make Pattern and Splitter static fields

rubenporras:
Make Pattern and Splitter static fields so that they do not need to created every time a new DiagnosticConverter is injected.


szarnekow (approved):

LorenzoBettini:
@szarnekow shouldn't this be tagged with 2.35 instead of 2.36?

szarnekow:
> @szarnekow shouldn't this be tagged with 2.35 instead of 2.36?

Sure. I think I misclicked. Thanks for fixing @cdietrich

PR #2954, opened 2024-03-06
prepare 2024-03 release target

cdietrich:


LorenzoBettini (approved):
You haven't increased the sequence number in the latest target platform, but it's not a big problem ;)

PR #2969, opened 2024-04-02
Generator fragment for textmate grammars and Xbase LSP server infra

szarnekow:


org.eclipse.xtend.core/xtend-gen/org/eclipse/xtend/core/macro/TransformationContextImpl.java
```diff
@@ -106,8 +106,8 @@ public List<? extends Problem> getProblems(final Element element) {
     return this.getProblemSupport().getProblems(element);
   }
 
-  public void validateLater(final Procedure0 validationCallback) {
```
szarnekow, line 109: To be honest, I don't know where these changes in the generated code come from.

LorenzoBettini: @szarnekow  Just a quick note about this, because I still haven't reviewed everything: this happens to me as well when you run the maven build and use the xtend maven plugin. If you run clean build from eclipse the parameters go back to their original names. I never investigated further. Maybe it's due to some settings in this project that's only used by maven.

cdietrich: There is an open issue for this. but of course I cannot find it

szarnekow: I'm aware of that workaround, but it doesn't work anymore. Maybe tycho always runs the generator now. Not sure.

LorenzoBettini: @szarnekow I meant: if you rely on xtend from eclipse in the workspace it will keep the parameter names. If you run the maven build it will generate the parameter names with "arg".

szarnekow: Yes, I understood what you were saying. But a clean-build in Eclipse doesn't fix it for me on a freshly oomphed Eclipse.

LorenzoBettini (approved):

org.eclipse.xtend.core/xtend-gen/org/eclipse/xtend/core/macro/TransformationContextImpl.java
LorenzoBettini: Oh! That's strange.. and you have a jdk, not only a jre?

szarnekow: I started looking into this: https://github.com/eclipse/xtext/pull/2995

Needs a change in the test-infra for common.types, though.

LorenzoBettini (commented):
@szarnekow I only left a few minor comments.

Issue #2975, opened 2024-04-05
Reduce usage of Guava


HannesWell:
Guava is a dependency that's often difficult to handle (one reason is that Xtext is re-exporting it) and many of the enhancements Guava offered are part of the Java's standard library in recent versions. For the most commonly used elements in Guava there are often (almost) drop-in replacements or equivalent alternatives that are simple to migrate to. Many elements in Guava even recommend to use standard alternatives in their doc.
Therefore I want to suggest to replace Guava where it is possible.

In general the usage is split into two parts:
- Internal usages, which can be replaced by a technical suitable solution at any time.
- Usage in APIs.
For APIs alternatives would have to be added and the existing methods would need to be deprecated for some time before they can be removed? This also relates to https://github.com/eclipse/xtext/issues/2668 and should probably be handled together.

In general Guava is heavily used by Xtext, searching the `*.java` files in this repo for `com.google.common` gives 3535 hits. So replacing it is a major task and probably more a journey. At its end maybe Guava can even be dropped completely.
But since it is hard to even get just a completely overview of its usage, my suggestion is to start with simple cases like replacing `Lists.newArrayList()` by `new ArrayList<>()` etc. only in internal usages and see how far we can get.

LorenzoBettini:
In general, I'd have no problem replacing things like `Lists.newArrayList()` with `new ArrayList<>()`, and that's what I've already done and doing when changing existing code.

I have doubts (see https://github.com/eclipse/xtext/pull/2978#issuecomment-2041071680) that having overloaded methods with functional interfaces would work: the Java compiler would issue an error if you pass a lambda expression, wouldn't it?
In fact, I see explicit casts in @HannesWell 's PR when passing lambda expressions to the new method from the deprecated one. If Java issued compilation errors, that would be a huge problem ;)

If switching to Java functional interfaces instead of the Guava ones is considered, I think it would be better to simply replace the functional interfaces from Guava with the ones of Java without deprecation and overloading. That would be API breakage, but it should go completely unnoticed if clients used to call such methods with lambda expressions. I don't know if it could give problems from a binary point of view, though.

szarnekow:
> I don't know if it could give problems from a binary point of view, though.

That would indeed be source-compatible but not binary-compatible.

HannesWell:
[...]

PR #2976, opened 2024-04-05
Replace usage of Maps.newHashMapWithExpectedSize() with standard Java

HannesWell:
In Java-19 there is even static factory method `HashMap.newHashMap(int)` that can be used as a drop-in replacement once Java-19 or later is required.
For now I calculate the necessary capacity explicitly using the formula: `expected-size *4/3+1`.

The change in XtextGeneratorTemplates also removes the usage of `Maps.newHashMapWithExpectedSize()` of many generated Activators, but I haven't changed them manually. I assume they should be regenerated, shouldn't they?

### DIFF
--- a/org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreResourceDescriptionStrategy.java
+++ b/org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreResourceDescriptionStrategy.java
@@ -8,6 +8,7 @@
  *******************************************************************************/
 package org.eclipse.xtext.ecore;
 
+import java.util.Collections;
 import java.util.Map;
 
 import org.apache.log4j.Logger;
@@ -19,7 +20,6 @@ import org.eclipse.xtext.resource.IEObjectDescription;
 import org.eclipse.xtext.resource.impl.DefaultResourceDescriptionStrategy;
 import org.eclipse.xtext.util.IAcceptor;
 
-import com.google.common.collect.Maps;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 
@@ -48,8 +48,7 @@ public class EcoreResourceDescriptionStrategy extends DefaultResourceDescription
 		try {
 			QualifiedName qualifiedName = qualifiedNameProvider.getFullyQualifiedName(eObject);
 			if (qualifiedName != null) {
-				Map<String, String> userData = Maps.newHashMapWithExpectedSize(1);
-				userData.put(NS_URI_INDEX_ENTRY, Boolean.toString(isNsURI));
+				Map<String, String> userData = Collections.singletonMap(NS_URI_INDEX_ENTRY, Boolean.toString(isNsURI));
 				IEObjectDescription description = EObjectDescription.create(qualifiedName, eObject, userData);
 				acceptor.accept(description);
 				return true;
--- a/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGeneratorTemplates.xtend
+++ b/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGeneratorTemplates.xtend
@@ -8,7 +8,6 @@
  *******************************************************************************/
 package org.eclipse.xtext.xtext.generator
 
-import com.google.common.collect.Maps
 import com.google.inject.Binder
 import com.google.inject.Guice
 import com.google.inject.Inject
@@ -18,6 +17,7 @@ import com.google.inject.Provider
 import com.google.inject.Singleton
 import com.google.inject.name.Names
 import java.util.Collections
+import java.util.HashMap
 import java.util.List
 import java.util.Map
 import java.util.Properties
@@ -36,12 +36,12 @@ import org.eclipse.xtext.xtext.generator.model.FileAccessFactory
 import org.eclipse.xtext.xtext.generator.model.GeneratedJavaFileAccess
 import org.eclipse.xtext.xtext.generator.model.GuiceModuleAccess
 import org.eclipse.xtext.xtext.generator.model.JavaFileAccess
+import org.eclipse.xtext.xtext.generator.model.TextFileAccess
 import org.eclipse.xtext.xtext.generator.model.TypeReference
 import org.eclipse.xtext.xtext.generator.model.annotations.SuppressWarningsAnnotation
 import org.eclipse.xtext.xtext.generator.model.project.IXtextProjectConfig
 
 import static extension org.eclipse.xtext.xtext.generator.model.TypeReference.*
-import org.eclipse.xtext.xtext.generator.model.TextFileAccess
 
 /**
  * Templates for generating the common language infrastructure.
@@ -505,7 +505,7 @@ class XtextGeneratorTemplates {
 				
 				private static «activator.simpleName» INSTANCE;
 				
-				private «Map»<String, «Injector»> injectors = «Collections».synchronizedMap(«Maps».<String, «Injector»> newHashMapWithExpectedSize(1));
+				private «Map»<String, «Injector»> injectors = «Collections».synchronizedMap(new «HashMap»<>(2));
 				
 				@Override
 				public void start(«'org.osgi.framework.BundleContext'.typeRef» context) throws Exception {
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/clustering/CopiedResourceDescription.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/clustering/CopiedResourceDescription.java
@@ -9,6 +9,7 @@
 package org.eclipse.xtext.builder.clustering;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -26,7 +27,6 @@ import org.eclipse.xtext.resource.impl.AbstractResourceDescription;
 import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Maps;
 
 /**
  * Makes an eager copy of the exported objects of another resource description.
@@ -56,7 +56,7 @@ public class CopiedResourceDescription extends AbstractResourceDescription {
                         Map<String, String> userData = null;
                         for (final String key : from.getUserDataKeys()) {
                             if (userData == null) {
-                                userData = Maps.newHashMapWithExpectedSize(2);
+                                userData = new HashMap<>(3);
                             }
                             userData.put(key, from.getUserData(key));
                         }
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/EObjectDescriptionProvider.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/serializer/impl/EObjectDescriptionProvider.java
@@ -9,6 +9,7 @@
 package org.eclipse.xtext.ide.serializer.impl;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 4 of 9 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/preferences/OptionsConfigurationBlock.java
```diff
@@ -1,697 +1,698 @@
...
-			logError("Project specific key is not qualified", null);
-		}
-		return key;
-	}
-
-	public static String isProjectSpecificPropertyKey(String propertyPrefix) {
-		return propertyPrefix + "." + IS_PROJECT_SPECIFIC;
-	}
-}
+/*******************************************************************************
```
HannesWell, line 1: There are so many changes in this file because its line-endings are not normalized.
I just created https://github.com/eclipse/xtext/pull/2977 to fix that for the entire repo and converted this to a draft.
Once https://github.com/eclipse/xtext/pull/2977 is decided I think this should be rebased/redone in order to minimize the changes.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGeneratorTemplates.xtend
```diff
@@ -505,7 +505,7 @@ class XtextGeneratorTemplates {
 				
 				private static �activator.simpleName� INSTANCE;
 				
-				private �Map�<String, �Injector�> injectors = �Collections�.synchronizedMap(�Maps�.<String, �Injector�> newHashMapWithExpectedSize(1));
+				private �Map�<String, �Injector�> injectors = �Collections�.synchronizedMap(new �HashMap�<String, �Injector�>(2));
```
HannesWell, line 508: Is there a lower for the Java version this template targets?
For example if a ConcurrentHashMap would be used here, below the method `getInjector(String)` could be simplified to:
```
public «Injector» getInjector(String language) {
	return injectors.computeIfAbsent(language, this::createInjector);
}
```
Which would also improve the concurrency since concurrent reads to not block each other and concurrent updates of **different keys** are possible.
But this requires at least Java-1.8.

LorenzoBettini: Xtext bundles require Java 11.

HannesWell: > Xtext bundles require Java 11.

Yes, I wasn't sure if the projects whose code is generated with this template also depend on Xtext or maybe could be independent and therefore target even older java version. But I now have the impression that the former is the case.

HannesWell:
> I think @szarnekow should review this, but personally, I'm not a big fan of such strange expressions for computing the size ;)

Is it about the expression or specifying the size at all?

If it is the latter, I can also remove it. If it is the former I'm fine to restrict this PR to the change of `XtextGeneratorTemplates.xtend`, which is 'required' for https://github.com/eclipse/mwe/pull/287#issuecomment-2036155568 or at least the changes that use only constants. I mainly did the others for completeness.
The remaining parts could then be replaced as soon as Java-21 is required using the method `HashMap.newHashMap(int)` introduced in `Java-19` as mentioned in my initial message.

> another note: `Maps.of` returns an immutable map, IIRC. Could this potentially lead to breakage if the current factory method returns a mutable list?

That's right. But I have only used it in two places where the map is passed as `userData` to `EObjectDescription.create()`. And if I have followed the code-path correctly that map is only read there.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGeneratorTemplates.xtend
szarnekow: As usual, the devil is in the details. Injector creation may be re-entrant if a language tries to obtain services from another injector, e.g. from a super-language in the same bundle.

Hash collisions on strings are not super rare:

```java
public class ActivatorCodeTest {

	@Test
	public void testCollision() {
		String a = "com.acme.some.package.name.DB";
		String b = "com.acme.some.package.name.Ca";
		assertEquals(a.hashCode(), b.hashCode());
	}
	
	@Test
	public void testConcurrentMapComputeIfAbsent() {
		String a = "com.acme.some.package.name.DB";
		String b = "com.acme.some.package.name.Ca";
		
		Map<Object, Object> injectors = new ConcurrentHashMap<>();
		assertNotNull(injectors.computeIfAbsent(a, first->{
			return injectors.computeIfAbsent(b, reentrant->{
				return new Object();
			});
		}));
	}
	
	@Test
	public void testSynchronizedMapComputeIfAbsent() {
		String a = "com.acme.some.package.name.DB";
		String b = "com.acme.some.package.name.Ca";
		
		Map<Object, Object> injectors = Collections.synchronizedMap(new HashMap<>());
		assertNotNull(injectors.computeIfAbsent(a, first->{
			return injectors.computeIfAbsent(b, reentrant->{
				return new Object();
			});
		}));
	}
	
}
```

szarnekow: What I try to say: Even if the code looks like a plain `computeIfAbsent`, it's not the same.

org.eclipse.xtext/src/org/eclipse/xtext/resource/impl/EObjectDescriptionLookUp.java
```diff
@@ -115,7 +115,8 @@ protected Map<QualifiedName, List<IEObjectDescription>> getNameToObjects() {
 		if (nameToObjects == null) {
 			synchronized (this) {
 				if (nameToObjects == null) {
-					Map<QualifiedName, List<IEObjectDescription>> nameToObjects = Maps.newHashMapWithExpectedSize(allDescriptions.size());
+					Map<QualifiedName, List<IEObjectDescription>> nameToObjects = new HashMap<>(
```
szarnekow, line 118: I don't like repeating that arithmetic expression all over the place. A logical step would be to extract a factory method. Let's call it `newHashMapWithExpectedSize` and put it into a type `Maps`. Oh wait.

Sorry for being sarcastic here, but I'm a little tired of the perceived panic mode when it comes to Guava dependencies.

HannesWell:
I have now reverted the changes that require computation of the capacity. That can be replaced once the Java-19 `HashMap.newHashMap(int)` is available for Xtext.
The main change for this PR in `XtextGeneratorTemplates.xtend` is not affected by that.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/XtextGeneratorTemplates.xtend
HannesWell: > As usual, the devil is in the details. Injector creation may be re-entrant if a language tries to obtain services from another injector, e.g. from a super-language in the same bundle.

That would indeed be a problem.

org.eclipse.xtext/src/org/eclipse/xtext/resource/impl/EObjectDescriptionLookUp.java
HannesWell: > Sorry for being sarcastic here, but I'm a little tired of the perceived panic mode when it comes to Guava dependencies.

It's not panic, it's more annoyance if you can't upgrade because of conflicting version in unfortunate situations. And I'm tired of it as well, which is why I try to work on a solution.

HannesWell:
Is there anything left here to land this?

cdietrich:
i guess a rebase would be helpful.
and any cappa in the team that does not need to be spent on breaking upstream changes....

HannesWell:
> i guess a rebase would be helpful.

Rebased it and looked into the currently open issues and answered where I can think I can help something. :)

org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreResourceDescriptionStrategy.java
```diff
@@ -48,8 +47,7 @@ protected boolean createEObjectDescriptions(IQualifiedNameProvider qualifiedName
 		try {
 			QualifiedName qualifiedName = qualifiedNameProvider.getFullyQualifiedName(eObject);
 			if (qualifiedName != null) {
-				Map<String, String> userData = Maps.newHashMapWithExpectedSize(1);
-				userData.put(NS_URI_INDEX_ENTRY, Boolean.toString(isNsURI));
+				Map<String, String> userData = Map.of(NS_URI_INDEX_ENTRY, Boolean.toString(isNsURI));
```
szarnekow, line 50: I'm reluctant to accept this change, since the map will now throw an NPE when queried with null as a key. Though that's a valid implementation of the Map interface, it doesn't follow the principle of least surprise, which would just be `containsKey(null) == false` and `get(null) == null`

HannesWell: In both cases the map is passed to `org.eclipse.xtext.resource.EObjectDescription` and only queried in
https://github.com/eclipse/xtext/blob/2bce8b6c62b1da7c57d81cff6e699eedb19d90a9/org.eclipse.xtext/src/org/eclipse/xtext/resource/EObjectDescription.java#L89-L94


The doc of that method already defines that the name argument must not be null:
https://github.com/eclipse/xtext/blob/2bce8b6c62b1da7c57d81cff6e699eedb19d90a9/org.eclipse.xtext/src/org/eclipse/xtext/resource/IEObjectDescription.java#L49-L54

It was indeed not enforced before, but the new behavior would be 100% compliant with the specification of the method.

LorenzoBettini:
@HannesWell two things I don't understand about test results:

- there's no way to see the actual results? I can't seem to be able to click anything on the results 
- if it works only for PR, then it won't work for committers, since we skip the building for our PRs and only build for pushes

LorenzoBettini:
Of course, for the second item, we could change our design and also build PRs for committers, cc @cdietrich

org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreResourceDescriptionStrategy.java
szarnekow: Understandable. Yet I prefer some simple guidelines in a code base, one of which would be to avoid the overly strict immutable collections. I've seen to many NPE when doing containsKey, indexOf or contains checks on these with a null in code paths that used to work great until they were refactored to the immutable collections.

HannesWell: Personally I think it is good that these immutable collections are strict with null elements since one is forced to be more careful in using null, which on the long run and in general is an improvement for a code base, even though it can be annoying in the beginning.

But you prefer not to use it here so I replaced these two usages of Map.of() with `HashMap`.

szarnekow: Curious: Why didn't you use Collections.singletonMap instead?
(No need to change this)

szarnekow:
@LorenzoBettini Does your comment on the other MR mean that the check result `Some checks were not successful` cannot be trusted?

LorenzoBettini:
> @LorenzoBettini Does your comment on the other MR mean that the check result `Some checks were not successful` cannot be trusted?

The "Build / Test Results" failing cannot be trusted: it fails even in the presence of flakes. We can only trust "Build / build..." jobs failing.

Unfortunately, the test results action cannot handle flaky tests currently.

org.eclipse.xtext.ecore/src/org/eclipse/xtext/ecore/EcoreResourceDescriptionStrategy.java
HannesWell: > Curious: Why didn't you use Collections.singletonMap instead?

I'm not sure anymore. I think I just forgot it during the discussion and initially I probably assumed that `Map.of()` is more elegant while meeting the spec. of the methods.

> (No need to change this)

Even if there is no need, I think that's the better solution and change it accordingly.

HannesWell:
Can this be considered for Xtext M3?

szarnekow (approved):
I'd be ok with merging für M3

PR #2979, opened 2024-04-05
Replace Guava-using APIs in ContainersStateFactory and its dependencies

HannesWell:
Also replace Guava where it is simply possible.


org.eclipse.xtext/src/org/eclipse/xtext/mwe/PathTraverser.java
```diff
@@ -52,22 +67,18 @@ public Set<URI> findAllResourceUris(String path, Predicate<URI> isValidPredicate
 		throw new IllegalArgumentException("Unsupported path : " + path + " (only folders and archives are supported).");
 	}
 
+	/** @deprecated Instead use {@link #traverseArchive(File, Predicate)} */
+	@Deprecated(since = "2.35.0", forRemoval = true)
+	protected Set<URI> traverseArchive(File file, com.google.common.base.Predicate<URI> isValidPredicate) {
```
HannesWell, line 72: I wonder if it is necessary to keep these protected methods as deprecated?
Are custom sub-classes of `PathTraverser` something that should be supported?
If yes, it should probably also be ensured that these deprecated methods are still called in order to not make custom implementation dysfunctional.

The same applies for `Reader` and `RuntimeResourceSetInitializer` adjusted in this PR.

LorenzoBettini:
Even in this case, I have the same perplexity I expressed in https://github.com/eclipse/xtext/pull/2978

My suspects might be proved by the fact that you have to add an explicit cast when you call the new method from the deprecated one when passing a lambda expression.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/ResourceSetBasedAllContainersState.java
```diff
@@ -9,36 +9,46 @@
...
  * @see FlatResourceSetBasedAllContainersState
  * 
  * @author Sven Efftinge - Initial contribution and API
  */
 public class ResourceSetBasedAllContainersState implements IAllContainersState {
-	
-	private SetMultimap<String, URI> container2URIs;
-	private SetMultimap<URI, String> uri2container;
+
+	private Map<String, Set<URI>> container2URIs;
```
szarnekow, line 32: Previous usages of container2URIs would never return null when queried via `get(someHandle)`. getContainedURIs might need refinedment.
Into the bargain, it used to be possible to add items to the collection returned from getContainedURis and the changed would be reflected in the containers state, regardless of the fact whether the containers was initially known or not.
We don't (ab)use that detail in the framework, but I'm not a fan of the subtle change nor of the effort we'd need to put into retained that behaviro without using Guava Multimaps.

HannesWell: > Previous usages of container2URIs would never return null when queried via `get(someHandle)`. getContainedURIs might need refinedment.

I adjusted `getContainedURIs` to reflect that, i.e. to `return container2URIs.getOrDefault(containerHandle, Set.of());`.

As a side-note, for `uri2container.get(uri)` the result was still checked for null.

> Into the bargain, it used to be possible to add items to the collection returned from getContainedURis and the changed would be reflected in the containers state, regardless of the fact whether the containers was initially known or not. We don't (ab)use that detail in the framework, but I'm not a fan of the subtle change nor of the effort we'd need to put into retained that behaviro without using Guava Multimaps.

I would call this an undesired implementation detail on which callers shouldn't rely on. Furthermore the documentation of this class states 
https://github.com/eclipse/xtext/blob/9e3b1b496e157a58fe3bf556bb910f03c3a7a9a7/org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/ResourceSetBasedAllContainersState.java#L25-L26

According to this contract it would not be allowed to add elements. And as a third point, adding elements to a value collection of `container2URIs`, would mean that it goes out of sync with `uri2container`.

PR #2983, opened 2024-04-08
Added IterableExtensions.lastOrNull and deprecated last

### LINKED ISSUE #2981: Problems with the new `getLast` method in Java 21 and the extension method `last`
Java 21 introduces the `getLast` method in a few collection classes.
Xbase will prefer `getLast` to our extension method `last` to generate Java code when the runtime is Java 21 when the `last` extension method is used in the DSL code, e.g., Xtend.
Such a strategy has been implemented in the past for example to use the Java 8 `forEach` instead of the extension method `forEach`. And that never generated problems.

The problem now is that our extension method `last` returns null when the collection is empty, while the `getLast` method in Java 21 throws a 

```
java.util.NoSuchElementException
 	at java.base/java.util.List.getLast(List.java:843)
``` 

https://github.com/openjdk/jdk/blob/d771ec6296bd3bb17ac2dd79671fdb6f154b4ea5/src/java.base/share/classes/java/util/List.java#L839

I'm already changing a few occurrences in our code base, mainly in XbaseFormatter and XtendFormatter while working on https://github.com/eclipse/xtext/issues/2686: instead of doing `list.last` I'm changing it into `IterableExtensions.last(list)`, thus forcing the use of our method.

but I think we should take counter measures in the Xbase compiler now to avoid generating code with different and breaking semantics when the runtime is Java 21.

@szarnekow do you think we could add a kind of exception rule when resolving methods in Xbase. I mean, when the method `last` is an extension method we should not look for possibly matching methods from the Java API, and force the use of our `last` method?

LorenzoBettini:
Closes #2981

I have updated the occurrences in our code base, but NOT in the input files of our tests.

### DIFF
--- a/org.eclipse.xtext.xbase.lib.tests/src/org/eclipse/xtext/xbase/tests/lib/IterableExtensionsTest.java
+++ b/org.eclipse.xtext.xbase.lib.tests/src/org/eclipse/xtext/xbase/tests/lib/IterableExtensionsTest.java
@@ -86,7 +86,7 @@ public class IterableExtensionsTest extends BaseIterablesIteratorsTest<Iterable<
 	}
 
 	@Override
-	protected Integer last(Iterable<Integer> input) {
+	protected Integer lastOrNull(Iterable<Integer> input) {
 		return IterableExtensions.lastOrNull(input);
 	}
 
--- a/org.eclipse.xtext.xbase.lib.tests/src/org/eclipse/xtext/xbase/tests/lib/IterableExtensionsTest.java
+++ b/org.eclipse.xtext.xbase.lib.tests/src/org/eclipse/xtext/xbase/tests/lib/IterableExtensionsTest.java
@@ -87,7 +87,7 @@ public class IterableExtensionsTest extends BaseIterablesIteratorsTest<Iterable<
 
 	@Override
 	protected Integer last(Iterable<Integer> input) {
-		return IterableExtensions.last(input);
+		return IterableExtensions.lastOrNull(input);
 	}
 
 	@Override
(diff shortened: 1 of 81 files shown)


org.eclipse.xtext.xbase.lib.tests/src/org/eclipse/xtext/xbase/tests/lib/IterableExtensionsTest.java
```diff
@@ -87,7 +87,7 @@ protected Integer findLast(Iterable<Integer> input, Function1<Integer, Boolean>
 
 	@Override
 	protected Integer last(Iterable<Integer> input) {
```
szarnekow, line 89: Can you please rename this method `last` also to `lastOrNull`?

LorenzoBettini: @szarnekow this method is defined as abstract in the superclass `org.eclipse.xtext.xbase.tests.lib.BaseIterablesIteratorsTest.last(IterableOrIterator)` so I should rename it also in `IteratorExtensionsTest`.
I changed to `lastOrNull` only in `IterableExtensions`, NOT in `IteratorExtensions`.

szarnekow: Yes, I saw that. I wonder if we should both keep in symmetry, though. Thoughts?

LorenzoBettini: I also thought about that, but how do we "justify" the deprecation for `IteratorExtensions`? Of course, we could use the "symmetry" justification.

Also `head` could be changed accordingly.

Unfortunately, we are forced to change `IterableExtensions.last` only for the clash with Java 21 new API.

szarnekow: So far I liked that the extension API for iterables and iterators was equivalent (as far as it goes).

LorenzoBettini: Ok, then I'll add lastOrNull to IteratorExtensions and update this PR

LorenzoBettini:
@szarnekow I've added `lastOrNull` to `IteratorExtensions` as well, and updated also the tests according to https://github.com/eclipse/xtext/pull/2983#discussion_r1555733997

when the build is green, I'll merge

Issue #2686, opened 2023-05-25
Java 21 Masterplan


cdietrich:
Once Java 21 LTS is out we should try to support it

- [ ] set minimum supported version to Java 17, Drop Java 11 (build)
- [ ] adapt JavaVersion Enum
- [ ] adapt Wizard tests
- [ ] adapt platform dependencies in dev-bom
- [ ] adapt minimal target platform
- [ ] adapt gradle version in wizard
- [ ] check if classgraph needs adaptation?
- [ ] check if tycho needs adaptation?
- [ ] Guice/Guava update needed?
- [ ] new MWE?
- [ ] check java version switches in tests
- [ ] new Jetty version needed?
- [ ] with new minimal target version, check which reflective switches we can get rid of (platform, jdt)
- [ ] deal with Java deprecations

TODO: 100 things i have forgot

cdietrich:
Guice and Guava is now on latest. we will see if this is sufficient

OLibutzki:
As Java 21 has been released today, do you have any roadmap when to support it? Especially solving the jdt issue mentioned in #2680 is something we are looking forward to.

cdietrich:
There is no road map and as Xtext is no longer my day job it might take long. First question: are Java 21 capable eclipse platform and jdt maven artifacts already available or will they ship in December ?

OLibutzki:
Hmm, this issue us "open", no new comments for months.

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/890

cdietrich:
https://github.com/eclipse-jdt/eclipse.jdt.core/commits/BETA_JAVA21 Shows some wip

mpalat:
Please see : https://www.eclipse.org/lists/eclipse-dev/msg12203.html

cdietrich:
with https://www.eclipse.org/lists/eclipse-dev/msg12207.html we wont have maven artifacts before end of year.
am not sure if there are maven snapshots meanwhile so development can start before

DanySK:
Hi, is there any update on this plan?

I guess that errors like `org.eclipse.xtext.common.types.access.TypeResource$CannotLoadTypeResourceException: java.lang.IllegalArgumentException: Unsupported class file major version 65` are due to Xtext not yet supporting Java 21 - am I correct?

[...]

PR #2986, opened 2024-04-09
Initial preparation for Java 21

LorenzoBettini:
This is part of https://github.com/eclipse/xtext/issues/2686

It adds another GitHub Actions workflow for building with JDK21 (only in Linux to avoid matrix explosion), making sure that everything builds and tests are green with Java 21 and Java 17.

Many tests had to be adapted:
- either by adding stub classes to avoid tests being fragile w.r.t. new Java API, both concerning new added methods (e.g., getLast) and changes in supertype hierarchies (added SequencedCollection: List extends SequencedCollection in Java 21)
- or by adapting tests according to the Java version (it's already been done in the past); e.g., the new private method in Object `wait0`.
- A particular case: `strictfp` doesn't seem to be present in the byte code anymore when compiling with JDK 21 so the corresponding tests are skipped with Java 21. By the way, a few tests where `strictfp` was tested contained a copy-and-paste error and `publicFinalMethod` was never tested (this has been fixed)
- `lastOrNull` (https://github.com/eclipse/xtext/pull/2983) is used only when strictly needed in test inputs.
- `maven-plugin-plugin` has been updated or it couldn't build plugin descriptor for Java 21

IMPORTANT: this does NOT make the JDK compile with `target` 21: the target is still 11; so, it doesn't test byte code with Java version 21; at least, that's my understanding. This means that Maven jdt.core has NOT been upgraded yet; similarly, the Java enum has not been updated. That part will be dealt with in a further PR.

JDK 21 is enforced with the toolchains: an automatic Maven profile is activated automatically. In that profile, the `release` option is `21` as well. Toolchains is not activated otherwise.

This is just the initial step towards Java 21, and I think it might be worthwhile to merge, if you think that's fine.


cdietrich:
what was the background behind the SuppressWarnings change?

LorenzoBettini:
> what was the background behind the SuppressWarnings change? (and other stubs)
> 

It's documented here https://github.com/eclipse/xtext/pull/2986/commits/27914447d0d711b3395e5e6d8145eb6f2fe76c3c#diff-331411999fe50a10ba8910972acaf1b716881e03a7f535f17573f49c8aabe528R18 
In Java 21 the annotation lost this part `@Target({ TYPE, FIELD, METHOD, PARAMETER, CONSTRUCTOR, LOCAL_VARIABLE, MODULE })`

> in the past we did if (JavaRuntimeVersion.isJava17OrLater()) ....

Concerning the stub classes in general it was suggested/discussed here https://github.com/eclipse/xtext/issues/2686#issuecomment-2039182574 by @szarnekow 

When it was unavoidable or simply easier I used the JavaRuntimeVersion but for other cases I created stubbed classes to keep use safer for future Java API evolution and to have our tests less fragile w.r.t. Java API evolutions.

LorenzoBettini:
@cdietrich @szarnekow I rebased the PR on top of current `main`.

Could you please give it a look and in case approve it so that I can continue working on the Java 21 issue?

cdietrich:
i am ok with the changes. @szarnekow wdyt

LorenzoBettini:
I rebased on the current main.

I recently added this https://github.com/eclipse/xtext/pull/2986/commits/8f8b905ed7080172ef40e119e9daaa016f2369c7
I know it sounds very strange, but that's the only way I found to make Maven/Tycho combination happy in using the toolchains generated by setup-java. I prefer to use that one especially if we also build with macOS Arm (https://github.com/eclipse/xtext/pull/3015) because manually managing the toolchains would become harder; it's not a problem in JIRO instead.

Finally, I'd like this one to be merged ASAP, please: my other fork's branch where I further experiment with Java 21 is build on top of this, and rebasing both this and the other branch is becoming a burden.

cc @cdietrich @szarnekow

LorenzoBettini:
@cdietrich @szarnekow I rebased and I'm waiting for the build to finished, then I'll merge this one and prepare the PR for the actual Java 21 support.

PR #2989, opened 2024-04-14
Various internal code simplifications and modernizations

HannesWell:
This PR contains the internal code simplifications and modernizations that happened as side effect of the following PRs, in order to ease their review:
- https://github.com/eclipse/xtext/pull/2978
- https://github.com/eclipse/xtext/pull/2979


org.eclipse.xtext/src/org/eclipse/xtext/mwe/RuntimeResourceSetInitializer.java
```diff
@@ -37,32 +37,21 @@ public class RuntimeResourceSetInitializer {
...
-				boolean result = true;
-				if (filter != null)
-					result = filter.matches(input);
-				if (result)
-					result = registry.getResourceServiceProvider(input) != null;
-				return result;
-			}
-		});
+		return traverser.resolvePathes(pathes,
+				input -> filter.matches(input) && registry.getResourceServiceProvider(input) != null);
```
szarnekow, line 54: Previously `filter` was only used if it wasn't null.

HannesWell: Right. I restored that behavior.
Since it isn't as simple as before then, I can also restore the previous code if you prefer.

szarnekow: The change is fine.

org.eclipse.xtext/src/org/eclipse/xtext/mwe/PathTraverser.java
```diff
@@ -54,20 +55,10 @@ public Set<URI> findAllResourceUris(String path, Predicate<URI> isValidPredicate
...
-						result.add(uri);
-					}
-				}
-				return result;
-			} finally {
-				zipFile.close();
+			try (ZipFile zipFile = new ZipFile(file);) {
+				return zipFile.stream().map(entry -> getUri(file, entry)) //
+						.filter(Objects::nonNull).filter(isValidPredicate) //
+						.collect(Collectors.toSet());
```
szarnekow, line 61: Another subtle change in behavior. `Collectors.toSet` doesn't make any API guarantees about mutability of the returned set. Regardless of my personal preference of iteration vs stream pipelines, I'd go for `collect(Collectors.toCollection(HashSet::new))`

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/ResourceSetBasedAllContainersState.java
```diff
@@ -84,7 +84,7 @@ public String toString() {
 				result.append("(empty)");
 			else {
 				result.append("{\n    ");
-				result.append(Joiner.on("\n    ").join(uris));
+				result.append(uris.stream().map(URI::toString).collect(Collectors.joining("\n    ")));
```
szarnekow, line 87: Quite a mouthful compared to Joiner.on. Both in terms of characters, cognitive load and object allocations. I'd keep this as a Joiner.

org.eclipse.xtext/src/org/eclipse/xtext/mwe/PathTraverser.java
HannesWell: That's right, but since `PathTraverser.traverseArchive()` respectively `findAllResourceUris()` does not make any guarantees as well I assumed it's fine to change this implementation detail. However, I applied your suggestion.

org.eclipse.xtext/src/org/eclipse/xtext/resource/containers/ResourceSetBasedAllContainersState.java
HannesWell: > Quite a mouthful compared to Joiner.on. Both in terms of characters, cognitive load and object allocations.

Yes it has more characters, but personally I would say it is simple since you can read the pipeline from left to right, but I'm more used to Streams than Guava's Joiner so it is a matter of taste.
I haven't counted all object creations, but Joiner also creates objects and the JVM+GCs are usually good in handling short living objects respectively even avoiding their creation on the heap.
But the biggest impact is probably due to the creation of the String result of the join operation and if you are really concerned about object creation here, I would try to avoid that by using something like the following instead:
```
for (Iterator<String> containers = invisibleContainers.iterator(); containers.hasNext();) {
	result.append(containers.next());
	if(containers.hasNext()) {
		result.append("\n    ");
	}
}
```
That being said, I reverted that change for now.

HannesWell:
Just rebased this on the latest master, is there anything else I should adjust?

org.eclipse.xtext/src/org/eclipse/xtext/scoping/Scopes.java
```diff
@@ -86,46 +80,38 @@ public static Iterable<IEObjectDescription> scopedElementsFor(Iterable<? extends
...
 		return Iterables.filter(transformed, Predicates.notNull());
 	}
-	
+
 	/**
 	 * indexes the IEObject description using the given
 	 */
 	public static <T> Multimap<T,IEObjectDescription> index(Iterable<IEObjectDescription> descriptions, Function<IEObjectDescription,T> indexer) {
-		ArrayList<IEObjectDescription> list = Lists.newArrayList(descriptions);
+		List<IEObjectDescription> list = new ArrayList<>();
```
szarnekow, line 97: It's ok here, but keep in mind that `Lists.newArrayList(Iterable)` has a special casing for Collections and will call the more optimal `ArrayList(Collection)` constructor, if possible.

HannesWell: Yes, that's right. But since the purpose of the entire creation of the `ArrayList` seems to be being able to determine the size of the `Iterable` and at the same time avoiding the second pass over the `Iterable` for the subsequent iteration, the best would be to to test if the `Iterable` is collection and act accordingly or provide an overload for collections.
The latter could then be called by the current method after the iterable content was copied into a List.

PR #2990, opened 2024-04-14
Replace Guava's ImmutableMap in XtextAntlrGeneratorFragment2

HannesWell:
All the changes are internal respectively just affects the internals of generated code.
 

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
```diff
@@ -390,31 +390,31 @@ class XtextAntlrGeneratorFragment2 extends AbstractAntlrGeneratorFragment2 {
 		'''
 			@�Singleton�
 			public static final class NameMappings {
-				
+
 				�IF partitions.size > 1�
 					�FOR partition : partitions.indexed�
 						private static final class Init�partition.key� {
-							private static void doInit(�ImmutableMap�.Builder<�AbstractElement�, �String�> builder, �grammar.grammarAccess� grammarAccess) {
+							private static void doInit(�Map�<�AbstractElement�, �String�> builder, �grammar.grammarAccess� grammarAccess) {
```
LorenzoBettini, line 397: I'd call the parameter `map` now

LorenzoBettini:
To be sure it doesn't break anything, shouldn't we regenerate all the languages with this change?

cdietrich:
yes but this is tedious work as many workflows dont work and we need to fiddle a lot with mwe deps to get them running

LorenzoBettini:
Actually, I fixed the workflows when I regenerated the test languages.

In any case, what I meant is that we must make sure this PR doesn't break the semantics

cdietrich:
@LorenzoBettini the workflows work but won’t start cause mwe launch and maybe others missing 
Last time I regenerated I added it. Regenerated and removed agin

LorenzoBettini:
@cdietrich which ones? The one of test languages works because I'm using it for other branches. Also the one of common types works IIRC. In that case it's a java launch, not a mwe2 launch. Same for xbase and xtend, they're java applications.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
HannesWell: I left it as it is since the passed map is still some kind of a 'builder' since it is just used to build the final immutable copy. And it required less charges. 
But if you prefer just `map`, I'll change that.

cdietrich:
@LorenzoBettini 

org.eclipse.xtext.builder.standalone.tests/src/org/eclipse/xtext/builder/tests/GenerateBuilderTestLanguages.mwe2
org.eclipse.xtext.extras.tests/src/org/eclipse/xtext/GenerateAllTestLanguages.mwe2
org.eclipse.xtext.web.example.entities/src/org/eclipse/xtext/web/example/entities/GenerateEntities.mwe2
org.eclipse.xtext.web.example.statemachine/src/org/eclipse/xtext/web/example/statemachine/GenerateStatemachine.mwe2

(Am not sure if i managed to run all)

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
LorenzoBettini: @HannesWell I see your point, but `builder` sounds strange to me; let's wait for @szarnekow 's review anyway.

LorenzoBettini:
@HannesWell now that https://github.com/eclipse/xtext/issues/2998 is fixed, you should be able to re-generate the languages and see if the build is still green. Maybe, however, it's better to wait for @szarnekow 's review before doing the re-generation of the languages.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
HannesWell: Understand. What about naming it about the content? So just `mapping`?
In general repeating the type in the variable name is often superfluous from a Clean-Code point of view.

HannesWell:
> @HannesWell now that #2998 is fixed, you should be able to re-generate the languages and see if the build is still green. Maybe, however, it's better to wait for @szarnekow 's review before doing the re-generation of the languages.

Thanks for fixing that. So should I simply import all languages into a secondard Eclipse that already contains this change and just re-run all workflows once we are fine with the change itself?

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
LorenzoBettini: `mappings` sounds fine to me

LorenzoBettini:
> Thanks for fixing that. So should I simply import all languages into a secondard Eclipse that already contains this change and just re-run all workflows once we are fine with the change itself?

@HannesWell, there's no need to do that in a secondary Eclipse: the mwe2 workflow uses the classes from the current workspace's classpath, where you already have the new fragments.

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/parser/antlr/XtextAntlrGeneratorFragment2.xtend
HannesWell: @szarnekow are you fine as well with `mappings`?

szarnekow: Yes, that’s a good fit

HannesWell:
> @HannesWell now that #2998 is fixed, you should be able to re-generate the languages and see if the build is still green. Maybe, however, it's better to wait for @szarnekow 's review before doing the re-generation of the languages.

Since we all agreed on the new variable name above and there where no other objectsion I rerun all worksflows and added the re-generated Parser implementations in a second commit to this PR.

Please let me know if there is anything that should be changed.

HannesWell:
In the failing test `dummyParser.getRuleName(null);` is called for which the implementation returned by `Map.copyOf()` throws an NPE. Is this a relevant use-case? If yes the generated code has to be adapted to guard null arguments.

szarnekow:
> Alternatively the HashMap could only be wrapped into a `Collections.unmodifiableMap()` instead of using `Map.copyOf()`.

That would be my preference

PR #2991, opened 2024-04-14
Replace concatenates Strings by Text-Blocks in CompilerTests

HannesWell:
This improves the readability of Strings defining the test-input and expected output. It avoids the need to explicitly add new-line characters and the need to escape double-quotes.

Currently the tests require only Java-11, but it was said that as part of https://github.com/eclipse/xtext/issues/2686 Java-17 will be the required minimum.


org.eclipse.xtext.xbase.tests/.classpath
```diff
@@ -20,7 +20,7 @@
 			<attribute name="test" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11/"/>
```
cdietrich, line 23: am not sure if i am ok with that just for the test

HannesWell: This PR aims to the time, when Java-17 is already the required minimum. I assumed this will the case as soon Java-21 is supported? So this is actually a draft.
If you are interested in a PR that bumps the required Java-version for all bundles, I can provide one just for that.

HannesWell:
Since the test projects require Java-17 in the meantime, this should be ready.

LorenzoBettini:
I seem to understand that the Jenkins build fails because it's referring to a non-existent TP.
@HannesWell, are you sure this PR is on top of the current main branch?

PR #2992, opened 2024-04-14
Replace Guava's Objects by java.util.Objects

HannesWell:
All changes are competently internal.

This change intentionally does not cover `XbaseCompiler` and `ObjectExtensions` (which were at least my actual goal). It will be covered in a separate PR together will all the Xtend generated Java sources and adjustments of the test expectations.


LorenzoBettini (approved):
@szarnekow I'm fine with this, what do you think?

LorenzoBettini:
Thanks @HannesWell

HannesWell:
Thank you for reviewing.

PR #2993, opened 2024-04-14
Emit java.util.Objects in generated code instead of Guava's Objects

HannesWell:
This PR adjusts `XbaseCompiler` and `ObjectExtensions` to emit `java.util.Objects` instead Guava's `Objects` class for generated code. `java.util.Objects.equals` is a drop-in replacement for Guava's `Objects.equal()`.
All other changes are either just adoption of compiler tests or updates of xtend-generated code committed into this repository.

Some changes in the code generated for the examples is done in anticipation of this change. Without this change being submitted the compiler respectivly the Xtend-Container cause the old code to be generated.

This is the complement of https://github.com/eclipse/xtext/pull/2992.


HannesWell:
Many changes in the generated code could be avoided if the (some of) the compiler tests would be migrated to pure Java classes using text-blocks, like in https://github.com/eclipse/xtext/pull/2991.

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/macro/AbstractReusableActiveAnnotationTests.java
```diff
@@ -4885,15 +4885,15 @@ public void testCreateTypeFromUsage() {
...
...e\r\n\t\t\t\timport org.eclipse.xtend.lib.macro.TransformationContext\r\n\t\t\t\timport org.eclipse.xtend.lib.macro.TransformationParticipant\r\n\t\t\t\timport org.eclipse.xtend.lib.macro.declaration.MutableFieldDeclaration\r\n\t\t\t\timport org.eclipse.xtend.lib.macro.declaration.MutableClassDeclaration\r\n\r\n\t\t\t\t@Active(InitProcessor)\r\n\t\t\t\tannotation Init { }\r\n\t\t\t\tclass InitProcessor implements TransformationParticipant<MutableFieldDeclaration> {\r\n\t\t\t\t\t\r\n\t\t\t\t\toverride doTransform(List<? extends MutableFieldDeclaration> annotatedTargetFields, extension TransformationContext context) {\r\n\t\t\t\t\t\tval ctor = annotatedTargetFields.head.declaringType.addConstructor [\r\n\t\t\t\t\t\t\tprimarySourceElement = declaringType\r\n\t\t\t\t\t\t\tbody = [\'\'\'\r\n\t\t\t\t\t\t\t\t�FOR f : annotatedTargetFields�\r\n\t\t\t\t\t\t\t\t\tthis.�f.simpleName� = \"foo\";\r\n\t\t\t\t\t\t\t\t�ENDFOR�\r\n\t\t\t\t\t\t\t\'\'\']\r\n\t\t\t\t\t\t]\r\n\t\t\t\t\t\tannotatedTargetFields.forEach [ field |\r\n\t\t\t\t\t\t\tfield.setFinal(true)\r\n\t\t\t\t\t\t\tfield.markAsRead\r\n\t\t\t\t\t\t\tfield.markAsInitializedBy(ctor)\r\n\t\t\t\t\t\t]\r\n\t\t\t\t\t}\r\n\t\t\t\t}\r\n\t\t\t");
```
LorenzoBettini, line 4896: Also here "\r" and below.

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/smoke/SmokeTestCase.java
```diff
@@ -417,7 +417,7 @@ public void test_03() {
...
+      this.file("\r\n\t\t\tpackage org.eclipse.xtend.core.tests.smoke\r\n\t\t\t\r\n\t\t\tclass Case_4 {\r\n\t\t\t\tdef richStrings_01() {\r\n\t\t\t\t\t\'\'\'foobar\'\'\'\r\n\t\t\t\t}\r\n\t\t\t\tdef richStrings_02()\r\n\t\t\t\t\t\'\'\'�\'start\'�\r\n\t\t\t\t\t  first line\r\n\t\t\t\t\t  second line\r\n\t\t\t\t    \t\t\'\'\'\r\n\t\t\t\tdef richStrings_03() {\r\n\t\t\t\t\t\'\'\'�IF false�foobar�ELSEIF false�foobar�ELSE�zonk�ENDIF�\'\'\'\r\n\t\t\t\t}\r\n\t\t\t\tdef richStrings_04() {\r\n\t\t\t\t\t\'\'\'  �FOR a:\'1\'.toCharArray��FOR a2:\'1\'.toCharArray�foobar�ENDFOR��ENDFOR�\'\'\'\r\n\t\t\t\t}\r\n\t\t\t\tdef richStrings_05() {\r\n\t\t\t\t\t\'\'\'\r\n\t\t\t\t\t\t�FOR a:\'1\'.toCharArray�\r\n\t\t\t\t\t\t  �FOR a2:\'12\'.toCharArray�foobar�ENDFOR�\r\n\t\t\t\t\t\t�ENDFOR�\r\n\t\t\t\t\t\'\'\'\r\n\t\t\t\t}\r\n\t\t\t}\r\n\t\t");
```
LorenzoBettini, line 420: Also here

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/EmfModelsTest.java
```diff
@@ -225,48 +225,48 @@ private Object getDefaultValue(final Class<?> clazz) {
     boolean _isPrimitive = clazz.isPrimitive();
     if (_isPrimitive) {
       boolean _matched = false;
-      if (Objects.equal(clazz, Boolean.TYPE)) {
+      if (com.google.common.base.Objects.equal(clazz, Boolean.TYPE)) {
```
LorenzoBettini, line 228: Why here there's still Objects.equal? I've seen that also below in another file

LorenzoBettini:
I've just left a few comments, I see @szarnekow is already reviewing that.

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/compiler/CompilerXtendIssue230Test.java
```diff
@@ -170,6 +170,6 @@ public void testIssue230() {
     _builder.newLine();
     _builder.append("}");
     _builder.newLine();
-    this.assertCompilesTo("\n\t\tpackage b\n\t\t\n\t\timport org.eclipse.xtend.lib.annotations.Accessors\n\t\t\n\t\tclass MyTest {\n\t\t\t\n\t\t\t@Accessors\n\t\t\tstatic class A1 {String name}\n\t\t\t@Accessors\n\t\t\tstatic class A2 {String name}\n\t\t\t\n\t\t\tdef doItWithNumber(Object n) \'\'\'\n\t\t\t�IF n instanceof A1�\n\t\t\t\tA1: �n.name�\n\t\t\t�ELSEIF n instanceof A2�\n\t\t\t\tA2: �n.name�\n\t\t\t�ELSE�\n\t\t\t\tElse: �n�\n\t\t\t�ENDIF�\n\t\t\t\'\'\'\n\t\t}\n\t\t", _builder);
+    this.assertCompilesTo("\r\n\t\tpackage b\r\n\t\t\r\n\t\timport org.eclipse.xtend.lib.annotations.Accessors\r\n\t\t\r\n\t\tclass MyTest {\r\n\t\t\t\r\n\t\t\t@Accessors\r\n\t\t\tstatic class A1 {String name}\r\n\t\t\t@Accessors\r\n\t\t\tstatic class A2 {String name}\r\n\t\t\t\r\n\t\t\tdef doItWithNumber(Object n) \'\'\'\r\n\t\t\t�IF n instanceof A1�\r\n\t\t\t\tA1: �n.name�\r\n\t\t\t�ELSEIF n instanceof A2�\r\n\t\t\t\tA2: �n.name�\r\n\t\t\t�ELSE�\r\n\t\t\t\tElse: �n�\r\n\t\t\t�ENDIF�\r\n\t\t\t\'\'\'\r\n\t\t}\r\n\t\t", _builder);
```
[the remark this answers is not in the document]
HannesWell: Ahh. Yes, I noticed already that this can happen as the changed java-files poluted my git staging. I actually wanted to check it before creating creating the PR, but forgot it. Sorry for that and thanks for spotting this.
I'm about to fix that.
In general this problem could be avoided if those tests would be converted to plain java classes using text-blocks, similar to https://github.com/eclipse/xtext/pull/2991, but this would require a BREE of Java-17.
As far as I understand the plan for Java-21 you intend to raise the required java version to 17 with the support of Java-21.
I don't know how far your work on the Java-21 support has already reached. But if you are fine with it, I can provide a PR already now to raise the required Java-version for all Xtext bundles (in the `MANIFEST.MF`, `.classpath` and JDT settings, plus root pom.xml) and then a subsequent one to migrate the tests as suggested above.
This would also reduce the number of changes in this PR.

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/EmfModelsTest.java
HannesWell: That remained since it is generated by the `XbaseCompiler` which is adapted just in this PR and my development-eclipse doesn't have this fix yet. I oversaw that since when I didn't have https://github.com/eclipse/xtext/pull/2992 in the same branch and also had my search configured to not contain derived resources.

I have now used a secondary Eclipse launched from my Xtext workspace to regenerated all the java files that still contained Guava's Objects class.
Now, with this change I don't find any occurrence of `com.google.common.base.Objects` in my Xtext workspace, not even in derived resources.

cdietrich: Usually we do this update with a milestone update for the xtend version in the root pom

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/compiler/CompilerXtendIssue230Test.java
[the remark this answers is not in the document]
LorenzoBettini: I'd wait for the transition to Java text blocks and Java 17.

In particular, similar to the others you experienced here, this issue is due to tests not using Xtend template strings (`'''`) but standard quoted strings that keep the `\r`.

I don't think those tests were meant to use standard strings, and maybe at that time, it went unnoticed because most developers were using macOS or Linux.

I've opened an issue to track that: https://github.com/eclipse/xtext/issues/2999

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/EmfModelsTest.java
LorenzoBettini: @HannesWell I didn't think about that.

I typically use the latest nightly in my Xtext workspace.

@cdietrich do you prefer to keep the old version of the generated files for the moment?

HannesWell:
M2 is approaching so I wonder if this should be landed before it and if there is anything I can do to make this to completion?

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/EmfModelsTest.java
cdietrich: Do you have Cappa to regen all

cdietrich: Also is this about fragment or about xtend 
If xtend regen with the m2 boot strap

LorenzoBettini: @cdietrich this is not about the fragment but about a change in XbaseCompiler, so the regenerated Java code from Xtend could be done once we install the new Xtend in the development workspace (i.e., when this is merged to master, the nightly p2 site is published, and the development workspace takes Xtend from the nightly p2 site). When we publish M1, the xtend maven plugin will also behave accordingly.

LorenzoBettini:
As for me, we could merge this; the updated generated Java files can be done in another PR, once this update is part of the nightly update site.

szarnekow:
Ok. But let's make sure we regen Xtend before M2

szarnekow (approved):

org.eclipse.xtend.core.tests/xtend-gen/org/eclipse/xtend/core/tests/EmfModelsTest.java
cdietrich: Yes this is why I hesitate to regen rn. If we can burn milestones we could build it now after merge and then for the real milestone but I guess it’s not Wort or

HannesWell:
> As for me, we could merge this; the updated generated Java files can be done in another PR, once this update is part of the nightly update site.

So should I revert the anticipatory changes of generated code or leave it as it is?

cdietrich:
I assume you can leave it as the build will replace then anyway

HannesWell:
> I assume you can leave it as the build will replace then anyway

Yes the build replaces it automatically, but at the moment this PR already contains the changes the XtendBuilder **will** do once this change is available in the development Eclipses (I created the current state with a secondary Eclipse launched with this PR).
So without this change available, i.e. if you don't update to the nightly build immediatly you will get the following changes in your WS that undo parts of this PR:
grafik

But from your previous comments I understand it that you don't want to have that anticipatory changes now but apply it once you update the dev-Eclipses to M2?

cdietrich:
I don’t care. Repo code differs from shipped code for m2

szarnekow:
> if you don't update to the nightly build immediatly

It's necessary in any case. So no worries. Thank you for the modernization of the generated code!

HannesWell:
> > if you don't update to the nightly build immediatly
> 
> It's necessary in any case. So no worries. Thank you for the modernization of the generated code!

Great and you are welcome. I'm glad to get more modern code generated. :)

Since everbody seem to be fine with this change can it be submitted or should I rebase it on the current master first?

LorenzoBettini (approved):

LorenzoBettini:
@HannesWell Thanks! I've just merged.

LorenzoBettini:
The first one who updates the development environment to the nightly update site can then create a PR with the regenerated code.

LorenzoBettini:
@cdietrich I don't understand what you were proposing here:

> Yes this is why I hesitate to regen rn. If we can burn milestones we could build it now after merge and then for the real milestone but I guess it’s not Wort or

HannesWell:
> @HannesWell Thanks! I've just merged.

Thank you. :)

cdietrich:
Normally I would wait after the milestone or build an extra milestone but I have no Cappa so I don’t care

PR #2996, opened 2024-04-15
Catch exceptions while finding references

rubenporras:
If one entry cannot be processed due to exceptions, catch such exceptions and log them, but still return other found references.

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
@@ -13,6 +13,7 @@ import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 
+import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EClass;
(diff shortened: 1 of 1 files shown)


rubenporras:
For https://github.com/eclipse/xtext/issues/2997

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/symbol/DocumentSymbolService.java
```diff
@@ -55,7 +56,7 @@
  * @author kosyakov - Initial contribution and API
  *
  * Contributors: 
- *   Rub�n Porras Campo (Avaloq Evolution AG) 
+ *   Rub�n Porras Campo (Avaloq Evolution AG)
```
cdietrich, line 59: encoding issue here

rubenporras: I was wondering about that as well. Do you know which encoding should I use?

LorenzoBettini:
I restarted the Jenkins build, I guess it was failed because of problems accessing Maven repositories

LorenzoBettini:
Yes, the Jenkins build is also green!

PR #3002, opened 2024-04-20
fixed generators for test and example languages

### LINKED ISSUE #2998: Check state of test and example languages generators
See https://github.com/eclipse/xtext/pull/2990#issuecomment-2058181195

LorenzoBettini:
Closes #2998

I also added the small chances after running generators

### DIFF
--- a/org.eclipse.xtext.builder.standalone.tests/build.properties
+++ b/org.eclipse.xtext.builder.standalone.tests/build.properties
@@ -4,4 +4,5 @@ bin.includes = .,\
                model/,\
                META-INF/,\
                plugin.xml
-               
+additional.bundles = org.eclipse.emf.mwe2.launch,\
+                     org.eclipse.xtext.xtext.generator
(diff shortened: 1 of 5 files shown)


org.eclipse.xtext.builder.standalone.tests/src-gen/org/eclipse/xtext/builder/tests/builderTestLanguage/impl/BuilderTestLanguagePackageImpl.java
```diff
@@ -72,7 +72,6 @@ private BuilderTestLanguagePackageImpl()
   {
     super(eNS_URI, BuilderTestLanguageFactory.eINSTANCE);
   }
-
```
cdietrich, line 75: is these changes cause you dont use emf 2022-03 as tp?

LorenzoBettini: @cdietrich it can be: I'm always using the latest TP. Do you want me to remove them from the PR?

cdietrich: this is an "argggggggg" problem. strictly i like to have the variant with the minimal emf version.
so yes. but i can also live with it.
am not sure if we will run into a new emf version where there is a real change

LorenzoBettini: @cdietrich I've pushed force without the modified Java files

PR #3004, opened 2024-04-20
Normalization of Line Terminators in Xbase strings

### LINKED ISSUE #2293: Wrong line delimiter when enclosing multi-line strings with double quotes.
After checking out a fresh Xtend workspace on windows, there are several java files in xtend-gen folder marked as dirty:
screenshot1

screenshot2
screenshot3
screenshot4
screenshot5

As far as I can see, the reason for the change is that multi line Strings enclosed by double quotes are translated differently on windows and on mac.

**Hint:** by multi line String enclosed by `'''` the `newLine()` method is used in the generated *.java file. I suggest to also use this approach for multi-line Strings enclosed by `"`.

LorenzoBettini:
Closes #2293

### DIFF
--- a/org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/evaluation/AbstractXbaseEvaluationTest.java
+++ b/org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/evaluation/AbstractXbaseEvaluationTest.java
@@ -912,7 +912,7 @@ public abstract class AbstractXbaseEvaluationTest extends Assert {
 	}
 	
 	@Test public void testStringLiteral_03() throws Exception {
-		assertEvaluatesTo("lite\r\nr\\al", "'lite\r\nr\\\\al'");
+		assertEvaluatesTo("lite\nr\\al", "'lite\r\nr\\\\al'");
 	}
 	
 	@Test public void testStringLiteral_04() throws Exception {
--- a/org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/evaluation/AbstractXbaseEvaluationTest.java
+++ b/org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/evaluation/AbstractXbaseEvaluationTest.java
@@ -914,7 +914,15 @@ public abstract class AbstractXbaseEvaluationTest extends Assert {
 	@Test public void testStringLiteral_03() throws Exception {
 		assertEvaluatesTo("lite\nr\\al", "'lite\r\nr\\\\al'");
 	}
-	
+
+	@Test public void testStringLiteral_03_Issue_2293_01() throws Exception {
+		assertEvaluatesTo("lite\nr\\al", "'lite\\nr\\\\al'");
+	}
+
+	@Test public void testStringLiteral_03_Issue_2293_02() throws Exception {
+		assertEvaluatesTo("lite\r\nr\\al", "'lite\\r\\nr\\\\al'");
+	}
+
 	@Test public void testStringLiteral_04() throws Exception {
 		assertEvaluatesTo('\n', "{ val char x = '\n' x}");
 	}
--- a/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/conversion/XbaseValueConverterService.java
+++ b/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/conversion/XbaseValueConverterService.java
@@ -276,7 +276,7 @@ public class XbaseValueConverterService extends DefaultTerminalConverters {
 	public static class WindowsEOLsAwareSTRINGValueConverter extends STRINGValueConverter {
 		@Override
 		public String toValue(String string, INode node) {
-			return Strings.toUnixLineSeparator(super.toValue(string, node));
+			return super.toValue(Strings.toUnixLineSeparator(string), node);
 		}
 	}
 }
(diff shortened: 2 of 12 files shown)


LorenzoBettini:
Looks like I haven't gotten the replacement correctly: it also leads to this unwanted difference:

image

I thought this test would check that https://github.com/eclipse/xtext/pull/3004/commits/efa5aa17de66895b3952271ad21b38b8a9e17619#diff-7217ccbb0669983e16ca897570a543e7ee1bfb23a156c11fe9425cc5db1536baR2210
but maybe this only checks the input string "\\r", while we need to check "\r", that is when the user enters the 2 characters "\r", while I'd like to remove the single character '\r'.

LorenzoBettini:
OK, this https://github.com/eclipse/xtext/pull/3004/commits/fadf9c715a71b064f7319392433257e7287a9956 reproduces the problem and fixes it

LorenzoBettini:
@szarnekow the build fails. I guess I haven't override all the needed methods in the custom value converter?

org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/conversion/XbaseValueConverterService.java
```diff
@@ -251,4 +254,29 @@ public Integer toValue(String string, INode node) {
...
+	 * https://github.com/eclipse/xtext/issues/2293 This is aligned with Java
+	 * text blocks' "Normalization of Line Terminators"
+	 * 
+	 * @author Lorenzo Bettini - Initial contribution and API
+	 * @since 2.35
+	 */
+	public static class WindowsEOLsAwareSTRINGValueConverter extends STRINGValueConverter {
+		@Override
+		public String toValue(String string, INode node) {
+			return Strings.toUnixLineSeparator(super.toValue(string, node));
```
szarnekow, line 279: It's an interesting case. I guess we want to treat an explicit `\r` in the string literal different from a regular windows newline. I've to think about it.

szarnekow: cf
```java
	public static void main(String[] args) {
		String s = """
				some\r\nstring""";
		System.out.println(s);
		System.out.println(s.length());
		System.out.println('\r' == s.charAt(4));
	}
```

LorenzoBettini: @szarnekow a quick thought: I guess the original string should be normalized before passing it to super, not afterward as I'm doing now

LorenzoBettini:
Passing the already normalized string to super:

```java
	public static class WindowsEOLsAwareSTRINGValueConverter extends STRINGValueConverter {
		@Override
		public String toValue(String string, INode node) {
			return super.toValue(Strings.toUnixLineSeparator(string), node);
		}
	}
```

leads to the expected test result:

```java
	@Test
	public void testStringLiteralWithCarriageReturn3_Issue2293() throws Exception {
		compilesTo(
				"{\"astring\".toString.replaceAll(\"\\r\\n\", \"\\n\");}",
				"return \"astring\".toString().replaceAll(\"\\r\\n\", \"\\n\");");
	}
```

LorenzoBettini:
@szarnekow concerning the failing tests in Xtend, I think I found the culprit: the way I customized the Xbase string converter did not allow for customizations, and the Xtend customizations were not picked anymore. This should do: https://github.com/eclipse/xtext/pull/3004/commits/9b8cc604f4921b2d7a80f58fe9b5d04e2033f9af

P.S., do you think it's worthwhile to rename the Xtend `StringValueConverter` to `XtendStringValueConverter`? That would allow for easier discovery.

LorenzoBettini:
@szarnekow I rebased the PR

This is still to be done if you think it's feasible:

> P.S., do you think it's worthwhile to rename the Xtend `StringValueConverter` to `XtendStringValueConverter`? That would allow for easier discovery.

szarnekow:
> feasible

Yes

> XtendStringValueConverter

Yes

LorenzoBettini:
@szarnekow rebased and renamed the Xtend string value converter.
As for me, this PR would be ready.

org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/evaluation/AbstractXbaseEvaluationTest.java
```diff
@@ -912,7 +912,7 @@ public abstract class AbstractXbaseEvaluationTest extends Assert {
 	}
 	
 	@Test public void testStringLiteral_03() throws Exception {
-		assertEvaluatesTo("lite\r\nr\\al", "'lite\r\nr\\\\al'");
+		assertEvaluatesTo("lite\nr\\al", "'lite\r\nr\\\\al'");
```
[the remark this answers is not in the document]
LorenzoBettini: @szarnekow this one https://github.com/eclipse/xtext/pull/3004/commits/d7c67a38d2ac345fd935ad15f3c8732f4f5d0e16 ?

Yes, we have a duplicated AbstractXbaseEvaluationTest ;)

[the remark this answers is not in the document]
szarnekow: I knew that I saw it somewhere ... damn it :)

Thanks!

szarnekow (approved):
Please add a reminder for the N&N

Issue #2778, opened 2023-08-11
Line ending confusion in generated files of Xtext project?


Bananeweizen:
When I create a new Oomph setup for the Xtext project on my Windows machine, then Oomph automatically sets autocrlf=true, even though there is no such attribute in the *.setup file:
![grafik](https://github.com/eclipse/xtext/assets/406876/b9752952-b986-4ead-bf3a-0d649ee19391)

That in turn leads to all the Xtend generated Java files having CRLF in my local index (while the Github repo uses LF). But during build and re-generation all these files get LF only (which fits to the projects having the Unix line ending setting in their project properties). Therefore the IDE has 800+ changed files after a restart, almost all of them with newline changes only, looking like this:
![grafik](https://github.com/eclipse/xtext/assets/406876/57aa8d36-90dd-4ab6-9ebb-3d03f92e1c0e)

I do use a personal Oomph setup, but I'm not aware of any setting that would cause Oomph to add the autocrlf just for me, so I suspect this should be broken similarly for everyone on Windows. If that is true, would it be useful to explicitly set autocrlf to another value in the *.setup or alternatively to set `*.java text eol=lf` in a `.gitattributes` file eventually? I've used the gitattributes approach locally, reset hard, and since then my staging area has only 24 changes files, where most contain actual non whitespaces changes (which would be another thing to investigate later).

cdietrich:
this issue is new to me. 
i am only aware on windows changes on fresh workspaces caused by 
https://github.com/eclipse/xtext/issues/2293

unfortunately with company change i have no suiteable windows machine anymore to have a look.

Bananeweizen:
The Windows changes that you linked are in the above mentioned remaining 24 changed files after I applied the gitattributes. I can try to have a colleague set up another Xtext installation on his machine next week to verify he sees the same.
My initial thought was that this may have been caused by the monorepo change some weeks/months ago, but I have not been able to pinpoint to any changes in the .setup or in the archived repos that would support that claim.

cdietrich:
maybe a default on oomph or egit side has changed?

pbrossel:
Hi @Bananeweizen,
can you tell me what changes you have made to temp fix the issue ?
I did an initial oomph setup on windows and have exactly the same problem as you had.
>...
I've used the gitattributes approach locally, reset hard, and since then my staging area has only 24 changes files, where most contain actual non whitespaces changes (which would be another thing to investigate later).

LorenzoBettini:
[...]

PR #3005, opened 2024-04-22
revert "Set UNIX-style linebreaks"

LorenzoBettini:
Take the EOL settings from the OS, so that Xtend generated Java files are not "dirty" when regenerated in a Windows system.

Closes https://github.com/eclipse/xtext/issues/2778

### DIFF
--- a/org.eclipse.xtend.core.tests/.settings/org.eclipse.core.runtime.prefs
+++ b/org.eclipse.xtend.core.tests/.settings/org.eclipse.core.runtime.prefs
@@ -1,2 +1,0 @@
-eclipse.preferences.version=1
-line.separator=\n
--- a/org.eclipse.xtend.core/.settings/org.eclipse.core.runtime.prefs
+++ b/org.eclipse.xtend.core/.settings/org.eclipse.core.runtime.prefs
@@ -1,2 +1,0 @@
-eclipse.preferences.version=1
-line.separator=\n
--- a/org.eclipse.xtend.doc/.settings/org.eclipse.core.runtime.prefs
+++ b/org.eclipse.xtend.doc/.settings/org.eclipse.core.runtime.prefs
@@ -1,2 +1,0 @@
-eclipse.preferences.version=1
-line.separator=\n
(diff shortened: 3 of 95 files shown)


LorenzoBettini:
By the way, this change is of course harmless to Linux and macOS

cdietrich:
@szarnekow wdyt?

LorenzoBettini:
In windows, this allows me to avoid tons of dirty files.

cdietrich:
unfortunately this is in the codebase "forever"
i also cannot test if there is any bad side effect.
what does this preference do?

what happens if someone opens a pr on windows?

LorenzoBettini:
@cdietrich I'll try to summarize the reasons for this PR and the problems of https://github.com/eclipse/xtext/issues/2778

> unfortunately this is in the codebase "forever" i also cannot test if there is any bad side effect. what does this preference do?

This preference, which I remove in this PR, tells Eclipse that every file generated from Eclipse must have Unix EOL, while the default is to use the System EOL.

Using such a setting, together with the recommended setting in Windows autocrlf=true, breaks the Windows experience: as soon as you run a clean build from Eclipse in Windows, **all** Java files generated by Xtend, in the projects with this setting, will be "dirty" since they will have Unix EOL. About 500 files will be dirty. Note that these settings are not even in all projects with Xtend files since they were added so many years ago.

Remember that autocrlf=true means that in Git files are with Unix EOL, and when checkout they get Windows EOL. When committed they will get Unix EOL.

As @szarnekow suggested, those settings, dated back to 2012, were probably needed because at that time, we committed trace files as well (we were using Buckminster) and that setting allowed for consistent trace files. Nowadays, trace files are generated during the build, where, by the way, those settings are ignored (they are relevant only in Eclipse), so they are not needed anymore for that matter.

> 
> what happens if someone opens a pr on windows?

Without this PR, files might end up in the Git repository with wrong EOLs and/or with too many unrelated changes. Not to mention that Windows contributors will have hard time understanding what to commit, and, as it happened recently, a few files will not be committed.

https://github.com/eclipse/xtext/pull/2977 already fixed many files that ended up with the wrong EOL. The presence of these settings is probably the culprit behind the files with wrong EOLs that ended up in the Git repository in the past.

Please let me know if you need further information.

cdietrich (approved):
So I assume this won’t have side effects

LorenzoBettini:
> So I assume this won’t have side effects

I haven't found any; I found lots of side effects without this PR ;)

@szarnekow are you OK with this PR?

szarnekow:
Yes

PR #3008, opened 2024-04-23
use stable tags for GitHub Actions virtual environments

LorenzoBettini:
macOS in GitHub Actions switched to arm by default

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -48,6 +48,19 @@ jobs:
         with: 
           run: ./mvnw clean verify "-Dmaven.home=${{ env.MAVEN_WRAPPER_HOME }}" -PuseJenkinsSnapshots -f org.eclipse.xtext.full.releng
 
+      - name: Archive Test Results
+        uses: actions/upload-artifact@v2
+        if: ${{ failure() || cancelled() }}
+        with:
+          name: test-results-${{ runner.os }}
+          path: '**/*.tests/target/surefire-reports'
+      - name: Archive Logs
+        uses: actions/upload-artifact@v2
+        if: ${{ failure() || cancelled() }}
+        with:
+          name: logs-${{ runner.os }}
+          path: '**/*.tests/target/**/*.log'
+
   build-maven-artifacts:
     if: github.event_name != 'pull_request' || github.event.pull_request.base.repo.clone_url != github.event.pull_request.head.repo.clone_url
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


szarnekow:
I think it's the right move. How do we know when to update the tag, though?

LorenzoBettini:
> I think it's the right move. How do we know when to update the tag, though?

Now and then, we have to keep an eye on https://github.com/actions/runner-images

When something new comes out, we should test it.
When something is discontinued, the builds will fail ;)

This month, the new Ubuntu LTS 24.04 will be released. It should be available on GitHub Actions after sometime.

If the build is green, I'd immediately merge this so other PRs will not fail. Is that OK?

szarnekow:
> immediately merge

Yes, please.

szarnekow (approved):

LorenzoBettini:
Argh! Now it fails on macOS for this

```
Results:
Tests run: 4233, Failures: 0, Errors: 0, Skipped: 63
Error:  Timeout of 2400s exceeded. Process was killed
```

but maybe it's just a hiccup, I saw that on Windows jobs as well... will try to restart it.

LorenzoBettini:
No... it always fails at the same point...

LorenzoBettini:
I updated the workflows to archive logs and test results in case of failures... let's see.
It might also be that previously we were using an older version of macOS, while this is too recent

```
macOS
  14.4.1
  23E224
```

in my fork, where it still works, it still uses

```
macOS
  12.7.4
  21H1123
```

cdietrich:
do we know if adding arm variants to the environments would help?

LorenzoBettini:
Yes, if we add it to the `target-platform-configuration` in the POM then it will use the arm version; I'm already doing that for my DSLs, though I had never tested it in GitHub Actions.

The problem is, IIRC, that the build will fail if we use a `.target` file for an old version of Eclipse that didn't provide the arm version of Eclipse.
However, the oldest platform we use is 2022-03, which already provided arm versions: https://download.eclipse.org/releases/2022-03/202203161000/binary/ so we should be safe!

LorenzoBettini:
The archived log https://github.com/eclipse/xtext/actions/runs/8799576072/artifacts/1439346225 of the xbase.ui.tests, where the job fails with timeout, shows lots of:

```
java.lang.Exception: ThreadDump for thread "Java indexing"
	at java.base@17.0.10/java.lang.Object.wait(Native Method)
	at java.base@17.0.10/java.lang.Object.wait(Object.java:338)
```

I'll try first to use an older version of macOS, OK?
Or should we go for arm?

LorenzoBettini:
> do we know if adding arm variants to the environments would help?

@cdietrich btw, I created this https://github.com/eclipse/xtext/pull/3015

PR #3009, opened 2024-04-24
fix dm hover test on j 21

cdietrich:


LorenzoBettini (commented):
But this would work only in Java 21, not in previous versions.

LorenzoBettini:
I mean that the modified test will rely only on the new Javadoc of List introduced in Java 21. In previous versions of Java it will fail because the Javadoc was different. If we're sure we change the domainmodel to require Java 21, then no problem.

Otherwise

- we rely on a custom stubbed class with a fixed javadoc (see my other changes in https://github.com/eclipse/xtext/pull/2986, which is the safest choice because we have full control)
- we use the check for the current Java version and specify the assertions accordingly
- we introduce a less strict method in the hover test, something like hasHoverContaining

cdietrich:
i dont understand.
https://github.com/xtext/xtext-reference-projects/actions/runs/8813287543
dm is green here will all java versions

LorenzoBettini:
Sorry, I missed the fact it was already a containing check.

And sorry: I thought you changed the tests to add the `(also known as a <i>sequence</i>)`, instead you changed it to remove it because it disappeared from Java 21 javadoc.

Sorry for the noise, I now understand :)

LorenzoBettini (approved):

PR #3015, opened 2024-04-25
Use macOS Arm in GitHub Actions

### LINKED ISSUE #3014: Use macOS Arm on GitHub Actions
I made a few experiments in my fork and the arm version works really well and fast, comparable to Ubuntu:

Screenshot from 2024-04-24 16-54-07

compare that with the x86 version of macOS we use now:

image

LorenzoBettini:
Closes #3014 

macOS arm is much faster than macOS x86, almost as Ubuntu.


.github/workflows/maven.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     if: github.event_name != 'pull_request' || github.event.pull_request.base.repo.clone_url != github.event.pull_request.head.repo.clone_url
     strategy:
       matrix:
-        os: ['ubuntu-22.04', 'macos-12', 'windows-2022']
+        os: ['ubuntu-22.04', 'macos-14', 'windows-2022']
```
szarnekow, line 17: Given that we also have cocoa x86_64 in our env, should we still run against macos-12?

LorenzoBettini: @szarnekow we could, but we'd lose the advantage of faster macOS environment in GitHub Actions ;)

LorenzoBettini: > True. It would be even faster if we skip all environments, no? Kidding.

And we would even have less releng/CI troubles! ;)

> 
> Question is: What do we aim for? Short build times or more coverage?

I see your point. I can also put back the old macOS x86 configuration. We can then decide whether to ditch one if it becomes too flaky

LorenzoBettini:
@szarnekow I've updated the PR accordingly.

I forgot to mention: the jobs will run in parallel but now we have lots of them, and one more once https://github.com/eclipse/xtext/pull/2986 is merged. I think Eclipse has been granted more GitHub Actions processing time, is that right @cdietrich ? If it's so, it shouldn't be a big problem having so many jobs.

LorenzoBettini:
@szarnekow @cdietrich macOS x86 in GitHub Actions is getting unusable. Mac 12 is so slow that it almost always fails in building in UI tests. I've just tried Mac 13 but it also fails with "process returned error code 143".

I'd switch to macOS Arm, which, as I said above, is as fast as Ubuntu, and seems reliable.
In the end, since we build on macOS anyway, I don't think we miss crucial cases if we only build on arm.

cdietrich (approved):

PR #3019, opened 2024-04-26
update to mwe 2.18.0.M2

cdietrich:


LorenzoBettini (approved):
You forgot to increment the sequence number in .target files, but that's not a problem ;)

szarnekow (approved):
Thank you!

PR #3020, opened 2024-04-29
Oomph clean build onlyNewProjects="true"

LorenzoBettini:
This should deal with the problem of always getting a clean build in Windows, each time you open the development workspace.

See https://github.com/eclipse/xtext/issues/2778#issuecomment-2071963236

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -565,6 +565,7 @@
     <setupTask
         xsi:type="projects:ProjectsBuildTask"
         id="clean-xtext-lib-complete"
+        onlyNewProjects="true"
         refresh="true"
         clean="true"/>
     <stream


cdietrich (commented):
unfortunately i cannot test this only any system if there are bad side effect

cdietrich (approved):

LorenzoBettini:
> unfortunately i cannot test this only any system if there are bad side effect

What I verified on Windows is that

- on the first Oomph run, all projects are effectively clean-built
- when restarting Eclipse the clean build does not happen, since everything is seen as up-to-date

Since I had no problems in Linux and macOS, I don't think this will cause any side effect.

As @merks suggested in https://github.com/eclipse/xtext/issues/2778#issuecomment-2071963236 this should be the correct setting

LorenzoBettini:
@merks on a second thought, I guess the problem fixed by this PR wasn't happening on macOS and Linux because in those machines I have this user setting:

image

can it be?

merks:
Yes.

PR #2784, opened 2023-08-21
New method `AddParameter` with index in `MutableExecutableDeclaration`

dvojtise:
This contributes a new API method in `MutableExecutableDeclaration` (and `JvmExecutableDeclarationImpl`) for active annotation processing.

It allows adding a parameter to a method declaration by specifying the position via an index.

The PR includes a test suite illustrating a minimal use case. It checks inserting at the beginning, in the middle of the parameter list and also checks that it raises an IndexOutOfBoundsException if the index is invalid.

This fixes https://github.com/eclipse/xtext/issues/2777

PR #2995, opened 2024-04-14
Work in progress

szarnekow:

PR #3022, opened 2024-04-29
Regenerated Xtend files

LorenzoBettini:
After installing the new Xtend in Eclipse I get a few changed files.

While this https://github.com/eclipse/xtext/commit/341050e055a9054582e74efea4b2b0da781c2dda comes surely from https://github.com/eclipse/xtext/pull/2993 (cc @HannesWell) I'm not quite sure about https://github.com/eclipse/xtext/commit/bc51517d56ee56516c095a097875185c8f35cfc5. Can it be due to https://github.com/eclipse/xtext/pull/2784? (cc @dvojtise)

### DIFF
--- a/org.eclipse.xtend.core/xtend-gen/org/eclipse/xtend/core/macro/TransformationContextImpl.java
+++ b/org.eclipse.xtend.core/xtend-gen/org/eclipse/xtend/core/macro/TransformationContextImpl.java
@@ -106,8 +106,8 @@ public class TransformationContextImpl implements TransformationContext {
     return this.getProblemSupport().getProblems(element);
   }
 
-  public void validateLater(final Procedure0 validationCallback) {
-    this.getProblemSupport().validateLater(validationCallback);
+  public void validateLater(final Procedure0 arg0) {
+    this.getProblemSupport().validateLater(arg0);
   }
 
   public MutableAnnotationTypeDeclaration findAnnotationType(final String qualifiedName) {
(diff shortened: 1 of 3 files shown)


org.eclipse.xtend.core/xtend-gen/org/eclipse/xtend/core/macro/TransformationContextImpl.java
```diff
@@ -106,8 +106,8 @@ public List<? extends Problem> getProblems(final Element element) {
     return this.getProblemSupport().getProblems(element);
   }
 
-  public void validateLater(final Procedure0 validationCallback) {
-    this.getProblemSupport().validateLater(validationCallback);
+  public void validateLater(final Procedure0 arg0) {
```
cdietrich, line 109: hmmm why does this now toggle back to arg0

LorenzoBettini: By the way, when working on another branch where I'm not yet rebased on master, where https://github.com/eclipse/xtext/pull/2784 had already been merged, I don't see such changes related to parameter's names; that's why I assume they are due to https://github.com/eclipse/xtext/pull/2784

Note that in this case, not all parameters are affected.

cdietrich (approved):

org.eclipse.xtend.core/xtend-gen/org/eclipse/xtend/core/macro/TransformationContextImpl.java
szarnekow: I’ve this problem since a while

LorenzoBettini: @szarnekow even before https://github.com/eclipse/xtext/pull/2784 and when automatic build in Eclipse?

cdietrich: in the past we checked in the eclipse built variant and not the maven one
but the maven one had the arg0s

cdietrich: maybe a `-parameters` or something the like got lost in the bytecode compilation?

cdietrich: grafik

LorenzoBettini: > in the past we checked in the eclipse built variant and not the maven one but the maven one had the arg0s

@cdietrich that's the eclipse built variant, that's why it's strange. Maybe it's better to revert those two files?

Moreover, only a few parameters miss the proper name.

cdietrich: yes even in the same file.

cdietrich: this seems to affect nested interfaces. maybe something broke in jdt.

cdietrich: @szarnekow which class is used to produce the ParameterDeclarations in this case?

szarnekow: I was hoping to fix it with https://github.com/eclipse/xtext/pull/2995

But I only have time to continue that work on Saturday or Sunday this week.

PR #3024, opened 2024-04-30
[#3007] Content Assistant test infrastructure improvements.

### LINKED ISSUE #3007: Extend the Content Assistant test infrastructure to test proposals from several resources.
Follow up issue of https://github.com/eclipse/xtext-eclipse/issues/1093 as requested by @cdietrich .

miklossy:
- Extend the Content Assistant test infrastructure to test proposals from several resources: modify the AbstractContentAssistTest getResourceFor method to ensure that the test file points to a file in the test project.
- Uplift the testContentAssistant method from the project example classes into the AbstractContentAssistTest and rename it to assertContentAssistant.
- Add the events_from_another_file() test case to the StatemachineContentAssistTest test cases to demonstrate how to test the content assistant with several resources.
- Also add the ContentAssistWithSeveralResourcesTest test cases that are continuously executed by the CI build.

Closes #3007

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2013, 2020 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2013, 2024 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -10,10 +10,12 @@ package org.eclipse.xtext.ui.testing;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
+import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.NullProgressMonitor;
@@ -26,8 +28,11 @@ import org.eclipse.xtext.common.types.access.jdt.JdtTypeProviderFactory;
 import org.eclipse.xtext.resource.FileExtensionProvider;
 import org.eclipse.xtext.resource.XtextResource;
 import org.eclipse.xtext.resource.XtextResourceSet;
+import org.eclipse.xtext.ui.XtextProjectHelper;
+import org.eclipse.xtext.ui.testing.util.IResourcesSetupUtil;
 import org.eclipse.xtext.ui.testing.util.JavaProjectSetupUtil;
 import org.eclipse.xtext.ui.testing.util.ResourceLoadHelper;
+import org.eclipse.xtext.util.Strings;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.jupiter.api.AfterAll;
@@ -53,11 +58,18 @@ public abstract class AbstractContentAssistTest implements ResourceLoadHelper, I
 	private Injector injector;
 
 	private static IJavaProject javaProject;
+	
+	/**
+	 * cursor position marker
+	 * @since 2.35 
+	 */
+	protected String c = "<|>";
 
 	@BeforeClass
 	@BeforeAll
 	public static void setUp() throws CoreException {
 		javaProject = JavaProjectSetupUtil.createJavaProject("contentAssistTest");
+		IResourcesSetupUtil.addNature(javaProject.getProject(), XtextProjectHelper.NATURE_ID);
 	}
 
 	@AfterClass
@@ -72,13 +84,33 @@ public abstract class AbstractContentAssistTest implements ResourceLoadHelper, I
 		XtextResourceSet resourceSet = resourceSetProvider.get();
 		initializeTypeProvider(resourceSet);
 		try {
-			URI resourceUri = URI.createURI("Test." + fileExtensionProvider.getPrimaryFileExtension());
-			Resource resource = resourceSet.createResource(resourceUri);
+			String projectFullPath = javaProject.getProject().getFullPath().toString();
+			URI resourceUri = URI.createPlatformResourceURI(projectFullPath + "/" + getProjectRelativePath() + "/" + "Test." + fileExtensionProvider.getPrimaryFileExtension(), true);
+
+			/*
+			 * Avoid the following java.lang.IllegalStateException:
+			 * A different resource with the URI 'platform:/resource/contentAssistTest/Test....' was already registered.
+			 */
+			Resource resource = resourceSet.getResource(resourceUri, false);
+			if (resource == null) {
+				resource = resourceSet.createResource(resourceUri);
+			}
 			resource.load(stream, null);
 			return (XtextResource) resource;
 		} catch (IOException e) {
 			throw new RuntimeException(e);
 		}
+	}
+
+	/**
+	 * Returns the project relative path where the files should be created.
+	 * The default implementation creates them directly in the src folder.
+	 * Clients may override.
+	 *
+	 * @since 2.35 
+	 */
+	protected String getProjectRelativePath() {
+		return "src";
 	}
 
 	@Override
@@ -104,4 +136,135 @@ public abstract class AbstractContentAssistTest implements ResourceLoadHelper, I
 		resourceSet.setClasspathURIContext(getJavaProject(resourceSet));
 	}
 
+	/**
+	 * Creates a dsl file.
+	 * 
+	 * @param fileName
+	 * 		The name of the file (without the file extension).
+	 * 		It is also possible to customize the project relative path, see {@link #getProjectRelativePath()}.
+	 * 		To use workspace relative path, use {@link IResourcesSetupUtil#createFile(String, String)} instead.
+	 * 
+	 * @param content
+	 * 		The content of the file.
+	 * 
+	 * @since 2.35
+	 */
+	protected IFile createDslFile(String fileName, CharSequence content) {
+		return createDslFile(fileName, fileExtensionProvider.getPrimaryFileExtension(), content);
+	}
+
+	/**
+	 * Creates a dsl file.
+	 * 
+	 * @param fileName
+	 * 		The name of the file (without the file extension).
+	 * 		It is also possible to customize the project relative path, see {@link #getProjectRelativePath()}.
+	 * 		To use workspace relative path, use {@link IResourcesSetupUtil#createFile(String, String)} instead.
+	 * 
+	 * @param fileExtension
+	 * 		The extension of the file.
+	 * 
+	 * @param content
+	 * 		The content of the file.
+	 * 
+	 * @since 2.35
+	 */
+	protected IFile createDslFile(String fileName, String fileExtension, CharSequence content) {
+		try {
+			return IResourcesSetupUtil.createFile(javaProject.getElementName(), getProjectRelativePath() + "/" + fileName, fileExtension, content.toString());
+		} catch (InvocationTargetException | CoreException | InterruptedException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	/**
+	 * Verifies whether the content assistant provides the expected proposals on the given cursor position.
+	 * 
+	 * @param text
+	 * 		The editor's input text. The text must contain the {@link #c} special symbol indicating the current cursor position.
+	 * 
+	 * @param expectedProposals
+	 * 		The proposals (separated by new lines) that are expected to be offered whenever the content assistant is triggered on the given cursor position.
+	 * 
+	 * @since 2.35
+	 */
+	protected void assertContentAssistant(CharSequence text, CharSequence expectedProposals) {
+		assertContentAssistant(text, expectedProposals, null, null);
+	}
+	
+	/**
+	 * Verifies whether the content assistant provides the expected proposals on the given cursor position.
+	 * 
+	 * @param text
+	 * 		The editor's input text. The text must contain the {@link #c} special symbol indicating the current cursor position.
+	 * 
+	 * @param expectedProposals
+	 * 		The proposals that are expected to be offered whenever the content assistant is triggered on the given cursor position.
+	 * 
+	 * @since 2.35
+	 */
+	protected void assertContentAssistant(CharSequence text, String[] expectedProposals) {
+		assertContentAssistant(text, expectedProposals, null, null);
+	}
+
+	/**
+	 * Verifies whether the content assistant provides the expected proposals on the given cursor position.
+	 * Furthermore, it applies the proposalToApply - if it is given - and verifies the expected content afterwards.
+	 * 
+	 * @param text
+	 * 		The editor's input text. The text must contain the {@link #c} special symbol indicating the current cursor position.
+	 * 
+	 * @param expectedProposals
+	 * 		The proposals (separated by new lines) that are expected to be offered whenever the content assistant is triggered on the given cursor position. 
+	 * 
+	 * @param proposalToApply
+	 * 		The proposal to apply from the list of the expected proposals.
+	 * 
+	 * @param expectedContent
+	 * 		The expected editor content after the given proposal has been applied.
+	 * 
+	 * @since 2.35
+	 */
+	protected void assertContentAssistant(CharSequence text, CharSequence expectedProposals, String proposalToApply, String expectedContent) {
+		String[] expectedProposalsArray = Strings.toUnixLineSeparator(expectedProposals.toString()).split("\n");
+		assertContentAssistant(text, expectedProposalsArray, proposalToApply, expectedContent);
+	}
+	
+	/**
+	 * Verifies whether the content assistant provides the expected proposals on the given cursor position.
+	 * Furthermore, it applies the proposalToApply - if it is given - and verifies the expected content afterwards.
+	 * 
+	 * @param text
+	 * 		The editor's input text. The text must contain the {@link #c} special symbol indicating the current cursor position.
+	 * 
+	 * @param expectedProposals
+	 * 		The proposals that are expected to be offered whenever the content assistant is triggered on the given cursor position. 
+	 * 
+	 * @param proposalToApply
+	 * 		The proposal to apply from the list of the expected proposals.
+	 * 
+	 * @param expectedContent
+	 * 		The expected editor content after the given proposal has been applied.
+	 * 
+	 * @since 2.35
+	 */
+	protected void assertContentAssistant(CharSequence text, String[] expectedProposals, String proposalToApply, String expectedContent) {
+
+		int cursorPosition = text.toString().indexOf(c);
+		if (cursorPosition == -1) {
+			throw new RuntimeException("Can't locate cursor position symbols '" + c + "' in the input text.");
+		}
+
+		String content = text.toString().replace(c, "");
+
+		try {
+			ContentAssistProcessorTestBuilder builder = newBuilder().append(content).assertTextAtCursorPosition(cursorPosition, expectedProposals);
+	
+			if (proposalToApply != null) {
+				builder.applyProposal(cursorPosition, proposalToApply).expectContent(expectedContent);
+			}
+		} catch (Exception exception) {
+			throw new RuntimeException(exception);
+		}
+	}
 }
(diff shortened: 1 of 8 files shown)


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
```diff
@@ -72,9 +83,18 @@ public XtextResource getResourceFor(InputStream stream) {
...
-			resource.load(stream, null);
+			String projectFullPath = javaProject.getProject().getFullPath().toString();
+			URI resourceUri = URI.createPlatformResourceURI(projectFullPath + "/" + "Test." + fileExtensionProvider.getPrimaryFileExtension(), true);
+			
+			/*
+			 * Avoid the following java.lang.IllegalStateException:
+			 * A different resource with the URI 'platform:/resource/contentAssistTest/Test....' was already registered.
+			 */
+			Resource resource = resourceSet.getResource(resourceUri, false);
+			if (resource == null) {
```
LorenzoBettini, line 94: But if the resource is not null, i.e., it exists, the passed input stream is not used to populate it, is it?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
```diff
@@ -104,4 +124,93 @@ protected void initializeTypeProvider(XtextResourceSet resourceSet) {
...
+	 * 		The extension of the file.
+	 * 
+	 * @param content
+	 * 		The content of the file.
+	 * 
+	 * @since 2.35
+	 */
+	protected IFile createDslFile(String fileName, String fileExtension, CharSequence content) {
+		try {
+			return IResourcesSetupUtil.createFile(javaProject.getElementName(), fileName, fileExtension, content.toString());
```
LorenzoBettini, line 158: it would be useful to specify also the path of the file, e.g., into the "src" directory. Does this allow that? If so, maybe it should be specified in the documentation.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
```diff
@@ -104,4 +124,93 @@ protected void initializeTypeProvider(XtextResourceSet resourceSet) {
...
+	 * 
+	 * @param text
+	 * 		The editor's input text. The text must contain the {@link #c} special symbol indicating the current cursor position.
+	 * 
+	 * @param expectedProposals
+	 * 		The list of proposals that are expected to be offered whenever the content assistant is triggered on the given cursor position.
+	 * 
+	 * @since 2.35
+	 */
+	protected void assertContentAssistant(CharSequence text, List<String> expectedProposals) {
```
LorenzoBettini, line 175: Since we're creating a new API, could we take the chance to allow for expected proposals to be specified also as a string where proposals are separated by a new line? This would allow for easier TDD or test adjustments: it's easier to specify or update a string than to create a list of proposals or update it. Turning the string into a list or array of strings is easy with Java streams. @szarnekow wdyt?

org.eclipse.xtext.xtext.ui.examples/projects/fowlerdsl/org.eclipse.xtext.example.fowlerdsl.ui.tests/src/org/eclipse/xtext/example/fowlerdsl/ui/tests/StatemachineContentAssistTest.xtend
```diff
@@ -8,27 +8,25 @@
...
 class StatemachineContentAssistTest extends AbstractContentAssistTest {
 
-	// cursor position marker
-	val c = '''<|>'''
-
 	@Test def empty() throws Exception {
 		'''
 			«c»
-		'''.testContentAssistant(#[
+		'''.assertContentAssistant(#[
```
LorenzoBettini, line 29: For example, with the additional API proposed above, here you could do

```
'''
commands
events
resetEvents
'''
```

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
```diff
@@ -72,9 +83,18 @@ public XtextResource getResourceFor(InputStream stream) {
 		XtextResourceSet resourceSet = resourceSetProvider.get();
 		initializeTypeProvider(resourceSet);
 		try {
-			URI resourceUri = URI.createURI("Test." + fileExtensionProvider.getPrimaryFileExtension());
-			Resource resource = resourceSet.createResource(resourceUri);
-			resource.load(stream, null);
+			String projectFullPath = javaProject.getProject().getFullPath().toString();
+			URI resourceUri = URI.createPlatformResourceURI(projectFullPath + "/" + "Test." + fileExtensionProvider.getPrimaryFileExtension(), true);
```
cdietrich, line 87: am not sure. dont in java projects. model files need to reside in source folder?

LorenzoBettini: Exactly! That's why I was suggesting below to allow for relative path specification.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
miklossy: Yes. Exactly.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
miklossy: To specity the worskpace relative path, the `IResourcesSetupUtil#createFile(String, String)` method can be used. I added this hint to the documentation.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
LorenzoBettini: @miklossy I still don't understand this part... see my comment and @cdietrich's one

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
LorenzoBettini: So I still don't understand... I was expecting the specified input stream to be taken into consideration and in case replace the current resource... maybe I'm still missing something

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
LorenzoBettini: So, I don't understand the benefit of this new method... again, I'm probably missing something, but why not simply consider the path in the first place?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
```diff
@@ -104,4 +124,132 @@ protected void initializeTypeProvider(XtextResourceSet resourceSet) {
...
+	 * @param text
+	 * 		The editor's input text. The text must contain the {@link #c} special symbol indicating the current cursor position.
+	 * 
+	 * @param expectedProposals
+	 * 		The proposals (separated by new lines) that are expected to be offered whenever the content assistant is triggered on the given cursor position.
+	 * 
+	 * @since 2.35
+	 */
+	protected void assertContentAssistant(CharSequence text, CharSequence expectedProposals) {
+		assertContentAssistant(text, expectedProposals.toString().split(System.lineSeparator()), null, null);
```
LorenzoBettini, line 178: Maybe I opened a Pandora's box... O:)
I'd like to avoid introducing further problems due to d**n line endings in Windows.
Your implementation works when using Xtend template strings but would fail when using Java text blocks or Xtend multi-line strings that are now (https://github.com/eclipse/xtext/pull/3004) normalized w.r.t. line endings (only Unix EOL is generated).

I'd like this solution to be guarded against different EOLs...

A possible solution would be to first normalize the `expectedProposals` with only Unix EOL (we have a utility method `Strings.toUnixLineSeparator`) and then split it only according "\n".

I haven't tried that, but it should work.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
```diff
@@ -104,4 +124,132 @@ protected void initializeTypeProvider(XtextResourceSet resourceSet) {
...
+	 * 
+	 * @param text
+	 * 		The editor's input text. The text must contain the {@link #c} special symbol indicating the current cursor position.
+	 * 
+	 * @param expectedProposals
+	 * 		The proposals that are expected to be offered whenever the content assistant is triggered on the given cursor position.
+	 * 
+	 * @since 2.35
+	 */
+	protected void assertContentAssistant(CharSequence text, String[] expectedProposals) {
```
LorenzoBettini, line 192: So you removed the version with List?

That could have been useful in Xtend with `#[ ... ]`. Or would it work anyway thanks to conversion?

Moreover, if we want to use arrays, why not using `String... expectedProposals`, which would not require the creation of an array when calling this method? Would it conflict with the one accepting a String?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
```diff
@@ -104,4 +124,132 @@ protected void initializeTypeProvider(XtextResourceSet resourceSet) {
...
+	 * @param proposalToApply
+	 * 		The proposal to apply from the list of the expected proposals.
+	 * 
+	 * @param expectedContent
+	 * 		The expected editor content after the given proposal has been applied.
+	 * 
+	 * @since 2.35
+	 */
+	protected void assertContentAssistant(CharSequence text, CharSequence expectedProposals, String proposalToApply, String expectedContent) {
+		assertContentAssistant(text, expectedProposals.toString().split(System.lineSeparator()), proposalToApply, expectedContent);
```
LorenzoBettini, line 215: As above, concerning EOL in the passed string.

org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/contentassist/ContentAssistWithSeveralResourcesTest.java
```diff
@@ -0,0 +1,42 @@
...
+public class ContentAssistWithSeveralResourcesTest extends AbstractContentAssistTest {
+
+	@Test public void test_with_one_resource() {
+		assertContentAssistant("type A extends " + c, new String[]{("A")});
+	}
+	
+	@Test public void test_with_two_resources() {
+		createDslFile("types", "type A extends A");
+		IResourcesSetupUtil.waitForBuild();
+		assertContentAssistant("import \"types.importuriuitestlanguage\" type B extends " + c, new String[]{"A", "B"});
```
LorenzoBettini, line 33: Here, I would call it with "A\nB" to make sure my proposal above about EOLs works as expected.

org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.ui.tests/src/org/eclipse/xtext/example/domainmodel/ui/tests/ContentAssistTest.xtend
```diff
@@ -56,7 +52,7 @@ class ContentAssistTest extends AbstractContentAssistTest {
 			entity E {
 				«c»
 			}
-		'''.testContentAssistant(#[
+		'''.assertContentAssistant(#[
```
LorenzoBettini, line 55: OK, so, concerning my comment above about `String[]` instead of `List`, this seems to work out of the box with Xtend, right?

LorenzoBettini (changes requested):

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
miklossy: @LorenzoBettini : The method `getResourceFor(InputStream stream)` will be called twice within each test case execution, in both cases the passed stream is created from the `currentModelToParse` variable. So basically it does not matter if we use the first call and ingore the second call or if we create the resource on the first call and overwrite it with the same content in the second call.

I modified the PR so that in the second call the current resource content will be replaced, as requested.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
miklossy: This `createDslFile` method is basically a shorter way of calling the `IResourcesSetupUtil.createFile` method. The user can call this method if he/she wants to pass less parameter and does not want to bother with the exception handling. But if that does not fullfill his/her usecase, than he/she can call directly the `IResourcesSetupUtil.createFile` method instead.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
miklossy: @LorenzoBettini : I have just modified the PR as requested:

```
	protected void assertContentAssistant(CharSequence text, CharSequence expectedProposals) {
		String[] expectedProposalsArray = Strings.toUnixLineSeparator(expectedProposals.toString()).split("\n");
		assertContentAssistant(text, expectedProposalsArray, null, null);
	}

	protected void assertContentAssistant(CharSequence text, CharSequence expectedProposals, String proposalToApply, String expectedContent) {
		String[] expectedProposalsArray = Strings.toUnixLineSeparator(expectedProposals.toString()).split("\n");
		assertContentAssistant(text, expectedProposalsArray, proposalToApply, expectedContent);
	}
```

org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/contentassist/ContentAssistWithSeveralResourcesTest.java
miklossy: I have just modified the PR as requested.

org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.ui.tests/src/org/eclipse/xtext/example/domainmodel/ui/tests/ContentAssistTest.xtend
miklossy: Yes, Xtend automatically converts the actual parameter from array to list (and vice versa) if the formal parameter is a list/array.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
LorenzoBettini: @miklossy but the question now is why is it called twice?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
LorenzoBettini: OK, but in any case, can the passed fileName consist of a path? I think that's important. I mean, it's important to be able to specify a path, otherwise that method is unusable for Xbase languages where files must be in source folders.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
LorenzoBettini: @miklossy Thanks! I feel safer now ;)

Just in case, can't the first method be implemented in terms of the second one, by simply passing `text, expectedProposals, null, null`?

LorenzoBettini:
Looks like GitHub Actions is having bad days lately... the ubuntu build failed with a timeout.

Let's wait for the build to finish and then we re-run the failing jobs, just in case.

LorenzoBettini (commented):
@miklossy I'm fine with the PR, but there's still my question pending about the path in the passed fileName

miklossy:
> @miklossy I'm fine with the PR, but there's still my question pending about the path in the passed fileName

It is not possible to use absolute path there, but you can use relative path, something like:
`src/foo/bar/testfile`

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
miklossy: It is a good question, but currently I am not sure if we could fix that without any breaking change. Let's analyse it in a follow up issue :-)

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
miklossy: @LorenzoBettini : I extended the javadoc as follows:

```
@param fileName
	 * 		The name of the file (without the file extension).
	 * 		It is also possible to use project relative path something like: <code>src/foo/bar/Test</code>
	 * 	 	To be able to specify the workspace relative path, use {@link IResourcesSetupUtil#createFile(String, String)} instead.
	 * 
```

LorenzoBettini:
> > @miklossy I'm fine with the PR, but there's still my question pending about the path in the passed fileName
> 
> It is not possible to use absolute path there, but you can use relative path, something like: `src/foo/bar/testfile`

No no, I meant project relative path.
I see you have updated the documentation, so to me that's fine! :)

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
LorenzoBettini: > It is a good question, but currently I am not sure if we could fix that without any breaking change. Let's analyse it in a follow up issue :-)

OK, if it's always been like that :)

LorenzoBettini (approved):
Thanks!

@cdietrich wdyt?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
cdietrich: from my pov this is still open

LorenzoBettini: @cdietrich relative paths are allowed, see the other answers. What did you mean?

cdietrich: i mean the default. unfortunately i dont have cappa to try it

cdietrich (approved):

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractContentAssistTest.java
LorenzoBettini: Ok, now maybe I see: the file on which the content assistant is triggered! That's always the same and cannot be created in a specific folder. Is that right @miklossy ? If it's so, then we should allow for some customization? Maybe a variable that can be changed? And maybe the default should be "src/Test"?

LorenzoBettini: @miklossy Thanks!
@cdietrich I think this also fixes your last issue.

miklossy:
@LorenzoBettini : do you see why the build is red?

LorenzoBettini:
> @LorenzoBettini : do you see why the build is red?

macOS on GitHub Actions is completely unreliable and slow lately.
I'll try to re-run. If it still fails, I'll still merge this.

miklossy:
Hmm.. The MacOS build is still red, but the windows and the ubuntu builds are green. Could we merge it, @LorenzoBettini ?

LorenzoBettini:
Done! Thanks @miklossy .
Do you think it's worth to say something in the release notes?

miklossy:
Yes, see https://github.com/eclipse/xtext-website/pull/43 :-)

Thanks @LorenzoBettini for your entire support on this PR!

PR #3027, opened 2024-05-02
assertNoErrors/Issues in CompilationTestHelper.Result

LorenzoBettini:
I often check for errors/issues when using CompilationTestHelper, and I always have to retrieve errors/warnings, do the filtering, and check for emptiness.

I thought adding such methods to `Result` would be worthwhile.

`assertNoWarnings` doesn't seem necessary to me, but I can add it in case.


LorenzoBettini:
@szarnekow there's also this proposal

szarnekow (approved):

PR #3031, opened 2024-05-06
Provide a JvmGenericTypeValidator

### LINKED ISSUE #2349: Provide a JvmGenericTypeValidator

LorenzoBettini:
Closes #2349 

This is my proposal for a `JvmGenericTypeValidator`, containing code extracted and refactored from XtendValidator for typical inheritance-based checks on the inferred Jvm model for Xbase languages. This is an opt-in feature, to be enabled through a composed validator.

This has been tested through Xtend, which is now using it, and a testlanguage for some corner cases (https://github.com/eclipse/xtext/compare/main...LorenzoBettini:xtext:lb_2349?expand=1#diff-25aca6d3c766f0d3e06a3fd1d893c1d3e3dc23b7e503b3facabffbd1e3e882f9).

The Xtend tests have been updated when I thought they would be improved according to the introduced JvmGenericTypeValidator. As for the rest, the results are the same. Note that Xtend customizes such a validator for a few special cases.

This validator has been applied to Domainmodel https://github.com/eclipse/xtext/commit/ef578b5fcfae18b37e63a163668562a3b0b89fb4 and the corresponding Entities web example https://github.com/eclipse/xtext/commit/08b5090f3b514a3d83eebe83f1fdc25af017048c

@szarnekow please let me know if I can provide further details to help the review.


LorenzoBettini:
@szarnekow I also added the commented `composedCheck` in the generated MWE2

szarnekow (approved):

szarnekow:
This is indeed an appreciated improvement for standard Xbase languages. Well done!

PR #3034, opened 2024-05-12
Use the Oomph setup configuration is the contribution documentation

merks:


merks:
See https://github.com/eclipse/xtext/pull/3030#issuecomment-2106167236 for details.

Note that I have tested creating a new from-scratch development environment.   The Git clone is not dirty, so I removed that comment too.

LorenzoBettini (approved):
Thank you @merks

Issue #3001, opened 2024-04-17
Elegant way to define additional or overriding Guice bindings in DSL tests


HannesWell:
For the test plugin of our custom DSL I want to define extra bindings for the Guice-Injector used to inject objects in the test.

Since the usual setup in DSL test classes using JUnit-5 is something 
```
@ExtendWith(InjectionExtension.class)
@InjectWith(MyDSLInjectorProvider.class)
public abstract class DC43BaseApplyTest {
  <the test code>
}
```
The approach to achieve that is described in [this SO question](https://stackoverflow.com/questions/40544719/how-to-use-different-object-injections-in-xtext-tests-than-in-productive-environ) and basically suggest to extend the generated `MyDSLInjectorProvider` by overriding its `createRuntimeModule()` method and return a customized Module that adds the desired bindings. The extended provider can then be used in the test classes.

This all works well (AFAICT) but it is a bit inconvenient because by default the `InjectorProvider` is generated as follows:
```
public class MyDSLInjectorProvider implements IInjectorProvider, IRegistryConfigurator {

...
	protected MyDSLRuntimeModule createRuntimeModule() {
		// make it work also with Maven/Tycho and OSGI
		// see https://bugs.eclipse.org/bugs/show_bug.cgi?id=493672
		return new MyDSLRuntimeModule() {
			@Override
			public ClassLoader bindClassLoaderToInstance() {
				return MyDSLInjectorProvider.class
						.getClassLoader();
			}
		};
	}
...
}
```

Because the generated `createRuntimeModule()` returns the specific module type generated for the DSL a sub-classes that want to override that method has to return the specific type as well and consequently has to repeat the ceremony to make the injector work in Maven/Tycho and OSGI or one has to override `internalCreateInjector()` as it is done in
https://github.com/eclipse/xtext/blob/73bd1580bcea48d7ed27f990cadbc5ae4d308572/org.eclipse.xtext.testlanguages.ide/src/org/eclipse/xtext/testlanguages/xtextgrammar/ide/XtextGrammarTestLanguageIdeInjectorProvider.java#L24-L35

But this also includes repeating ceremony.

If `createRuntimeModule()` would declare its return type to be just `com.google.inject.Module` then a customized provider could be just
```
public class MyDSLExtendedInjectorProvider extends MyDSLInjectorProvider {
	@Override
	protected Module createRuntimeModule() {
		return Modules.override(super.createRuntimeModule()).with(binder -> {
[...]

PR #3042, opened 2024-05-16
Make the generated testing InjectorProvider more flexible

LorenzoBettini:
Part of #3001 

As hinted here https://github.com/eclipse/xtext/issues/3001#issuecomment-2067658136 instead of generating an injector provider of this shape:

```java
protected MyDslRuntimeModule createRuntimeModule() {
	return new MyDslRuntimeModule() {
			@Override
			public ClassLoader bindClassLoaderToInstance() {
				return MyDslInjectorProvider.class
						.getClassLoader();
			}
	};
}
```

This PR generates it like that (main commit https://github.com/eclipse/xtext/commit/1181731293c910ae29a5f2af41624a01d9009659):

```java
protected MyDslRuntimeModule createRuntimeModule() {
	return new MyDslRuntimeModule() {
			@Override
			public ClassLoader bindClassLoaderToInstance() {
				return MyDslInjectorProvider.this.getClass()
						.getClassLoader();
			}
	};
}
```

This section was introduced in the past for running tests in OSGI, e.g., by Maven Tycho, to get the correct class loader.

This new version makes the class loader be retrieved from the instance of the injector provider, NOT from the static class of the injector provider, as before.

Thus, simply inheriting an injector provider in another test bundle will make Guice take the correct class loader.

Instead, before, one had also to redefine the `createRuntimeModule`.

This makes things easier in testing. Some examples of simplification due to this change are:

- https://github.com/eclipse/xtext/commit/a85207c6cdda23656d7e18f48bd3621c144c24dd
- https://github.com/eclipse/xtext/commit/7aeee04e5b669203cbc6980c327b719bbb4c7f0f

I regenerated all our languages accordingly.

Concerning purexbase, I cleaned up things a bit https://github.com/eclipse/xtext/commit/8230adad0defe72c5d30ae30e0036a9fefe20e11 : the old `RuntimeInjectorProvider` was already useless, maybe something from the past.

The PR can be seen as a first step/preparation in case we want to address #3001


szarnekow (approved):

szarnekow:
I think I like this. Small change and nice improvement.

cdietrich:
find . -name "*InjectorProvider.java" | grep -v /ide/ | grep -v UiInjectorProvider | wc -l
     108
     
     did you double check if you really regenerated all files?

LorenzoBettini:
> find . -name "*InjectorProvider.java" | grep -v /ide/ | grep -v UiInjectorProvider | wc -l 108
> 
> ```
>  did you double check if you really regenerated all files?
> ```

@cdietrich With "Quick Search" I searched for the string `show_bug.cgi?id=493672` and regenerated the languages in each project with such an occurrence, so I should have caught them all. I haven't touched the injector providers that were not generated, but manually crafted, like XbaseInjectorProvider and the one of Xtend.

cdietrich:
ok

PR #3047, opened 2024-05-17
only validate once per primary source element

### LINKED ISSUE #3045: Compilation Error in `xtend-annotation-examples-client`
When installing the new Xtend in Eclipse, with this https://github.com/eclipse/xtext/pull/3031, I get a compilation error in `extract.ExtractExample`:

image

If I change it to `def`, as also suggested by the quickfix, I get another one:

image

So I guess `JvmGenericTypeValidator` gets confused with active annotations?

The failure is also visible in the CI build of https://github.com/xtext/xtext-reference-projects/pull/421

cc @szarnekow

LorenzoBettini:
Closes #3045


LorenzoBettini:
I also recreated a corner case with the specific test language for JvmGenericTypeValidator https://github.com/eclipse/xtext/pull/3047/commits/acacaaa6850aaab699262b60a63f459ecad7e35b

@szarnekow I would merge this to fix this regression, OK?

PR #3037, opened 2024-05-14
Support for Java 21

LorenzoBettini:
Closes https://github.com/eclipse/xtext/issues/2823

Addresses most of https://github.com/eclipse/xtext/issues/2686

Issue #2823, opened 2023-10-18
Does Xtend work with java records?


tripleo1:
I am testing Xtend with a Java 17 project and it complains about record classes:

```
Starting a Gradle Daemon (subsequent builds will be faster)

> Task :project:generateXtext FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':project:generateXtext'.
> Cannot handle type public @SuppressWarnings("unused") record GRecord(Parm1 parm1, Parm2 parm2, Supplier<Dep2> arg3) {
```

```
plugins { id "org.xtext.xtend" version "4.0.0" }
```

```
java {
    sourceCompatibility = JavaVersion.VERSION_17
    targetCompatibility = JavaVersion.VERSION_17
}
```

cdietrich:
hi, can you please provide more information what you are actually doing?

tripleo1:
> hi, can you please provide more information what you are actually doing?

I just dropped the gradle plugin into an existing java project, and am trying to get a clean build.

On a side note, I don't really like records: Should I just convert the records to Xtend?

cdietrich:
there is no record in xtend. there is also no in java source lang.
so the gradle plugin would be able to parse the java files i guess

tripleo1:
> ```
> Caused by: java.lang.IllegalArgumentException: Cannot handle type public record MyRecord(String a, int b) {
> /* Implicit */  private final String a;
> /* Implicit */  private final int b;
>   public MyRecord(String a, int b) {
>     .a = a;
>     .b = b;
>   }
> }
>         at org.eclipse.xtext.java.resource.JavaDerivedStateComputer.createType(JavaDerivedStateComputer.java:130)
>         at org.eclipse.xtext.java.resource.JavaDerivedStateComputer.installStubs(JavaDerivedStateComputer.java:107)
>         at org.eclipse.xtext.java.resource.JavaResource.lambda$0(JavaResource.java:174)
>         at org.eclipse.xtext.java.resource.JavaResource.initializing(JavaResource.java:190)
>         at org.eclipse.xtext.java.resource.JavaResource.installStubs(JavaResource.java:173)
>         at org.eclipse.xtext.java.resource.JavaResourceDescriptionManager.getResourceDescription(JavaResourceDescriptionManager.java:48)
> ```

1. I am assuming this is the same problem that I have...
2. it could be said then that Xtend does not support Java records?
3. What does this even mean?

> there is no record in xtend. 

[...]

PR #3048, opened 2024-05-17
fix java source lang record handling

cdietrich:


org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.java
```diff
@@ -126,6 +126,9 @@ public JvmDeclaredType createType(TypeDeclaration type, String packageName) {
 		case TypeDeclaration.ANNOTATION_TYPE_DECL:
 			jvmType = TypesFactory.eINSTANCE.createJvmAnnotationType();
 			break;
+		case 5 /*TypeDeclaration.RECORD_DECL*/:
```
cdietrich, line 129: see also discussion in

https://github.com/eclipse/xtext/issues/2823
and 
https://github.com/eclipse/xtext/pull/3037

cdietrich: should we use the preview constant? it seems to be in 33 too

org.eclipse.xtext.java.tests/src/org/eclipse/xtext/java/tests/JavaSourceLanguageTest.java
```diff
@@ -217,9 +213,19 @@ public void testJava21Record() {
 
 	@Inject
 	private IJvmTypeProvider.Factory typeProviderFactory;
-
+
```
LorenzoBettini, line 216: Nip: line with only tab

org.eclipse.xtext.java.tests/src/org/eclipse/xtext/java/tests/JavaSourceLanguageTest.java
```diff
@@ -217,9 +213,19 @@ public void testJava21Record() {
...
+	
 	protected XtextResourceSet resourceSet(Map<String, String> files) {
+		return this.resourceSet(files, null);
+	}
+
+	protected XtextResourceSet resourceSet(Map<String, String> files, JavaVersion javaVersion) {
 		XtextResourceSet result = resourceSetProvider.get();
+		if (javaVersion != null) {
+			JavaConfig javaConfig = new JavaConfig();
+			javaConfig.setJavaSourceLevel(JavaVersion.JAVA21);
```
LorenzoBettini, line 225: I guess here you meant to pass `javaVersion`, the parameter, not the hardcoded version

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.java
LorenzoBettini: I would use `TypeDeclaration.RECORD_DECL`: we already use other internal constants; it should not be a preview feature anymore in Java 21; maybe they forgot to remove the comment? Or maybe they refer to previous Java versions.

I don't understand the other comment. You're not talking about specifically about records, but in general about the `default` case? Maybe it's better to default to generic type as we do in JdtBasedTypeFactory, though I don't understand the part about super type.

LorenzoBettini (commented):
In general, I'd like to have a test that would fail without this change. Instead, `org.eclipse.xtext.java.tests.JavaSourceLanguageTest.testJava21Record()` still succeeds without this change.

For example, in the current shape, if I comment

```java
		javaConfig.setJavaSourceLevel(JavaVersion.JAVA21);
		javaConfig.setJavaTargetLevel(JavaVersion.JAVA21);
```

the test fails; so I was misled that everything else would work by just setting the Java version.

Or did I miss something?

See also the other small changes.

cdietrich:
grafik

cdietrich:
for me it fails without change

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.java
cdietrich: the jdtbasedfactory does not throw. it falls back to jvm generic type

org.eclipse.xtext.java.tests/src/org/eclipse/xtext/java/tests/JavaSourceLanguageTest.java
cdietrich: yes

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.java
cdietrich: i am asking if we should do it here to. any idea @szarnekow

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.java
```diff
@@ -126,6 +126,9 @@ public JvmDeclaredType createType(TypeDeclaration type, String packageName) {
 		case TypeDeclaration.ANNOTATION_TYPE_DECL:
 			jvmType = TypesFactory.eINSTANCE.createJvmAnnotationType();
 			break;
+		case TypeDeclaration.RECORD_DECL:
```
cdietrich, line 129: its a final. so it will get inlined anyway. afaik

LorenzoBettini:
> for me it fails without change

@cdietrich 

Yes, I verified that in your branch.

I was referring to the previous shape of the test: without the JAVA_21 specification it was failing, with the specification it was succeeding, but it wasn't testing much as you said.

Maybe it's worthwhile to leave a comment in the `resourceSet` method in the test, or in the test class in general stressing that to correctly test something in this test you should... etc. etc.

From what you said in the previous comments also in the other issues, my mistake was not thinking about the install stub step which doesn't report errors, and then the binary class is found. Or did I misunderstand that as well? In general, it would be better to avoid such problems in the future, that's why I'd leave a comment in the test. Basically, I didn't understand how that test was supposed to work.

As for the rest, I'm fine with the PR.

org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.java
LorenzoBettini: Even better ;)

LorenzoBettini (approved):

cdietrich:
@LorenzoBettini have no longer access to a dev setup. feel free to push a comment with a commit to this brAnch

LorenzoBettini:
> @LorenzoBettini have no longer access to a dev setup. feel free to push a comment with a commit to this brAnch

Yes I can do that, but I still don't understand what was wrong with the previous version of the test.
That's why I asked you. I can put the comment if you explain what the problem was before

LorenzoBettini:
Thank you @cdietrich for adding the comment and fixing this!

LorenzoBettini:
@szarnekow could you please explain where the JavaDerivedStateComputer is used?

cdietrich:
Cause gradle uses the Java source language under the hood to get index entries for Java files in the project

LorenzoBettini:
Ok! That's why it never came into play in my tests and PRs! So Gradle doesn't use Standalone builder. 

So, even more sensible, please try the experiments I said in my comment in the reference projects issue.

cdietrich:
Yes it uses the new builder

org.eclipse.xtext.build.IncrementalBuilder

PR #3058, opened 2024-05-21
another lsp4j minor pump

cdietrich:


LorenzoBettini (approved):
It might be worthwhile to use properties in the BOM for artifacts sharing the same version, like I did for jdt

Issue #3057, opened 2024-05-21
UI tests extremely flaky on macOS in GitHub Actions


LorenzoBettini:
Some examples:

```
Failures: 
org.eclipse.xtend.ide.tests.compiler.ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef_4
  Run 1: ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef_4:295->testResolvingXFunctionTypeRef:447->assertNoErrorsInWorkspace:460->Assert.assertFalse:65->Assert.assertTrue:42->Assert.fail:89 Couldn't find the mandatory library 'org.eclipse.xtext.xbase.lib' 2.8.0 or higher on the project's classpath.
  Run 2: ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef_4:295->testResolvingXFunctionTypeRef:421->assertNoErrorsInWorkspace:460->Assert.assertFalse:65->Assert.assertTrue:42->Assert.fail:89 org.eclipse.emf.ecore.EObject cannot be resolved to a type.
  Run 3: ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef_4:295->testResolvingXFunctionTypeRef:421->assertNoErrorsInWorkspace:460->Assert.assertFalse:65->Assert.assertTrue:42->Assert.fail:89 org.eclipse.emf.ecore.EObject cannot be resolved to a type.
  Run 4: ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef_4:295->testResolvingXFunctionTypeRef:447->assertNoErrorsInWorkspace:460->Assert.assertFalse:65->Assert.assertTrue:42->Assert.fail:89 myannotation.MyAnnotation cannot be resolved to a type.

Flakes: 
org.eclipse.xtend.ide.tests.compiler.ActiveAnnotationsInSameProjectTest.testActiveAnnotationInSameProject
  Run 1: ActiveAnnotationsInSameProjectTest.testActiveAnnotationInSameProject:144->assertHasErrors:529->Assert.fail:89 Expected an error marker containing 'same project' on /test.project/src/mypack/Client.xtend but found Couldn't find the mandatory library 'org.eclipse.xtext.xbase.lib' 2.8.0 or higher on the project's classpath.
  Run 2: PASS

[...]

PR #3063, opened 2024-05-23
DetectVMInstallationsJob.disabled as system property of tycho-surefire

LorenzoBettini:
The option seems to reduce the chances that flakes fail; at least in my experiments in macOS and Windows there are still flakes but they finally succeed

For https://github.com/eclipse/xtext/issues/3057

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -695,6 +695,10 @@
 							<exclude />
 						</excludes>
 						<failIfNoTests>true</failIfNoTests>
+						<systemProperties>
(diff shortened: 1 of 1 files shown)


cdietrich (approved):

LorenzoBettini:
@cdietrich I seem to recall that in reference projects you had strange failures showing that the wrong version of Java was picked when running UI tests.

Maybe it's worthwhile to always use this option in reference projects and maybe let the wizard generate that?

cdietrich:
yes

PR #3069, opened 2024-05-27
Small documentation improvement

tom-fonda:


LorenzoBettini:
@tom-fonda thank you for the contribution, but just to make sure I don't miss anything: the script would skip tests even if you passed `-DskipTests`, wouldn't it?

LorenzoBettini (approved):

tom-fonda:
@LorenzoBettini
It would, but I found that argument a bit confusing. I had initially thought I had to manually add it to the Maven invocations inside the script (my mistake, I assumed so for a weak reason, i.e. its `-D` + camelcase structure, which I associate to Java flags and not to bash scripts flag). Then, glancing quickly at the script I immediately noticed the `--no-tests` argument, and went for it.

LorenzoBettini:
OK, thanks for the confirmation and for the contribution :)

PR #3071, opened 2024-05-28
regenerated Eclipse doc

LorenzoBettini:
There are some `<br>` removed and I don't understand why.

If that's critical, at least I can create a PR with only the changes in the TOC and Xbase part.


org.eclipse.xtext.doc/contents/310_eclipse_support.html
```diff
@@ -532,7 +532,7 @@ <h2 id="folding">Folding</h2>
 }
 </code></pre>
 
-<p>As a result, not only the <code>state</code>, but also the <code>events</code>, <code>resetEvents</code> and <code>commands</code> regions become foldable:<br />
```
cdietrich, line 535: are the changes fine without the br?
dont see these locally

LorenzoBettini: I've done that from the command line because the launch couldn't find ruby through env or something.

Maybe for the moment it's better to only keep the change adding documentation for JvmGenericTypeValidator and the chamged TOC.

I'll update the PR later and then start the release.

PR #3080, opened 2024-06-07
annotate dispatch method entry point with @Generated

### LINKED ISSUE #3077: Make Xtend generated code easier to cover
In the presence of dispatch methods Java code generated from Xtend is not easy to fully cover, due to the shape of the entry method of dispatch methods. Consider this Xtend code:

```xtend
package example

class Foo {
	def dispatch String m(String s) {
		s
	}

	def dispatch String m(Integer i) {
		i.toString
	}
}
```

And this test

```java
package example.tests;

import static org.junit.Assert.assertEquals;

import org.junit.Test;

import example.Foo;

public class FooTest {

	@Test
	public void test() {
		assertEquals("1", new Foo().m(1));
		assertEquals("1", new Foo().m("1"));
	}

}
```

running it through JaCoCo leads to uncovered lines in the entry method because of the last case:

image

Starting from JaCoCo 0.8.2, we can exclude classes and methods by annotating them with a custom annotation such that:

- The name of the annotation includes Generated.
- The retention policy of annotation is runtime or class.

We could mark the generated entry method with a custom annotation, which would be part of the Xtend library.

For example, I manually create this annotation in my project:

```java
package annotations;

import static java.lang.annotation.ElementType.*;

import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

@Retention(RetentionPolicy.RUNTIME)
@Target({TYPE, METHOD, CONSTRUCTOR})
public @interface XtendGenerated {
}
```

and manually modify the generated Xtend code as such:

```java
...
  @XtendGenerated
  public String m(final Object i) {
    if (i instanceof Integer) {
      return _m((Integer)i);
    } else if (i instanceof String) {
      return _m((String)i);
    } else {
      throw new IllegalArgumentException("Unhandled parameter types: " +
        Arrays.<Object>asList(i).toString());
    }
  }
...
```

And now I have this code coverage:

image

Covering generated entry method for dispatch methods is even much harder in the presence of class inheritance where some dispatch methods are overridden, because of lots of if-then-else branches.

@szarnekow what do you think? I can prepare a PR.

Maybe the same strategy could be applied in the presence of `create` methods? Honestly, I never used them.

LorenzoBettini:
Closes #3077 

I put the new `Generated` annotation in xbase.lib instead of xtend.lib because it could be useful in general for Xbase languages

### DIFF
--- a/org.eclipse.xtend.core.tests/batch-compiler-data/xtendClass/XtendA.xtend
+++ b/org.eclipse.xtend.core.tests/batch-compiler-data/xtendClass/XtendA.xtend
@@ -9,4 +9,7 @@
 
 class XtendA {
 	val public String foo = ""
+
+	def dispatch m(String s) {}
+	def dispatch m(int i) {}
 }
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug423631Test.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug423631Test.xtend
@@ -30,7 +30,7 @@ class CompilerBug423631Test extends AbstractXtendCompilerTest {
 			}
 		''', '''
 			import java.util.Arrays;
-			import org.eclipse.xtext.xbase.lib.Generated;
+			import org.eclipse.xtext.xbase.lib.XbaseGenerated;
 			
 			@SuppressWarnings("all")
 			public class C {
@@ -42,7 +42,7 @@ class CompilerBug423631Test extends AbstractXtendCompilerTest {
 			    return null;
 			  }
 			
-			  @Generated
+			  @XbaseGenerated
 			  public Object m(final A a) {
 			    if (a instanceof B) {
 			      return _m((B)a);
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug423631Test.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug423631Test.xtend
@@ -30,6 +30,7 @@ class CompilerBug423631Test extends AbstractXtendCompilerTest {
 			}
 		''', '''
 			import java.util.Arrays;
+			import org.eclipse.xtext.xbase.lib.Generated;
 			
 			@SuppressWarnings("all")
 			public class C {
@@ -41,6 +42,7 @@ class CompilerBug423631Test extends AbstractXtendCompilerTest {
 			    return null;
 			  }
 			
+			  @Generated
 			  public Object m(final A a) {
 			    if (a instanceof B) {
 			      return _m((B)a);
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug427637Test.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug427637Test.xtend
@@ -1082,9 +1082,9 @@ class CompilerBug427637Test extends AbstractXtendCompilerTest {
 			import org.eclipse.xtext.xbase.lib.CollectionLiterals;
 			import org.eclipse.xtext.xbase.lib.Extension;
 			import org.eclipse.xtext.xbase.lib.Functions.Function1;
-			import org.eclipse.xtext.xbase.lib.Generated;
 			import org.eclipse.xtext.xbase.lib.IterableExtensions;
 			import org.eclipse.xtext.xbase.lib.Procedures.Procedure1;
+			import org.eclipse.xtext.xbase.lib.XbaseGenerated;
 			import org.eclipse.xtext.xbase.lib.util.ToStringHelper;
 			
 			@SuppressWarnings("all")
@@ -1590,7 +1590,7 @@ class CompilerBug427637Test extends AbstractXtendCompilerTest {
 			    return _function;
 			  }
 			
-			  @Generated
+			  @XbaseGenerated
 			  protected Function1<? super FormattableDocument, ? extends Iterable<FormattingData>> newFormattingData(final HiddenLeafs leafs, final PreferenceKey key, final FormattingDataFactory.FormattingDataInit it) {
 			    if (key instanceof BlankLineKey) {
 			      return _newFormattingData(leafs, (BlankLineKey)key, it);
--- a/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug427637Test.xtend
+++ b/org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/compiler/CompilerBug427637Test.xtend
@@ -1082,6 +1082,7 @@ class CompilerBug427637Test extends AbstractXtendCompilerTest {
 			import org.eclipse.xtext.xbase.lib.CollectionLiterals;
 			import org.eclipse.xtext.xbase.lib.Extension;
 			import org.eclipse.xtext.xbase.lib.Functions.Function1;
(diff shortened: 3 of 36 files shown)


szarnekow:
I don't really like the generic name `@Generated` - instead I'd prefer a more targeted that could even carry additional information, e.g. `@GeneratedDispatchMethod(dispatchCase="_myMethod", arity=3)` that could be used to unambiguously detect dispatch cases from generated Java code. What do you think about that?
Can Sonar consider meta-annotations, e.g. `@MyAnnotation` if `MyAnnotation` is annotated with `@XtextGenerated` or similar?

LorenzoBettini:
> I don't really like the generic name `@Generated` - instead I'd prefer a more targeted that could even carry additional information, e.g. `@GeneratedDispatchMethod(dispatchCase="_myMethod", arity=3)` that could be used to unambiguously detect dispatch cases from generated Java code. What do you think about that? 

OK, but in that case the annotation should be in `xtend.lib`, right? In that project all annotations are active annotations. Would it still be OK to have a standard annotation there?

> Can Sonar consider meta-annotations, e.g. `@MyAnnotation` if `MyAnnotation` is annotated with `@XtextGenerated` or similar?

Do you mean JaCoCo or Sonar?
Do you mean to also exclude meta annotations from code coverage? In that case, what would be an example?
To be honest I'm not familiar with meta-annotations.

cdietrich:
how does this overlap with the preference

grafik

szarnekow:
> OK, but in that case the annotation should be in xtend.lib

If we agree on such a specific annotation, I'd put it next to StringConcatenationClient with class retention. 

> meta-annotations

Here is a very concise description of meta-annotations: https://docs.spring.io/spring-framework/reference/core/beans/classpath-scanning.html#beans-meta-annotations

In a nutshell: Rather than putting the `@Generated` directly on the annotation target, another annotation is annotated as `@Generated` and that one is put on the annotation target. That would allow `@Dispatcher` to be used (arguably a better name than `@GeneratedDispatcher`). 

But @cdietrich raises a fair question: Would we always emit `@XbaseGenerated` or whatever the name should be?

cdietrich:
yes. am not sure if we still have the old javax.annotation.Generated present.
am also not sure if we want the annotation to be used configurable

LorenzoBettini:
@szarnekow so a meta-annotation is just an annotation put on another annotation; if I understand correctly what you suggest, I manually tried and it didn't work:

```java
@Retention(RetentionPolicy.CLASS)
@Target({TYPE, METHOD, CONSTRUCTOR})
public @interface Generated {
}
```

and

```java
@Retention(RetentionPolicy.CLASS)
@Target({TYPE, METHOD, CONSTRUCTOR})
@Generated
public @interface Dispatcher {

	String dispatchCase();

	int arity();
}
```

but JaCoCo does NOT ignore the annotated method:

```java
  @Dispatcher(dispatchCase = "_m", arity = 1)
  public String m(final Object i) {
    if (i instanceof Integer) {
      return _m((Integer)i);
    } else if (i instanceof String) {
      return _m((String)i);
    } else {
      throw new IllegalArgumentException("Unhandled parameter types: " +
        Arrays.<Object>asList(i).toString());
    }
  }
```

As I said in https://github.com/eclipse/xtext/issues/3077 JaCoCo expects that "The name of the annotation includes Generated."

Concerning `javax.annotation.Generated`, related to the configuration option shown by @cdietrich above,

```java
@Retention(SOURCE)
@Target({PACKAGE, TYPE, ANNOTATION_TYPE, METHOD, CONSTRUCTOR, FIELD, 
        LOCAL_VARIABLE, PARAMETER})
public @interface Generated {
```

That is a source annotation so it cannot be used to this aim.

Concerning always generating the new annotation, I'd go with always generating that. And it's unrelated to the other existing one.

The current `Generated` annotation, like `SuppressWarning`, is something handled by Xbase, while this new one would be specific of Xtend if we go with the version with `GeneratedDispatcher` in the name and arguments related to the dispatcher method as suggested by Sebastian. In that case, Sebastian, you say to put it next to `org.eclipse.xtend2.lib.StringConcatenationClient` but it's still in the project `xbase.lib`.

LorenzoBettini:
So: we'll use `@XbaseGenerated` and to keep the annotation with class retention policy.

cdietrich:
Can we check the annotation removal in lsp4j

cdietrich:
(Lsp4j uses xtend/xbase lib as compile time only dependencies

LorenzoBettini:
> Can we check the annotation removal in lsp4j

Could you please elaborate? I don't understand what you mean

cdietrich:
in lsp4j the java classes generated from xtend classes
shall not make use of any classes in xbase.lib and xtend.lib

we have some code in annotation processing that takes care.
my quesiton is: do these changes break this

LorenzoBettini:
> in lsp4j the java classes generated from xtend classes shall not make use of any classes in xbase.lib and xtend.lib
> 
> we have some code in annotation processing that takes care. my quesiton is: do these changes break this

Could you please point me to where we take care of this (annotation processing)?
And some pointers to lsp4j using Xtend?

cdietrich:
just check the lsp4j repo and https://github.com/eclipse-lsp4j/lsp4j/blob/f235e91fbe2e45f62e185bbb9f6d21bed48eb2b9/org.eclipse.lsp4j.generator/src/main/java/org/eclipse/lsp4j/generator/JsonRpcDataProcessor.xtend#L27

unfortunately your branch does not build.
so i cannot try to use it

cdietrich:
maybe we are safe if they dont use dispatch

szarnekow:
We should also make this optional (enabled by default) and expose a generator config that allows to opt-out from that

LorenzoBettini:
Sorry I didn't have time yet to fix the failing tests.
I'll try to do that ASAP, but maybe tomorrow.
I'll also try to add a new option enabled by default, taking inspiration from the current options. Of course, if you have suggestions on how the option could be named...

szarnekow:
Label in UI could be `Annotate synthetic members with @XtextGenerated`

LorenzoBettini:
@szarnekow So `XtextGenerated`, not `XbaseGenerated`?

szarnekow:
> @szarnekow So `XtextGenerated`, not `XbaseGenerated`?

Freudian slip. XbaseGenerated

cdietrich:
lsp4j does not seem to use dispatches.
and if they were the Architecture Test i have written would catch it

java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'no classes that reside in a package 'org.eclipse.lsp4j..' and not reside in a package 'org.eclipse.lsp4j.generator..' and not reside in a package 'org.eclipse.lsp4j.test..' should depend on classes that reside in a package 'org.eclipse.xtext..'' was violated (1 times):
Method <org.eclipse.lsp4j.Viech.x(java.io.InputStream)> is annotated with <org.eclipse.xtext.xbase.lib.XbaseGenerated> in (Viech.java:0)

LorenzoBettini:
@cdietrich Thanks for testing it!

I'll add the compiler option soon anyway.

LorenzoBettini:
@szarnekow I've implemented the new option, the involved commit is https://github.com/eclipse/xtext/pull/3080/commits/a33af2e49acf4bfc33003fb258fbaa89910481de

Here's how the properties dialog looks like

Screenshot from 2024-06-26 13-38-12

By enabling project specific settings, and disabling the option:

Screenshot from 2024-06-26 13-40-45

By enabling the option again:

Screenshot from 2024-06-26 13-41-10

LorenzoBettini:
ping...

LorenzoBettini:
@cdietrich @szarnekow could we take a decision on this please?

cdietrich:
sry i have zero cappa for looking at anything atm.
the efforts for the MWE and Xtext milestones already exeed by cappa.

mehmet-karaman:
The failing tests work on my branch.. I am going to rebase and run these tests again.

PR #3090, opened 2024-06-14
Instantiate ClassFileReader via reflection

### LINKED ISSUE #3089: Fix api breakage with newest jdt
05:09:28  [ERROR] The method read(byte[], String) in the type ClassFileReader is not applicable for the arguments (InputStream, String)

Cc @iloveeclipse is this an intentional change upstream

cdietrich:
Fixes #3089

### DIFF
--- a/org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.java
+++ b/org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/IndexAwareNameEnvironment.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2018, 2020 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2018, 2024 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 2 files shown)


cdietrich:
problems

- new jdt maven wont be published before the Xtext release
- likely newer platform artifacts will even require java 21 to run.

=> we would have to drop out from 2024-09 or explode there and or set minimal version to 21
i dont know what the exact plans there are though

cdietrich:
@LorenzoBettini where would the parkplatz for this utility be

cdietrich:
but xbase.testing does not depend to does it?

LorenzoBettini:
I would go with the project that uses JDT internals, which are used by both projects involved in this PR, and move that utility method there; I don't have an Xtext workspace at the moment.

In any case, I can work on that later or in the near future without blocking this PR.

cdietrich:
lets see what @szarnekow says

szarnekow:
I would land this here. It’s unfortunate that JDT breaks in such a simple way but there’s not much we can do about that. Subsequently we could perform a systematic refactoring to extract code into a new bundle. That won’t help with old versions though. Same as this PR does only help for future versions

LorenzoBettini:
@cdietrich so shall we merge this?

PR #3097, opened 2024-06-28
fix ProgressReportingTest

cdietrich:


cdietrich:
am not sure if this is too naive

LorenzoBettini:
That's what we've always done...

However, I don't know the value of that test.

LorenzoBettini (approved):

PR #3099, opened 2024-07-09
Refactoring the AbstractLinkedEditingIntegrationTest test cases.

miklossy:
- Deprecate the `waitForDisplay()` method in the `AbstractLinkedEditingIntegrationTest` class in favor of the `waitForEventProcessing()` method in the `AbstractWorkbenchTest` class.
- Modify the test cases inherited from the `AbstractLinkedEditingIntegrationTest` base class to use the preferred method instead of the deprecated one.

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractLinkedEditingIntegrationTest.java
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractLinkedEditingIntegrationTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2011, 2024 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractLinkedEditingIntegrationTest.java
```diff
@@ -32,6 +32,10 @@ public void process(XtextResource state) throws Exception {
 		});
 	}
 	
+	/**
+	 * deprecated, use the inherited {@link AbstractWorkbenchTest#waitForEventProcessing} method instead.
```
LorenzoBettini, line 36: Shouldn't "deprecated" in the javadoc be an annotation as well?

miklossy: Yes, indeed. It should be an annotation. I have just updated the PR.

LorenzoBettini (approved):
I'm fine with the PR.
Would you mind leaving a comment on the PR on why we deprecate this method?

miklossy:
The method `AbstractLinkedEditingIntegrationTest.waitForDisplay()` has been deprecated in favor of the inherited `AbstractWorkbenchTest.waitForEventProcessing()` method. Both methods are doing the same (only the method names are different). For a long term we would like to clean up the duplicated code. The first step is to deprecate the method we would like to remove with the `@Deprecated` annotation , the second step is to mark it as `@Deprecated(forRemoval=true)` and the third step is to delete the method. But it will take a couple of more releases. This PR takes care of the first step.

miklossy:
Thanks for approving my PR @LorenzoBettini !

LorenzoBettini:
@miklossy Thanks! Please update the release notes about this change :)

miklossy:
> @miklossy Thanks! Please update the release notes about this change :)

done: https://github.com/eclipse/xtext-website/pull/54

PR #3103, opened 2024-07-12
support J21 in jenkins

cdietrich:


LorenzoBettini (approved):

cdietrich:
Unfortunately this breaks the j11 build

LorenzoBettini:
mh... why is it related?
However, I see we don't have a specification for jdk 11 in the tools in Jenkinsfile

cdietrich:
Assume.assumeTrue("Active only on Java 21 and later", JavaRuntimeVersion.isJava21OrLater());

this guard does not seem to work

LorenzoBettini:
This

```
11:26:42  [INFO] Command line:
11:26:42  	[/opt/tools/java/temurin/jdk-21/latest/bin/java
```

shows that Tycho is not using Java 11 to run the tests.

Jenkinsfile
```diff
@@ -26,6 +26,7 @@ pipeline {
 
   tools {
      jdk "temurin-jdk17-latest"
```
LorenzoBettini, line 28: Here we miss the specification for jdk11

LorenzoBettini:
And this

```
11:19:26  [INFO] --- toolchains:3.1.0:toolchain (default) @ xtext-parent ---
11:19:26  [INFO] Required toolchain: jdk [ version='21' ]
11:19:26  [INFO] Found matching toolchain for type jdk: JDK[/opt/tools/java/temurin/jdk-21/latest]
```

shows that we either specify the wrong jdk version or it doesn't find a matching one

LorenzoBettini:
I'll do some experiments on top of your branch.

We should check whether the whole thing used to work before this change...

LorenzoBettini:
Either I miss something or from the main builds I cannot see any build with Java 11... wasn't that part of the timer-based builds (building with Java 11 and the olderst platform 4.23)?

cdietrich:
grafik

cdietrich:
maybe the daily java 11 buidls there broke too.
or it got lost.
i would have expected its still running.
but as i check fails only i have no idea

LorenzoBettini:
@cdietrich I found the problem, and it's my fault O:)

we enable the `strict-release-jdk` profile (https://github.com/eclipse/xtext/blob/839911c9ebcc1fb7c18ae4563f03f6334b4e596b/pom.xml#L423) but then I had added another profile that gets activated automatically when the JDK is 21 (https://github.com/eclipse/xtext/blob/839911c9ebcc1fb7c18ae4563f03f6334b4e596b/pom.xml#L449) this other profile forces `<maven.compiler.release>21</maven.compiler.release>` and so the `strict-release-jdk` asks Maven to use Java 21 for the toolchain, voiding our goals for that profile.

I'll have to adjust a few things with those profiles, so that they won't disturb each other.

Maybe, the easiest and cleanest thing would be to have 2 profiles `strict-jdk-11` and `strict-jdk-21`, both activated explicitly, what do you think? This requires updating a few scripts of course.

LorenzoBettini:
> maybe the daily java 11 buidls there broke too.
> or it got lost.

the one you put in the screenshot was started manually by me

LorenzoBettini:
@cdietrich this should be superseded by #3105

PR #3105, opened 2024-07-19
support Java 21 in Jenkins

LorenzoBettini:
Now the profiles for Java 11 and Java 21 are explicit and will not disturb each other.
Since the profile for Java 21 is not activated automatically anymore, the GitHub Actions workflows have been updated accordingly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ pipeline {
   parameters {
     choice(name: 'TARGET_PLATFORM', choices: ['r202203', 'r202206', 'r202209', 'r202212', 'r202303', 'r202306', 'r202309', 'r202312', 'r202403', 'r202406', 'latest'], description: 'Which Target Platform should be used?')
     // see https://wiki.eclipse.org/Jenkins#JDK
-    choice(name: 'JDK_VERSION', choices: [ '11', '17' ], description: 'Which JDK version should be used?')
+    choice(name: 'JDK_VERSION', choices: [ '11', '17', '21' ], description: 'Which JDK version should be used?')
   }
 
   triggers {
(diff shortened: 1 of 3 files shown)


LorenzoBettini:
Supersedes https://github.com/eclipse/xtext/pull/3103

Jenkinsfile
```diff
@@ -64,7 +66,8 @@ pipeline {
         xvnc(useXauthority: true) {
           sh """
             ./full-build.sh --tp=${selectedTargetPlatform()} \
-              ${javaVersion() == 17 ? '' : '--toolchains releng/toolchains.xml -Pstrict-release-jdk'}
+              ${javaVersion() == 11 ? '--toolchains releng/toolchains.xml -Pstrict-jdk-11' : ''} \
```
cdietrich, line 69: no j17 clause needed?

cdietrich:
looks good lets see what jenkins does. will also schedule 17 and 21 builds there

Jenkinsfile
LorenzoBettini: I forgot that we wanted to always build with Java 11, I'll revert the check as it was before

cdietrich:
@LorenzoBettini scheduled them against 2024-06

Jenkinsfile
LorenzoBettini: @cdietrich No wait: the check is correct like that: previously there were only Java 17 and Java 11, and if it was Java 17 we didn't do anything otherwise (Java 11) we set the toolchain (which corresponds to what I do now).
Now there's also Java 21, so we must explicitly check for Java 11.
For Java 17 nothing is required.

cdietrich (approved):

cdietrich:
have rebased. how now builds go through

cdietrich:
thx @LorenzoBettini

Issue #2385, opened 2019-09-20
Refresh on Access does not work for Xtext editors


cdietrich:
Refresh on Access does not work for Xtext editors

(2) configure workspace for refresh on access
(1) open a Xtext based dsl file
(2) modify on disk
(3) click into the editor

Expected behaviour:
Editor refreshes (e.g. as `.java` or `.txt`)
Actual behaviour:
Editor does not refresh

This is caused by the customization in

`org.eclipse.xtext.ui.editor.model.XtextDocumentProvider.isSynchronized(Object)`

i have no idea why the code there is like it is and what doing some copy&paste from superclass would break

Question: can the `XtextDocumentProvider` be fixed without breaking something else

iloveeclipse:
> This is caused by the customization in
> `org.eclipse.xtext.ui.editor.model.XtextDocumentProvider.isSynchronized(Object)`
> i have no idea why the code there is like it is and what doing some copy&paste from superclass would break
> Question: can the `XtextDocumentProvider` be fixed without breaking something else

That is what we've also just found. We plan to provide a PR that should fix that, current state is just not acceptable from the user point of view.

mehmet-karaman:
Removing the isSynchronized implementation and make use of its super implementation works. But for the case if the editor was dirty and the file was edited outside of eclipse, the user gets a dialog in which he can choose loading of the new content or ignore the change. If the user selects to ignore the changes, the dirty editor can't save it anymore, because the file is out of sync. So I am going to investigate this problem too.

mehmet-karaman:
@szarnekow 
@cdietrich 

Does this PR has any issues? Do I have to do something?

mehmet-karaman:
@szarnekow 
I guess this ticket can be closed now?

szarnekow:
Indeed. Thank you for the heads-up.

LorenzoBettini:
@mehmet-karaman one of the added tests seems to be flaky (especially in macOS):

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertFalse(Assert.java:65)
	at org.junit.Assert.assertFalse(Assert.java:75)
	at org.eclipse.xtext.ui.tests.editor.DirtyStateEditorSupportIntegrationTest.testModifyFileInExternEditor(DirtyStateEditorSupportIntegrationTest.java:148)
```

[...]

Issue #3102, opened 2024-07-12
org.eclipse.xtend.ide.tests.buildpath.JavaClasspathTest.testNoJavaInClasspath fails on latest


cdietrich:
org.eclipse.xtend.ide.tests.buildpath.JavaClasspathTest.testNoJavaInClasspath fails on latest
reason unclear

cdietrich:
```
2024-07-12T17:28:51.1734694Z Running org.eclipse.xtend.ide.tests.buildpath.JavaClasspathTest
2024-07-12T17:28:56.7282243Z Tests run: 7, Failures: 6, Errors: 0, Skipped: 0, Time elapsed: 5.568 s <<< FAILURE! -- in org.eclipse.xtend.ide.tests.buildpath.JavaClasspathTest
2024-07-12T17:28:56.7284493Z org.eclipse.xtend.ide.tests.buildpath.JavaClasspathTest.testNoJavaInClasspath -- Time elapsed: 0.382 s <<< FAILURE!
2024-07-12T17:28:56.7285903Z java.lang.AssertionError: 
2024-07-12T17:28:56.7286382Z LogEntry [
2024-07-12T17:28:56.7287194Z   message = "Error initializing type java:/Objects/org.eclipse.xtext.xbase.lib.ArrayExtensions"
2024-07-12T17:28:56.7288468Z   source = "org.eclipse.xtext.common.types.access.jdt.JdtTypeMirror"
2024-07-12T17:28:56.7289278Z   timeStamp = 1720805331605
2024-07-12T17:28:56.7289711Z   level = ERROR
2024-07-12T17:28:56.7290112Z ] expected:<true> but was:<false>
2024-07-12T17:28:56.7290655Z 	at org.junit.Assert.fail(Assert.java:89)
2024-07-12T17:28:56.7291331Z 	at org.junit.Assert.failNotEquals(Assert.java:835)
2024-07-12T17:28:56.7292085Z 	at org.junit.Assert.assertEquals(Assert.java:120)
2024-07-12T17:28:56.7293232Z 	at org.eclipse.xtend.ide.tests.buildpath.JavaClasspathTest$1.run(JavaClasspathTest.java:79)
2024-07-12T17:28:56.7294742Z 	at org.eclipse.xtext.testing.logging.LoggingTester.captureLogging(LoggingTester.java:73)
[...]

PR #3108, opened 2024-07-23
Fix editor sync bug, when model file was edited outside of eclipse.

mehmet-karaman:
This bugfix contains the fix for unsync workspace files wrt. the file system (see the
DirtyStateEditorSupportIntegrationTest.testModifyFileInExternEditor) and a one liner for beeing able to save the file when the changes were ignored. (see the ignored test case
DirtyStateEditorSupportIntegrationTest.testModifyDirtyFileInExternEditor).

The issue for this fix can be found here:
https://github.com/eclipse/xtext/issues/2385

### DIFF
--- a/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
+++ b/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
@@ -8,28 +8,36 @@
  */
 package org.eclipse.xtext.ui.tests.editor;
 
+import java.io.File;
+import java.io.FileWriter;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.jface.text.IUndoManager;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.StyledText;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Event;
+import org.eclipse.ui.IFileEditorInput;
+import org.eclipse.ui.editors.text.FileDocumentProvider;
 import org.eclipse.xtext.resource.IEObjectDescription;
 import org.eclipse.xtext.resource.IResourceDescription;
 import org.eclipse.xtext.resource.IResourceServiceProvider;
 import org.eclipse.xtext.ui.editor.XtextEditor;
 import org.eclipse.xtext.ui.editor.XtextSourceViewer;
+import org.eclipse.xtext.ui.editor.model.IXtextDocument;
 import org.eclipse.xtext.ui.refactoring.ui.SyncUtil;
 import org.eclipse.xtext.ui.testing.AbstractEditorTest;
 import org.eclipse.xtext.ui.testing.util.IResourcesSetupUtil;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import com.google.common.collect.Iterables;
@@ -116,7 +124,72 @@ public class DirtyStateEditorSupportIntegrationTest extends AbstractEditorTest {
 				.getQualifiedName().getLastSegment());
 		Iterables.getLast(events);
 	}
+	
+	/*
+	 * @see https://github.com/eclipse/xtext/issues/2385
+	 */
+	@Test public void testModifyFileInExternEditor() throws Exception {
+		IXtextDocument document = editor.getDocument();
 
+		Display.getDefault().readAndDispatch();
+		
+		assertNotEquals(document.get(), "");
+		assertTrue(editor.getDocumentProvider() instanceof FileDocumentProvider);
+		
+		FileDocumentProvider fileDocumentProvider = (FileDocumentProvider)editor.getDocumentProvider();
+		
+		getWorkbench().getActiveWorkbenchWindow().getActivePage().getViewReferences()[0].getView(false).setFocus();
+		
+		File externalEditFile = ((IFileEditorInput)editor.getEditorInput()).getFile().getLocation().toFile();
+		try (FileWriter fw = new FileWriter(externalEditFile)) {
+			fw.write("");
+		}
+		
+		assertFalse(fileDocumentProvider.isSynchronized(editor.getEditorInput()));
+
+		editor.setFocus();
+		Job.getJobManager().join(ResourcesPlugin.FAMILY_AUTO_REFRESH, null);
+		syncUtil.yieldToQueuedDisplayJobs(new NullProgressMonitor());
+		syncUtil.waitForReconciler(editor);
+		syncUtil.waitForBuild(new NullProgressMonitor());
+		
+		assertTrue(fileDocumentProvider.isSynchronized(editor.getEditorInput()));
+		
+		assertEquals(document.get(), "");
+	}
+
+	/**
+	 * This test needs a manual button click.
+	 */
+	@Ignore
+	@Test public void testModifyDirtyFileInExternEditor() throws Exception {
+		IXtextDocument document = editor.getDocument();
+
+		Display.getDefault().readAndDispatch();
+		
+		assertNotEquals(document.get(), "");
+		document.set("/* Hello World! */");
+		assertEquals("/* Hello World! */", document.get());
+		assertTrue(editor.isDirty());
+		
+		getWorkbench().getActiveWorkbenchWindow().getActivePage().getViewReferences()[0].getView(false).setFocus();
+		
+		File externalEditFile = ((IFileEditorInput)editor.getEditorInput()).getFile().getLocation().toFile();
+		try (FileWriter fw = new FileWriter(externalEditFile)) {
+			fw.write("");
+		}
+		editor.setFocus();
+		// Dialog opens and the "Ignore file change" button has to be clicked.
+		Job.getJobManager().join(ResourcesPlugin.FAMILY_AUTO_REFRESH, null);
+		syncUtil.yieldToQueuedDisplayJobs(new NullProgressMonitor());
+		syncUtil.waitForReconciler(editor);
+		assertEquals(document.get(), "/* Hello World! */");
+		
+		assertTrue(editor.isDirty());
+		editor.doSave(new NullProgressMonitor());
+		assertFalse(editor.isDirty());
+	}
+	
 	private IEObjectDescription getFirstExportedObjectInLastEventDelta() {
 		return Iterables.getFirst(Iterables.getLast(events).getDeltas().get(0).getNew().getExportedObjects(), null);
 	}
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/XtextEditor.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/XtextEditor.java
@@ -76,6 +76,8 @@ import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.editors.text.TextEditor;
 import org.eclipse.ui.texteditor.DefaultMarkerAnnotationAccess;
+import org.eclipse.ui.texteditor.IDocumentProvider;
+import org.eclipse.ui.texteditor.IDocumentProviderExtension3;
 import org.eclipse.ui.texteditor.ITextEditor;
 import org.eclipse.ui.texteditor.ITextEditorActionConstants;
 import org.eclipse.ui.texteditor.ITextEditorActionDefinitionIds;
@@ -327,6 +329,20 @@ public class XtextEditor extends TextEditor implements IDirtyStateEditorSupportC
 		callback.afterSave(this);
 	}
 	
+	@Override
+	protected void performSave(boolean overwrite, IProgressMonitor progressMonitor) {
+		super.performSave(overwrite || isSynchronized(), progressMonitor);
+	}
+	
+	
+	private boolean isSynchronized() {
+		IResource resource = getResource();
+		if (resource != null) {
+			return resource.isSynchronized(IResource.DEPTH_ZERO);
+		}
+		return false;
+	}
+
 	@Override
 	protected void updateState(final IEditorInput input) {
 		new DisplayRunnable() {
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
@@ -409,20 +409,6 @@ public class XtextDocumentProvider extends FileDocumentProvider {
 	}
 	
 	@Override
-	public boolean isSynchronized(Object element) {
-		ElementInfo info = getElementInfo(element);
-		long synchronizationStamp;
-		if (info instanceof FileInfo) {
-			synchronizationStamp = ((FileInfo) info).fModificationStamp;
-		} else if (info instanceof URIInfo) {
-			synchronizationStamp = ((URIInfo) info).synchronizationStamp;
-		} else {
-			return super.isSynchronized(element);
-		}
-		return synchronizationStamp == getModificationStamp(element);
-	}
-	
-	@Override
 	public boolean isModifiable(Object element) {
 		if (isWorkspaceExternalEditorInput(element)) {
 			URIInfo info= (URIInfo) getElementInfo(element);


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
```diff
@@ -410,16 +411,29 @@ public long getModificationStamp(Object element) {
...
 	public boolean isSynchronized(Object element) {
-		ElementInfo info = getElementInfo(element);
-		long synchronizationStamp;
-		if (info instanceof FileInfo) {
-			synchronizationStamp = ((FileInfo) info).fModificationStamp;
-		} else if (info instanceof URIInfo) {
-			synchronizationStamp = ((URIInfo) info).synchronizationStamp;
-		} else {
-			return super.isSynchronized(element);
+		if (element instanceof IFileEditorInput) {
```
iloveeclipse, line 414: I personally avoid many nested if levels, so I would revert the condition and also use modern instanceof
if (!(element instanceof IFileEditorInput input)){
   return super.isSynchronized(element);
}

// continue with the new code
// "input" is already defined here with the right type

iloveeclipse:
There are 7 test failures:

[Test Result](https://ci.eclipse.org/xtext/job/xtext/job/PR-3108/1/testReport/) (7 failures )
[org.eclipse.xtend.ide.tests.hover.XtendHoverInEditorTest.testHoverOfJavaTypeInJavadocLink](https://ci.eclipse.org/xtext/job/xtext/job/PR-3108/1/testReport/junit/org.eclipse.xtend.ide.tests.hover/XtendHoverInEditorTest/testHoverOfJavaTypeInJavadocLink/)
[org.eclipse.xtend.ide.tests.hover.XtendHoverInEditorTest.testHoverOfFQNJavaTypeInJavadocLink](https://ci.eclipse.org/xtext/job/xtext/job/PR-3108/1/testReport/junit/org.eclipse.xtend.ide.tests.hover/XtendHoverInEditorTest/testHoverOfFQNJavaTypeInJavadocLink/)
[org.eclipse.xtend.ide.tests.hover.XtendHoverInEditorTest.testHoverOfJavaType](https://ci.eclipse.org/xtext/job/xtext/job/PR-3108/1/testReport/junit/org.eclipse.xtend.ide.tests.hover/XtendHoverInEditorTest/testHoverOfJavaType/)
[org.eclipse.xtend.ide.tests.hover.XtendHoverInEditorTest.testHoverOfFQNJavaType](https://ci.eclipse.org/xtext/job/xtext/job/PR-3108/1/testReport/junit/org.eclipse.xtend.ide.tests.hover/XtendHoverInEditorTest/testHoverOfFQNJavaType/)
[org.eclipse.xtend.ide.tests.hover.XtendHoverInEditorTest.testHoverOfJavaTypeInJavadocSee](https://ci.eclipse.org/xtext/job/xtext/job/PR-3108/1/testReport/junit/org.eclipse.xtend.ide.tests.hover/XtendHoverInEditorTest/testHoverOfJavaTypeInJavadocSee/)
[org.eclipse.xtend.ide.tests.hover.XtendHoverInEditorTest.testHoverOfFQNJavaTypeInJavadocSee](https://ci.eclipse.org/xtext/job/xtext/job/PR-3108/1/testReport/junit/org.eclipse.xtend.ide.tests.hover/XtendHoverInEditorTest/testHoverOfFQNJavaTypeInJavadocSee/)
[org.eclipse.xtend.ide.tests.quickfix.QuickfixTest.missingConcreteMembers2](https://ci.eclipse.org/xtext/job/xtext/job/PR-3108/1/testReport/junit/org.eclipse.xtend.ide.tests.quickfix/QuickfixTest/missingConcreteMembers2/)

@cdietrich : are the test failures known?
@mehmet-karaman : could you please check if the tests are failing locally with your patch?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
mehmet-karaman: Currently the project uses java 11. Am I allowed to upgrade to java 17 or 21?

iloveeclipse: > Am I allowed to upgrade to java 17 or 21?

I would say, definitely not **in the context of this ticket**, but it is a good question and probably deserves a dedicated ticket in Xtext.

Note that Eclipse Platform is switched to Java 17 in Eclipse 4.28, including ecj (see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/886).

cdietrich: there is already a ticket for drop 11 17 minimum
https://github.com/eclipse/xtext/issues/2686

cdietrich:
i am only aware of https://github.com/eclipse/xtext/issues/3102
the others are new to me

cdietrich:
started j11 build on main
https://ci.eclipse.org/xtext/job/xtext/job/main/1325/

cdietrich:
i can reproduce locally, but no idea why
@LorenzoBettini do you know why this
https://ci.eclipse.org/xtext/job/xtext/job/main/1325/consoleFull
seems to run with 17 and not 11?

LorenzoBettini:
> i can reproduce locally, but no idea why @LorenzoBettini do you know why this https://ci.eclipse.org/xtext/job/xtext/job/main/1325/consoleFull seems to run with 17 and not 11?

@cdietrich from what I see

```
09:52:11  [INFO] Command line:
09:52:11  	[/opt/tools/java/temurin/jdk-11/latest/bin/java
```

it runs using JDK 11 (of course, the build runs with Java 17, but the toolchain uses JDK11)

mehmet-karaman:
The failing tests are running successful on my branch. I am going to rebase on the latest state and try agian.

mehmet-karaman:
Tests are also successful after rebase.. Pushed the branch again.

iloveeclipse:
Jenkins is fine, new test is executed. Whatever the pipeline is doing and why it is failing is not our beer, it seem to fail on all other recent PR's too.

Xtext committers: Could we proceed with the review?

mehmet-karaman:
@szarnekow @cdietrich Do i have to do something on my side? Does this PR has any problems?

cdietrich:
problem: ustream is never green cause all of the jdt changes.
i want sebastian to have a review cause i cannot judge if it may have bad side effect i dont see.

cdietrich:
@mehmet-karaman can you please rebase

iloveeclipse:
All tests are green now. Cool.

iloveeclipse:
@cdietrich, @szarnekow : could we proceed with the review now?

cdietrich:
as i said i cannot judge it it will break something. and in my 12 and 13th working hour i also have no capacity to think about it.

 and sebastian has no time too i assume

org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
```diff
@@ -116,7 +123,62 @@ public void testUndoRedo() throws Exception {
...
+		}
+		editor.setFocus();
+		Job.getJobManager().join(ResourcesPlugin.FAMILY_AUTO_REFRESH, null);
+		syncUtil.yieldToQueuedDisplayJobs(new NullProgressMonitor());
+		syncUtil.waitForReconciler(editor);
+		assertEquals(document.get(), "");
+	}
+
+	/**
+	 * This test needs a manual button click.
```
LorenzoBettini, line 151: This comment sounds strange: does a human have to click on the button to make the test proceed?

LorenzoBettini: Mine is not meant to be a review, but I don't understand what the test does: if it's ignored, it doesn't test anything.

I don't think clicking on a dialog button is (easily) doable in a plug-in test, but it's easy in a SWTBot test.

Maybe a SWTBot test to effectively verify the fix would be better.

But, again, I did not follow the whole bug and fix.

iloveeclipse: Sure, therefore there is one automated test and one that needs manual interaction. 
Not sure if SWTBot is used in Xtext code.

LorenzoBettini: We use SWTBot tests, just a few, to test Xtend dialogs.

My main concern is that there's no automatic test for verifying the fix, again, if I understand correctly.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/XtextEditor.java
```diff
@@ -327,6 +328,14 @@ public void doSave(IProgressMonitor progressMonitor) {
 		callback.afterSave(this);
 	}
 	
+	@Override
+	protected void performSave(boolean overwrite, IProgressMonitor progressMonitor) {
+		boolean unsyncOverwrite = !((XtextDocumentProvider)getDocumentProvider()).isSynchronized(getEditorInput());
```
szarnekow, line 333: Preferably use an instance-of guarded cast to `org.eclipse.ui.texteditor.IDocumentProviderExtension3.isSynchronized(Object)` for the unlikely though possible case where the document provider is overridden

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
```diff
@@ -419,7 +428,15 @@ public boolean isSynchronized(Object element) {
...
 			return super.isSynchronized(element);
 		}
-		return synchronizationStamp == getModificationStamp(element);
+
+		// Check has to fit the following method: 
+		// org.eclipse.ui.editors.text.FileDocumentProvider.checkSynchronizationState(long, IResource)
+		boolean isSynchronized = synchronizationStamp == getSynchronizationStamp(input.getFile());
+		if (!isSynchronized) {
+			// checks for is sync and triggers sync if necessary
+			input.getFile().isSynchronized(IResource.DEPTH_ZERO);
```
szarnekow, line 437: We would still return `false` though. Why is this the correct result and why is the async refresh useful here?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/XtextEditor.java
mehmet-karaman: IResource.isSynchornized() doesn't have the same understanding of FileDocumentProvider.checkSynchronizationState() of synchronzied.. If XtextDocumentProvider.isSynchronized is called multiple times, it will be for the first time unsynchron (return false) and for the next calls it will IResource.isSynchronized return true. But isSynchronized() on FileDocumentProvider will be false.. So I've to trigger it and don't care about the result.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
mehmet-karaman: IResource.isSynchornized() doesn't have the same understanding of FileDocumentProvider.checkSynchronizationState() of synchronzied.. If XtextDocumentProvider.isSynchronized is called multiple times, IResource.isSynchronized() will be for the first time unsynchron (return false) and for the next calls it will IResource.isSynchronized return true. But isSynchronized() on FileDocumentProvider will be false.. So I've to trigger it and don't care about the result.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/XtextEditor.java
mehmet-karaman: done.

mehmet-karaman:
I am going to rebase against the latest master.

org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
mehmet-karaman: There is one test which is executed automatically, which tests the basic functionality.. The ignored test is testing the case when the editor is dirty and the file is out of sync case.. I didn't saw an example how to automate a click on a button, without introducing a new dependency.

iloveeclipse:
@szarnekow, @cdietrich : anything is missing here? Could we continue with review or merge?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
```diff
@@ -419,7 +428,15 @@ public boolean isSynchronized(Object element) {
 		} else {
 			return super.isSynchronized(element);
 		}
-		return synchronizationStamp == getModificationStamp(element);
+
+		// Check has to fit the following method: 
+		// org.eclipse.ui.editors.text.FileDocumentProvider.checkSynchronizationState(long, IResource)
+		boolean isSynchronized = synchronizationStamp == getSynchronizationStamp(input.getFile());
```
szarnekow, line 434: I think I don't understand the idea behind this change.
Line 425 takes the modificationStamp and stores it as a local `synchronizationStamp`. Now we try to read `getSynchronizationStamp(someIFile)` which does
```
if (element instanceof IFileEditorInput) {
	FileInfo info= (FileInfo) getElementInfo(element);
	if (info != null)
		return info.fModificationStamp;
}
return super.getSynchronizationStamp(element);
```

it'll always go into the super.getSynchronizationStamp() which returns 0.
Something doesn't really work for me with this code.
Did you mean to call `computeModificationStamp` instead of getSynchronizationStamp.

If so: There must be something wrong with the tests. They should have caught this.
If not: Please explain (possibly again) in full detail what the idea is.

mehmet-karaman: You're right.. It could never be synchron, the file was reloaded usually. I guess I've to to store the time when the file was loaded and reloaded and compare with isSynchron.. Otherwise it will just compare the real files modification time stamp and the IResource modification timestamp.. But in this case the editor content has to be synchron, not just the file. 

I will add a few more asserts to check for isSynchronized() true/false, then we should be more secure with the tests.

mehmet-karaman: I've checked why this override was introduced, but then it doesn't makes a sense to keep this override. 
It was introduced to fix the the saving of files which are out of sync: https://bugs.eclipse.org/bugs/show_bug.cgi?id=343894. So If i remove this override, we will have the same problem back, but if I keep my changes in the XtextEditor class and call performSave(true, progressMonitor) .. (example from TextEditor which also forces the save) 

So might be someone has fixed the first bug in a wrong place and caused the unsync content problem.

mehmet-karaman: Removed the isSynchronized() overriden method, adjusted the performSave to force override .. Added more assertions to test.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/XtextEditor.java
```diff
@@ -327,6 +329,12 @@ public void doSave(IProgressMonitor progressMonitor) {
 		callback.afterSave(this);
 	}
 	
+	@Override
+	protected void performSave(boolean overwrite, IProgressMonitor progressMonitor) {
+		super.performSave(true, progressMonitor);
```
szarnekow, line 334: This is not something we can merge. Always enforcing an overwrite seems to be the wrong attack vector for the problem at hand.

mehmet-karaman: In comparison to the functionality of the org.eclipse.ui.editors.text.TextFileDocumentProvider we see that they force overwrite in the method org.eclipse.ui.editors.text.TextFileDocumentProvider.createFileFromDocument(IProgressMonitor, IFile, IDocument) . 

Also JDT forces the write if the file is synchronized.. 
org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.commitWorkingCopy(IProgressMonitor, Object, CompilationUnitInfo, boolean)

I've adjusted my solution similar to JDT .. I hope this is fine now.

memin20:
I am now on my private MacBook and try to understand why it fails constantly on macOS but works on windows and ubuntu..

mehmet-karaman:
Added waitForBuild to solve this async problem on Mac.. Seems that the build takes longer on mac machines .. Locally it worked already.

org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
iloveeclipse: @szarnekow : also the manual test is green without changes?

memin20: For this the ignored test has to be unignored and executed manually..

org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
```diff
@@ -116,7 +124,72 @@ public void testUndoRedo() throws Exception {
...
+		
+		assertTrue(fileDocumentProvider.isSynchronized(editor.getEditorInput()));
+		
+		assertEquals(document.get(), "");
+	}
 
+	/**
+	 * This test needs a manual button click.
+	 */
+	@Ignore
```
szarnekow, line 164: When I get this dialog:

Screenshot 2024-09-04 at 08 52 51

I'd expect that `Replace editor content` sets the editor content to the text in the file. But apparently it'll override the file change on disk. That's not ok.

memin20: really? I can't test It yet, as far as i remember it discards the changes in the editor.. Its not supposed to write anything on the disc..

memin20: Btw. You have to ignore the changes and then the changes in XtextEditor is necessary..

szarnekow: > really? I can't test It yet, as far as i remember it discards the changes in the editor.. Its not supposed to write anything on the disc..

Yes, I tried this a couple of times and added additional assertions to confirm my observation.

szarnekow: Might be that this is due to the doSave() though. Testing again.

szarnekow: Ok, test works as expected, indeed.

mehmet-karaman: Thanks for clarification.. Is this PR ok now, or is there something else what should be checked?

mehmet-karaman: @szarnekow is it possible to merge this PR today? :)

szarnekow: Yes, it is. @iloveeclipse can you confirm that it works for you guys?

mehmet-karaman: It works for me..

iloveeclipse: > Yes, it is. @iloveeclipse can you confirm that it works for you guys?

Mehmet is in my team, and thanks for asking.

szarnekow: Excellent. So let's move this forward. Thank you for the fix and the patience with the review process @mehmet-karaman

mehmet-karaman: @szarnekow Thank you for your review and yours hints..

Issue #3110, opened 2024-07-25
NPE is thrown, when editor is closed and the worker thread asks for the document


mehmet-karaman:
Hello,

in a project where code minings are active, there is a NPE thrown when editor is closed and a worker thread asks for the document.

cdietrich:
Can you provide a stacktrace

mehmet-karaman:
Sorry, I have fixed this bug already. Just wanted to open a ticket to refer to it before forgetting where it belongs to..

In the following method org.eclipse.xtext.ui.codemining.AbstractXtextCodeMiningProvider.provideCodeMinings(ITextViewer, IProgressMonitor) the xtextDocumentUtil.getXtextDocument(viewer) is returning in the case null and that has to be checked.

mehmet-karaman:
java.lang.NullPointerException: Cannot invoke "org.eclipse.xtext.ui.editor.model.IXtextDocument.tryReadOnly(org.eclipse.xtext.util.concurrent.IUnitOfWork, java.util.function.Supplier)" because the return value of "org.eclipse.xtext.ui.editor.model.XtextDocumentUtil.getXtextDocument(org.eclipse.jface.text.ITextViewer)" is null
	at org.eclipse.xtext.ui.codemining.AbstractXtextCodeMiningProvider.lambda$0(AbstractXtextCodeMiningProvider.java:90)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)

cdietrich:
fixed in 2.36

PR #3111, opened 2024-07-25
Fixed NPE when code minings are enabled and the editor closed

mehmet-karaman:
The NPE is happening when the Editor is closed and the worker thread is trying to access the document object, which is not existing anymore.

see https://github.com/eclipse/xtext/issues/3110


cdietrich (approved):

cdietrich:
thx @mehmet-karaman

PR #3115, opened 2024-07-31
Update minimum Supported Java version in Xbase and Xtend to Java 8

cdietrich:


cdietrich:
from my pov we are good to go. wdyt @szarnekow

cdietrich:
am not sure if we need to do something like
org.eclipse.xtext.ui.testing.util.JavaProjectSetupUtil.makeJava7Default()

szarnekow:
> org.eclipse.xtext.ui.testing.util.JavaProjectSetupUtil.makeJava7Default()

Wow that's annoying logic

szarnekow:
But luckily it only cloaks a Java7 JRE as a Java6 JRE

LorenzoBettini (approved):
@cdietrich @szarnekow thank you both for updating all these tests!

PR #3116, opened 2024-08-02
Add some documentation about IDerivedStateComputer

tom-fonda:
Hi,

Lately I have found this interface quite useful in the project I am working on, but it was not trivial to, first of all, learn of its existence, and then figure out how exactly to implement it. Thus, I thought of adding a dedicated section to the online documentation.

Still, there are a few things I am not sure about, so I would appreciate some feedback about the following:

- the section title;
- the usage of the `preLinkingPhase` parameter of `installDerivedState()`. Both in this example and in my project I've simply ignored it. I am unsure what could be a scenario in which considering it is a hard requirement;
- the provided `discardDerivedState()` implementation is very naive and does not check if an Entity's name has been manually set by the user to that of the overridden Entity plus the suffix. Preserving the name field is not strictly necessary in this case, but it would probably be more elegant to introduce a collection as a class member field, add all Entities processed by `installDerivedState()` to this collection, and then have `discardDerivedState()` only operate on the Entities in the collection.

Of course, any other feedback is welcome!
Thanks.

### DIFF
--- a/org.eclipse.xtext.doc/contents/303_runtime_concepts.html
+++ b/org.eclipse.xtext.doc/contents/303_runtime_concepts.html
@@ -638,6 +638,67 @@ terminal Plural:
 }
 </code></pre>
 
+<h2 id="tree-manipulation">Tree manipulation</h2>
+
+<p>An implementation of <a href="https://github.com/eclipse/xtext/blob/main/org.eclipse.xtext/src/org/eclipse/xtext/resource/IDerivedStateComputer.java">IDerivedStateComputer</a> can be used to modify the objects in the abstract syntax tree after parsing. Commonly, implementations of this interface are used to modify the objects' fields programmatically.</p>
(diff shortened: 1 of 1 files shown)


cdietrich:
thx @tom-fonda 
please note. this html is derived from
https://github.com/eclipse/xtext-website/blob/main/xtext-website/documentation/303_runtime_concepts.md

tom-fonda:
> thx @tom-fonda please note. this html is derived from [eclipse/xtext-website@`main`/xtext-website/documentation/303_runtime_concepts.md](https://github.com/eclipse/xtext-website/blob/main/xtext-website/documentation/303_runtime_concepts.md?rgh-link-date=2024-08-02T08%3A33%3A46Z)

Oh, I had totally missed this! I'm closing this PR and will soon open a new one over there.

Issue #2853, opened 2023-11-22
xtext clean output - not removing all aoutput folders, if previous does not exists


marcinjunger:
Hi,
I would like to report a possible bug. Please take a look at line 490 in org.eclipse.xtext.builder.BuilderParticipant -> cleanOutput(IBuildContext, OutputConfiguration, EclipseResourceFileSystemAccess2, IProgressMonitor)

https://github.com/eclipse/xtext/blob/cab7c302fcdc9b8975d58b12061da06c12ec369c/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java#L490

This method removes output folders based on configuration: the general logic is:
for (IContainer container : getOutputs(project, config)) { 
  delete the output
}

The problem is with the following lines:
		for (IContainer container : getOutputs(project, config)) {
			if (!container.exists()) {
				return;
			}

It stops deleting on the first non-existing resource to be deleted. So, if getOuputs returns two folders, and oth exists - both will be deleted. If the first one does not exists, but the second does - it will not be deleted.

I propose to change "return" into "continue" - this should fix the issue.
Kind regards
Marcin

szarnekow:
Thank you for the analysis. Do you plan to create a pull request? That would be greatly appreciated.

marcinjunger:
Hi, no problem, tried to search for a guide how to do it, naming conventions etc... is there any? Should I check out the git repo (I am using sourcetree) locally, create locally a branch named eg. bugfix\issue2853, commit and push a fix, and then create pull request in sourcetree?

tripleo1:
That sounds like a plan.

LorenzoBettini:
> Hi, no problem, tried to search for a guide how to do it, naming conventions etc... is there any? Should I check out the git repo (I am using sourcetree) locally, create locally a branch named eg. bugfix\issue2853, commit and push a fix, and then create pull request in sourcetree?

@marcinjunger you might want to look at https://github.com/eclipse/xtext/blob/main/CONTRIBUTING.md in particular how to setup your Eclipse workspace (we use Oomph)

PR #3117, opened 2024-08-02
Fix: Ensure xtext cleans all output folders even if previous output is missing

dani0370:
The issue for this fix can be found here: [#2853]


cdietrich (approved):

cdietrich:
thx @dani0370

PR #3120, opened 2024-08-02
Optimize performance of text replacer context with autowrap and already formatted document

### LINKED ISSUE #3119: [formatting2] Performance issues with autowrap and already formatted document
There are performance issues when formatting an already properly formatted document in conjunction with autowrap.

When autowrap was enabled once in the current line, `FormattableDocument.createReplacements(...)` then checks for each replacer if it `needsAutowrap(...)`. This calls `ITextReplacerContext.getLeadingCharsInLineCount()` to determine if the line has exceeded the maximum line width.

`TextReplacerContext.getLeadingCharsInLineCount()` then iterates over all previous contexts until it finds a newline character in the respective replacements. However, if there are no replacements made because the document is already properly formatted, this iterates over all previous contexts until the beginning of the document.

There are also many replacer contexts created using `TextReplacerContext.withDocument(...)`, `withIndentation(...)`, and `withReplacer(...)`, where the resulting context is identical to the current one.

In summary, once autowrap is enabled, this may iterate over all replacer contexts for each formatter in the current line. This can be millions of iterations in total for a large document.

mx990:
When calculating the leading chars in a line, the replacer context iterates over all previous contexts if there are no replacements. This PR also uses the region of the associated replacer to find the beginning of the current line.

Furthermore, it avoids unnecessary text replacer contexts by returning the current context when the resulting context would be identical.

Closes #3119

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/TextReplacerContext.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/formatting2/internal/TextReplacerContext.java
@@ -131,6 +131,19 @@ public class TextReplacerContext implements ITextReplacerContext {
 				count += logicalLength(text);
 				lastOffset = rep.getOffset();
 			}
+			final ITextReplacer replacer = current.getReplacer();
+			if (replacer != null) {
+				final int offset = replacer.getRegion().getOffset();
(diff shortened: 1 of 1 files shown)


cdietrich:
thx

PR #3122, opened 2024-08-05
mitigate pot. prototype pollution problem

cdietrich:


org.eclipse.xtext.web.example.jetty/src/main/webapp/i18n.js
```diff
@@ -74,7 +74,10 @@
     function mixin(target, source, force) {
         var prop;
         for (prop in source) {
-            if (source.hasOwnProperty(prop) && (!target.hasOwnProperty(prop) || force)) {
+            if (prop === '__proto__' || prop === 'constructor') {
```
szarnekow, line 77: Looks reasonable to me.

PR #3128, opened 2024-08-08
use method handles for reflection

cdietrich:


merks (approved):
This looks very nice and will perform better.

szarnekow:
I think we can simplify this a bit more:
```
	private final static MethodHandle GET_ROOT_PATH_TO_RESOLVED_ENTRIES = findAccessor();
	
	private static MethodHandle findAccessor() {
		try {
			if (JavaCore.getPlugin().getBundle().getVersion().compareTo(new Version(3, 39, 0)) >= 0) {
				return MethodHandles.lookup().findGetter(PerProjectInfo.class, "rootPathToResolvedEntries", Map.class);
			} else {
				return MethodHandles.lookup().findVirtual(PerProjectInfo.class, "getRootPathToResolvedEntries", MethodType.methodType(Map.class));
		}
		} catch (ReflectiveOperationException e) {
			
			return MethodHandles.constant(Map.class, Collections.emptyMap());
		}
	}
	
	protected Map<IPath, IClasspathEntry> getRootPathToResolvedEntries(PerProjectInfo info) throws JavaModelException {
		try {
			return (Map<IPath, IClasspathEntry>) GET_ROOT_PATH_TO_RESOLVED_ENTRIES.invoke(info);
		} catch(Throwable t) {
			throw new JavaModelException(Status.error(t.getMessage(), t));
		}
	}
```

I'll push to this branch. WDYT?

cdietrich:
@szarnekow do you think we still will be able to detect future breakages?
via AccessRestrictionInWorkspaceTest?

cdietrich:
order was wrong. have reversed

szarnekow:
> via AccessRestrictionInWorkspaceTest?

Yes, that'll certainly break if things change in an incompatible way.

szarnekow (approved):

Issue #2457, opened 2020-03-24
ReplaceFileContentChange does not work together with a move change (UNDO)


cdietrich:
Doing a Move Refactoring together with a change serializer based refactoring on an emf resource (non text resource) then the created `ReplaceFileContentChange` UNDO wont work as it keeps only one file (the old one) and thus `org.eclipse.xtext.ui.refactoring2.ReplaceFileContentChange.isValid(IProgressMonitor)` will fail on the undo

(am not sure if this is a legit usecase but we were able to run into this usecase)

also the order of the move and ReplaceFileContentChange on do/undo seems different.

frankbenoit:
I also run now in this issue, when during a directory rename a contained resource needs to have content be changed via a change serializer.
The `isValid` tests if the file is present. Is it possible to skip this check during undo?

cdietrich:
i have no idea.

frankbenoit:
To be more precise... would it be acceptable to have a change/PR for `ReplaceFileContentChange` that will ignore the check during undo?

cdietrich:
i cannot really judge. maybe overload contructors with opt-in into that new behaviour.

PR #3130, opened 2024-08-08
#2457 Disable verify in ReplaceFileContentChange for UNDO

frankbenoit:
The verify is executed before the any change is performed. When the verify checks for the existance of files beforehand, this might not work for UNDO when there are move/rename is involved. Hence skipping the check for UNDO.

Signed-off-by: frank.rene.benoit@gmail.com

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/ChangeConverter.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/ChangeConverter.java
@@ -130,7 +130,7 @@ public class ChangeConverter implements IAcceptor<IEmfResourceChange> {
 				change.getResource().save(outputStream, null);
 				byte[] newContent = outputStream.toByteArray();
 				String lastSegment = change.getOldURI().lastSegment();
-				ReplaceFileContentChange ltkChange = new ReplaceFileContentChange(lastSegment, file, newContent);
+				ReplaceFileContentChange ltkChange = ReplaceFileContentChange.createWithSkippingUndoVerify(lastSegment, file, newContent);
 				addChange(ltkChange);
 			} catch (IOException e) {
 				Exceptions.throwUncheckedException(e);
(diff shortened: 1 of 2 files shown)


cdietrich:
@iloveeclipse can you test this in your codebase
@szarnekow do you have an idea if this is a good idea.

iloveeclipse:
> can you test this in your codebase

@mehmet-karaman : may be you can run some of our DSL tests that might rely on this code from IDE with the change here and se if anything breaks (assuming you could identify right test cases)?

@cdietrich : running all automated tests from command line on a "manually" patched code is not possible because they need deployed binaries to run on multiple workstations in parallel. Our test time is > 24 hours if executed on a single host as of today.

We run tests every week with latest platform & Xtext nightly builds, so should there be a regression, we will investigate & report it.

cdietrich:
@frankbenoit can you please rebase

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/ChangeConverter.java
```diff
@@ -130,7 +130,7 @@ protected void _handleReplacements(IEmfResourceChange change) {
 				change.getResource().save(outputStream, null);
 				byte[] newContent = outputStream.toByteArray();
 				String lastSegment = change.getOldURI().lastSegment();
-				ReplaceFileContentChange ltkChange = new ReplaceFileContentChange(lastSegment, file, newContent);
+				ReplaceFileContentChange ltkChange = ReplaceFileContentChange.createWithSkippingUndoVerify(lastSegment, file, newContent);
```
cdietrich, line 133: wonder if this should go into a methid so that it can be more easily cusomized.
@szarnekow @LorenzoBettini wdyt?

LorenzoBettini: Yes. In that case some javadoc should explain it.

frankbenoit: You mean an overridable method?
````
protected Change createLtkChange(name, file, content){ ....
````

LorenzoBettini:
@frankbenoit why has this been closed?

frankbenoit:
Sorry, I have some proxy/git problems, so I deleted my fork and during recreating it, still issues. I will create a fresh PR the next days

frankbenoit:
Please continue in https://github.com/eclipse/xtext/pull/3134

PR #3133, opened 2024-08-09
Fix uninitialized target field in RenameElementProcessor2

### LINKED ISSUE #3132: [refactoring2] Uninitialized target field in RenameElementProcessor2
The `target` field in `RenameElementProcessor2` is not initialized, since it is shadowed by a local variable in `initialize(IRenameElementContext)`.

As a result, the `target` passed to `IRenameNameValidator.validate(...)` in `validateNewName(String)` is always `null`.

mx990:
The `target` field in `RenameElementProcessor2` is not initialized, since it is shadowed by a local variable in `initialize(IRenameElementContext)`. As a result, the target passed to `IRenameNameValidator.validate(...)` in `validateNewName(String)` is always `null`.

This commit removes the local variable, so that the field is properly initialized instead.

Closes #3132


LorenzoBettini:
I think this can be accepted, @szarnekow wdyt?

It's kind of strange no test detected the problem, but maybe this part is not covered much.

cdietrich:
we just merge. users can always complain afterwards ....

cdietrich:
thx @mx990

PR #3134, opened 2024-08-09
#2457 Disable verify in ReplaceFileContentChange for UNDO

frankbenoit:
The verify is executed before the any change is performed. When the verify checks for the existance of files beforehand, this might not work for UNDO when there are move/rename is involved. Hence skipping the check for UNDO.

Signed-off-by: frank.rene.benoit@gmail.com

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/ReplaceFileContentChange.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/ReplaceFileContentChange.java
@@ -38,14 +38,27 @@ public class ReplaceFileContentChange extends ResourceChange {
 
 	private String name;
 
+	private final VerifyMode verifyMode;
+
+	protected enum VerifyMode {
+		BOTH,
+		NOT_FOR_UNDO_NOW_FWD,
+		NOT_FOR_UNDO_NOW_UNDO
+	}
+
 	public ReplaceFileContentChange(String name, IFile file, byte[] newContents) {
-		this(file, newContents);
-		this.name = name;
+		this(name, file, newContents, VerifyMode.BOTH);
 	}
 
 	public ReplaceFileContentChange(IFile file, byte[] newContents) {
+		this(null, file, newContents, VerifyMode.BOTH);
+	}
+
+	protected ReplaceFileContentChange(String name, IFile file, byte[] newContents, VerifyMode verifyMode) {
+		this.name = name;
 		this.file = file;
 		this.newContents = newContents;
+		this.verifyMode = verifyMode;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


LorenzoBettini:
@cdietrich I never dealt with this refactoring code, so I really have no clue on how this is supposed to work. Moreover, without tests it's hard to see if and how this works correctly. Maybe @szarnekow can judge better.

cdietrich:
I have no longe access to real world projects so I cannot test neither.
Maybe we just merge and wait formpeople tomcomplain 9 months later

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/ReplaceFileContentChange.java
```diff
@@ -38,6 +38,13 @@ public class ReplaceFileContentChange extends ResourceChange {
 
 	private String name;
```
cdietrich, line 39: This class does not really seem extensible

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/ReplaceFileContentChange.java
```diff
@@ -38,6 +38,13 @@ public class ReplaceFileContentChange extends ResourceChange {
...
 
+	private VerifyMode verifyMode = VerifyMode.BOTH;
```
cdietrich, line 41: Should this be more accessible ?

frankbenoit: Would having the 3 arg ctor and `getNextUndoVerifyMode()` to be protected be sufficient?

cdietrich: Problem is have no time to try out what a possible subclasser would do

frankbenoit: Users can change the instantiation point, and can use the old ctor to get the old behavior. 
Isn't that sufficient for this PR?

cdietrich: As I have said. I don’t have any time to try it actually out . Maybe someone else has

cdietrich: i propose we just merge and fix when users complain....

cdietrich:
thx @frankbenoit

Issue #3127, opened 2024-08-07
Investigate if we can move xtext and mwe to  apache.commons.commons-logging 1.3.3


cdietrich:
starter for mwe https://github.com/eclipse/mwe/tree/cd-commons-logging

for xtext
- targets + build properties (also in wizard)
- maybe some remainder require bundles that are not import package yet
- ....

see also https://github.com/eclipse-simrel/simrel.build/pull/482

HannesWell:
For MWE it looks doable to me to migrate off commons.logging (plus log4j) and just use for example slf4j:
- https://github.com/eclipse/mwe/pull/296

I have not yet checked if something similar can be done for xtext too, but maybe it's also possible.

In general instead of slf4j log4j-2 could be used as well. I have no strong preference and I guess for the usage in MWE/Xtext both have the same capabilities and both APIs cna be redirected if necessary.

PR #3137, opened 2024-08-09
Migrate to latest commons-logging (currently 1.3.3)

HannesWell:
Adapt to the new Bundle-SymbolicName in target-files, the `additional.bundles` required for development and defined in `build.properties` and adjust the templates to generate the new name.

Should fix https://github.com/eclipse/xtext/issues/3127 for Xtext.

### DIFF
--- a/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
+++ b/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
@@ -113,7 +113,7 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
 			"org.eclipse.emf.mwe2.launch",
 			"org.eclipse.emf.mwe2.lib",
 			"org.objectweb.asm",
-			"org.apache.commons.logging", 
+			"org.apache.commons.commons-logging",
 			"org.apache.log4j"
 		)
 		if (isFromExistingEcoreModels) {
--- a/org.eclipse.xtext.ide.tests/build.properties
+++ b/org.eclipse.xtext.ide.tests/build.properties
@@ -8,4 +8,4 @@ source.. = src-gen,\
            testlang-src-gen
 src.includes = about.html
 additional.bundles = org.eclipse.emf.mwe2.launch,\
-                     org.apache.commons.logging
+                     org.apache.commons.commons-logging
--- a/org.eclipse.xtext.purexbase/build.properties
+++ b/org.eclipse.xtext.purexbase/build.properties
@@ -16,7 +16,7 @@ additional.bundles = org.eclipse.xtext.xbase,\
                      org.eclipse.emf.mwe2.launch,\
                      org.eclipse.emf.mwe2.lib,\
                      org.objectweb.asm,\
-                     org.apache.commons.logging,\
+                     org.apache.commons.commons-logging,\
                      org.apache.log4j
 
 
--- a/org.eclipse.xtext.testlanguages/build.properties
+++ b/org.eclipse.xtext.testlanguages/build.properties
@@ -9,4 +9,4 @@ src.includes = about.html
 additional.bundles = org.eclipse.emf.mwe.utils,\
                      org.eclipse.emf.mwe2.launch,\
                      org.eclipse.emf.mwe2.lib,\
-                     org.apache.commons.logging
+                     org.apache.commons.commons-logging
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/eclipsePlugin/eclipsePlugin/build.properties
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/eclipsePlugin/eclipsePlugin/build.properties
@@ -14,5 +14,5 @@ additional.bundles = org.eclipse.xtext.xbase,\
                      org.eclipse.emf.mwe2.launch,\
                      org.eclipse.emf.mwe2.lib,\
                      org.objectweb.asm,\
-                     org.apache.commons.logging,\
+                     org.apache.commons.commons-logging,\
                      org.apache.log4j
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/eclipsePluginP2/eclipsePluginP2/build.properties
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/eclipsePluginP2/eclipsePluginP2/build.properties
@@ -14,5 +14,5 @@ additional.bundles = org.eclipse.xtext.xbase,\
                      org.eclipse.emf.mwe2.launch,\
                      org.eclipse.emf.mwe2.lib,\
                      org.objectweb.asm,\
-                     org.apache.commons.logging,\
+                     org.apache.commons.commons-logging,\
                      org.apache.log4j
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/full/full.parent/full.target/full.target.target
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/full/full.parent/full.target/full.target.target
@@ -26,7 +26,7 @@
 			<unit id="org.junit" version="4.13.2.v20230809-1000"/>
 			<unit id="org.hamcrest" version="2.2.0"/>
 			<unit id="org.hamcrest.core" version="2.2.0.v20230809-1000"/>
-			<unit id="org.apache.commons.logging" version="1.2.0"/>
+			<unit id="org.apache.commons.commons-logging" version="0.0.0"/>
 			<unit id="org.objectweb.asm" version="9.7.0"/>
 			<unit id="io.github.classgraph.classgraph" version="4.8.174"/>
 			<repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2024-09"/>
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/full/full.parent/full/build.properties
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/full/full.parent/full/build.properties
@@ -14,5 +14,5 @@ additional.bundles = org.eclipse.xtext.xbase,\
                      org.eclipse.emf.mwe2.launch,\
                      org.eclipse.emf.mwe2.lib,\
                      org.objectweb.asm,\
-                     org.apache.commons.logging,\
+                     org.apache.commons.commons-logging,\
                      org.apache.log4j
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/lsMavenTychoApp/lsMavenTychoApp.parent/lsMavenTychoApp.target/lsMavenTychoApp.target.target
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/lsMavenTychoApp/lsMavenTychoApp.parent/lsMavenTychoApp.target/lsMavenTychoApp.target.target
@@ -26,7 +26,7 @@
 			<unit id="org.junit" version="4.13.2.v20230809-1000"/>
 			<unit id="org.hamcrest" version="2.2.0"/>
 			<unit id="org.hamcrest.core" version="2.2.0.v20230809-1000"/>
-			<unit id="org.apache.commons.logging" version="1.2.0"/>
+			<unit id="org.apache.commons.commons-logging" version="0.0.0"/>
 			<unit id="org.objectweb.asm" version="9.7.0"/>
 			<unit id="io.github.classgraph.classgraph" version="4.8.174"/>
 			<repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2024-09"/>
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/lsMavenTychoApp/lsMavenTychoApp.parent/lsMavenTychoApp/build.properties
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/lsMavenTychoApp/lsMavenTychoApp.parent/lsMavenTychoApp/build.properties
@@ -14,5 +14,5 @@ additional.bundles = org.eclipse.xtext.xbase,\
                      org.eclipse.emf.mwe2.launch,\
                      org.eclipse.emf.mwe2.lib,\
                      org.objectweb.asm,\
-                     org.apache.commons.logging,\
+                     org.apache.commons.commons-logging,\
                      org.apache.log4j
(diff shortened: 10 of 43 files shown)


cdietrich:
there is still the copy-dependencies the wizard does in some configs that mentions the old name in excludeArtifactIds (IdeProjectDescriptor.xtend)
we also need to check into the failing build.

as my team at work in decimated over the next weeks and after that i am on vacation myself i am not sure when we should bring it in. either immediately or right after the release

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
```diff
@@ -32,46 +32,46 @@ class RuntimeProjectDescriptor extends TestedProjectDescriptor {
...
 		testProject = new RuntimeTestProjectDescriptor(this)
 	}
-	
+
 	override setEnabled(boolean enabled) {
 		if (!enabled)
 			throw new IllegalArgumentException("The runtime project is always enabled")
 		super.enabled = enabled
 	}
-
```
cdietrich, line 47: can you undo the ws changes.

cdietrich: (you can run cliwizardintegrationtest as java main to update test expectations

HannesWell:
> there is still the copy-dependencies the wizard does in some configs that mentions the old name in excludeArtifactIds (IdeProjectDescriptor.xtend)

Yes I saw these (templates for) pom.xml files, but since it is a configuration for the maven-dependency-plugin and the element is named `excludeArtifactIds` I'm not sure if the current name is even correct?
To me it looks like that should be the Maven artifactId which would be `commons-logging`:
https://mvnrepository.com/artifact/commons-logging/commons-logging

Or does this mojo also consider the PDE dependencies injected by Tycho?
In general I have to admit that I have not yet understood why the dependencies are copied at all.

> AbstractPluginProjectCreator also mentions it

That's an import-package header value (at least the method name implies it:
https://github.com/eclipse/xtext/blob/2bce8b6c62b1da7c57d81cff6e699eedb19d90a9/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/wizard/AbstractPluginProjectCreator.java#L58-L62


> we also need to check into the failing build.

I'm looking into this.

> as my team at work in decimated over the next weeks and after that i am on vacation myself i am not sure when we should bring it in. either immediately or right after the release

I think it also depends on how fast MWE2 is adjusted, but the commit on your branch looks good, I just have a one or two things that are missing. Do you want to create a PR in MWE2 for that? Then I can add my remarks.

Or if we have something like suggested in https://github.com/eclipse/mwe/pull/296#issuecomment-2278867002, we can maybe completely remove the `additional.bundles` entries in the `build.properties`.

cdietrich:
```
	«ENDIF»
						«IF config.languageServer!=LanguageServer.NONE»
							«IF isEclipsePluginProject»
```

so this seems to be if you select tycho and language server

cdietrich:
about

additional.bundles

this wont work if you just create a project without any tycho. 

unfortunately there is no 
additional.imports ....

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/RuntimeProjectDescriptor.xtend
HannesWell: I cannot reproduce the test failure locally. And if I read the error message correct it looks like there is just an additional space between square brackets but I don't understand how this change influences that.
But I pushed an update where I removed the whitepace as requested.

... While typing this and thinking again. These whitespace changes probably caused the test-failures so they are hopefully fixed now.

HannesWell:
> as my team at work in decimated over the next weeks and after that i am on vacation myself i am not sure when we should bring it in. either immediately or right after the release

Thinking about this again. Since I assume that mwe workflows will work with the old an new version of commons-logging (since it is just a plain old flat-classpath java app) I think this change is quite save. We don't even have to await the corresponding change for mwe2 since one could already now change existing the 'devlopment-bundles' of an xtext project and the workflow should still run.

So once the execludedArtifacts in pom(templates) are correct, I'm for submitting this now.

HannesWell:
@cdietrich regarding the corresponding change in MWE2 are you fine if I take your commit to a separate branch add my changes and create a PR from it?

cdietrich:
Yes. What are the additions?

cdietrich:
The idea would be

merge mwe 
Mwe milestone m2
Merge this
Create mwe bump to m2
Merge it 
Check if running workflow in Xtext dev env works
Update reference projects
Merge them
Wait for fallout and fix

HannesWell:
> Yes. What are the additions?

See https://github.com/eclipse/mwe/pull/298.

> The idea would be

As you prefer. When is the mwe m2 planned? Couldn't we test it with a nightly build? This would make the cycle shorter.

cdietrich:
I can do mwe m2 it later today when home from hiking (there is no planned date as so far no changes )

cdietrich:
Maybe you can build a branch of Xtext/xtext-reference-projeczs with your mwe and Xtext build results but I fear doing all the adjustments will be a lot

cdietrich:
Do you have some capacity to investigate fallout if there is any contra to my expectations

HannesWell:
> Maybe you can build a branch of Xtext/xtext-reference-projeczs with your mwe and Xtext build results but I fear doing all the adjustments will be a lot

Which project's are you exactly referring to when you say `Xtext/xtext-reference-projeczs` ? The examples in this repository like `org.eclipse.xtext.example.arithmetics` or `org.eclipse.xtext.example.domainmodel`?

> Do you have some capacity to investigate fallout if there is any contra to my expectations

It depends on when the fallout will happen and what the time-frame is to fix it.
This weekend I would have some more time, next week a bit less, but there should still be spots in the evening.

> I can do mwe m2 it later today when home from hiking (there is no planned date as so far no changes )

Enjoy your hike :)

cdietrich:
Github.com/xtext/xtext-reference-projects

there is launch to regenerate wizard projecte 
Domain model needs manual copy paste 
Integratintests/ (swt) might need target adoption

cdietrich:
@HannesWell i am back home. 
if you are d'accord i'd propose to merge the MWE and then build a MWE milestone

HannesWell:
Yes I'm fine with that. 
I have limited availability in the next 2-3 hours but will be fully available in the afternoon/evening.

cdietrich:
https://github.com/eclipse/xtext/pull/3141

cdietrich:
Mwe bump build is still running. Will merge once it is through

cdietrich:
@HannesWell mwe bump is in. Can you please rebase

HannesWell:
> @HannesWell mwe bump is in. Can you please rebase

Done.

Thanks for the milestone and adjustments!

cdietrich:
ok. will create a branch for reference projects

cdietrich:
Regarding the IdeProjectDescriptor.
maybe "viel hilft viel" and adding both might be an option.
i also see this in domainmodel.ide

cdietrich:
https://github.com/xtext/xtext-reference-projects/pull/435

cdietrich:
hmmmm https://github.com/xtext/xtext-reference-projects/actions/runs/10332344902/job/28603464575

cdietrich:
https://github.com/xtext/xtext-reference-projects/blob/25cdadea4af21df72378ac3783f59039ff223f08/domainmodel/2.36.0/org.eclipse.xtext.example.domainmodel.releng/tp/domainmodel.target#L28

seems to help but of course is very meh...

HannesWell:
Sorry installing my 'Balcony power plant' took longer than expected, but I'm back now.

> do we have to add it to extra requirements? shouldnt tycho resolve it anyway?
> 
> https://github.com/xtext/xtext-reference-projects/blob/5718762516bc29e1af3374f4de322bc806cd9407/domainmodel/2.36.0/org.eclipse.xtext.example.domainmodel.releng/pom.xml#L110 does not help neither

That should not be necessary at all. Tycho should resolve the `additional.bundles` as dependencies too.

Are you sure the target-platform in use contains the new commons-logging artifact?

cdietrich:
it does but not explicitely. we removed it (explicit listings from domain model) with
https://github.com/eclipse/xtext-eclipse/issues/1686
https://github.com/eclipse/xtext-eclipse/issues/1994

now using planner instead of slicer

tycho bug?

cdietrich:
https://github.com/xtext/xtext-reference-projects/blob/5718762516bc29e1af3374f4de322bc806cd9407/domainmodel/2.36.0/org.eclipse.xtext.example.domainmodel.releng/pom.xml#L110
would partially revert that

HannesWell:
Ok the `domainmodel.target` contains the new `org.apache.commons.commons-logging` at the repos
- https://download.eclipse.org/modeling/emft/mwe/updates/milestones/S202408101130
- https://download.eclipse.org/releases/2024-09

The problem is, no IU listed in the TP seems to require the new version, but something pulls in the old `org.apache.commons.logging`. Probably some older feature or a bundle using require-bundle or a tight import-package version range? I have not yet investigated in detail.

So to solve this you can just add the corresponding IU to one of the locations. E.g.
```
      <unit id="org.apache.commons.commons-logging"/>
      <repository location="https://download.eclipse.org/releases/2024-09"/>
```

A separate location is not necessary.

> tycho bug?

With the explanation above, I don't think so. Problem is that in general, Tycho, like PDE only resolves requirements of IUs listed in the target, but not extra requirements of the projects being built. Oomph targlets are different in that sense. You only have to specify a list of repos and get the full transitive dependency closure in your TP of the Plug-in projects in your workspace.

HannesWell:
In general you can also use multiple repositories in one location. Under the hood PDE/Tycho uses the repos of all locations when resolving IUs using the planner. It can be useful if you view the content of your TP and sort it by location, but for the overall TP content there is no difference if you have one location with all your IUs and repo-URLs listed or one location per repo.

cdietrich:
ok can you adapt your pr?
with org.apache.commons.logging have no idea what could pull this.
dont see it in plugin dependencies view

HannesWell:
> with org.apache.commons.logging have no idea what could pull this.
> dont see it in plugin dependencies view


Looks like its apache-httpclient

grafik


> ok can you adapt your pr? 

Sorry I don't understand what has to be adapted in this PR?
I assumed we are talking about yours https://github.com/xtext/xtext-reference-projects/pull/435?

LorenzoBettini:
> why does the
> 
> ```
>     <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
>       
>       <repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2024-09"/>
>     </location>
> ``` no longer work ...
> ```

That breaks if Orbit provides different bundles like commons logging with different names and transitive dependencies kick in, like in this case.

I added that part some time ago; I see you already added the explicit dependency on the new `commons-logging`; maybe the comment has to be tweaked accordingly.

cdietrich:
@HannesWell 
org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/tp/domainmodel.target

cdietrich:
and what depends on jgit?

cdietrich:
so what is left is IdeProjectDescriptor

HannesWell:
> @HannesWell org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/tp/domainmodel.target

Indeed, thanks.

> and what depends on jgit?

In Xtext I believe nothing. My other guess is ECF on which P2 depends. I could invesitgate further, but in the end I think it is sufficient to assume that something requires the old commons.logging but nothing in the TP requires the new one.



> so what is left is IdeProjectDescriptor

Yes. The maven-dependency-plugin config.
I'll have a look at this.

Btw. the build of the previous state succeeded. 

> > why does the
> > ```
> >     <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
> >       
> >       <repository location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2024-09"/>
> >     </location>
> > ``` no longer work ...
> > ```
> 
> That breaks if Orbit provides different bundles like commons logging with different names and transitive dependencies kick in, like in this case.
> 
> I added that part some time ago; I see you already added the explicit dependency on the new `commons-logging`; maybe the comment has to be tweaked accordingly.

I think this is still correct. P2/PDE/Tycho resolves the latest possible version of all requirements. And as long as something requires the 'old' commons-logging, either because of using the old-name and Require-Bundle or Import-Package with a version below 1.3 that old version is resolved (too). It can also be that both versions are resolved. If one cannot satisfies all requirements.

It could of course also happen that the old one can serve all requirements and thus the new one is not resolved/used.

cdietrich:
jar tf org.xtext.example.mydsl.ide-1.0.0-SNAPSHOT-ls.jar  | grep logging
shows common logging is included in the uber jar

no idea how this works

=> propose to add the new bundle name
and also 
`<exclude>*:org.xtext.example.mydsl.ide-commons-logging*</exclude>`

LorenzoBettini:
> jar tf org.xtext.example.mydsl.ide-1.0.0-SNAPSHOT-ls.jar | grep logging shows common logging is included in the uber jar
> 
> no idea how this works
> 
> => propose to add the new bundle name and or commons-logging and also `<exclude>*:org.xtext.example.mydsl.ide-commons-logging*</exclude>` or something similar (seems not to be enhough)

I seem to understand that the exclusion has not been updated so the jar is in the uber jar for the moment.

cdietrich:
guess we need

```
<exclude>*:org.xtext.example.mydsl.ide-commons-logging*</exclude>
							<exclude>*:*commons-logging*</exclude>
```

and

```
<excludeArtifactIds>
								com.ibm.icu,
								org.apache.ant,
								org.apache.commons.lang,
								org.apache.commons.logging,
								org.apache.commons.commons-logging,
								commons-logging,
```

did not check the non shade variant yet

HannesWell:
The configuration already seems to be broken, to fix it I created a separate PR:
- https://github.com/eclipse/xtext/pull/3143

You can notice the problem when looking into the `<dsl-project>.ide/target/libs` folder into which the dependencies are copied.
It still happens that e.g. commons.logging ends up in the uber-jar, even if it is not in the libs folder. This is probably because the shade plugin finds it as a 'Maven'-Depenendency and it includes all of them in the uber-jar.
I assume the exclusion is only to avoid duplicates.

cdietrich (approved):

cdietrich:
thx @HannesWell

PR #3139, opened 2024-08-10
Run GH workflow build in batch-mode and fail at end

HannesWell:
Without the -B Maven CLI option a lot of print-out in the build logs of Github workflows is just transfer progress.

The `full-Build.sh` uses the same arguments for a build in Jenkins.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Build and test
         uses: coactions/setup-xvfb@v1.0.1
         with: 
-          run: ./mvnw clean verify "-Dmaven.home=${{ env.MAVEN_WRAPPER_HOME }}" -PuseJenkinsSnapshots ${{ matrix.additional-maven-args }} -f org.eclipse.xtext.full.releng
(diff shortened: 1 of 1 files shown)


HannesWell:
Btw. are you interested in having test-results posted as comment in a PR as it done for all Eclipse SDK projects.
You can see it in any PR e.g. of eclipse-platform:
https://github.com/eclipse-platform/eclipse.platform/pulls

cdietrich:
In general yes

LorenzoBettini:
@HannesWell, this will not remove download logging from Tycho, right?
I'd like to keep that. I seem to understand it doesn't remove Tycho logging, but just in case...

concerning test results that'd be great! :)

LorenzoBettini (approved):
@HannesWell, will you contribute the test report to another PR?

You haven't changed the Maven command in the other job `build-maven-artifacts` but maybe we can leave the other one as it is

HannesWell:
> @HannesWell, will you contribute the test report to another PR?
> 
Yes, will do that later this day.


> You haven't changed the Maven command in the other job `build-maven-artifacts` but maybe we can leave the other one as it is

Oh indeed, I oversaw this. Will push an update later this day too.

HannesWell:
> @HannesWell, this will not remove download logging from Tycho, right? I'd like to keep that. I seem to understand it doesn't remove Tycho logging, but just in case...
> 
I don't know that Tycho disables any logging based on this option and I could not found anything in this regard.

> You haven't changed the Maven command in the other job `build-maven-artifacts` but maybe we can leave the other one as it is

Just included that.

I'll create a separate PR for the test results.

PR #3140, opened 2024-08-10
Deal with Java 11, 17, and 21 for the Jenkins build

### LINKED ISSUE #3135: There is still something wrong with the build setup
https://ci.eclipse.org/xtext/job/xtext/job/main/1370/

@LorenzoBettini coukd you please have a look

LorenzoBettini:
and rely on the toolchain to force compilation and tests with other Java versions.

I left comments to document what I think it's the semantics of those Jenkinsfile parts

Closes #3135

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,9 +25,9 @@ pipeline {
   }
 
   tools {
-     // the Java version we use to run the build (minimal requirement for Maven/Tycho)
-     // we force other Java versions through Maven toolchains
-     jdk "temurin-jdk17-latest"
+     // the Java version we use to run the build
+     // we force the effective JDK version for compilation/testing through Maven toolchains
+     jdk "temurin-jdk21-latest"
   }
 
   stages {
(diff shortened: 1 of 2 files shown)


LorenzoBettini:
@cdietrich let's see whether this fixes the problem for good.

We'll have to manually run a few parameterized Jenkins build.

cdietrich:
One the build has started once you should be able to schedule them

LorenzoBettini:
unfortunately JIRO is overloaded, thus also the started build is blocked... I'll check later today.

LorenzoBettini:
On Jenkins, it looks like it's working with 4.23, JDK11. I manually started 4.23, JDK17, and  latest, JDK21.

Of course, 4.23, JDK21 is expected not to work due to the test failures due to missing draw2d correct dependency providing the Java record we use in Java 21 tests.

LorenzoBettini:
The failure on macOS is of course flaky, nothing to do with this change.

LorenzoBettini:
It fails when using Java 21 because we're not using the toolchain in that case, so it still uses Java 17, and the profile requires 21 for the source and target of the compiler.

We could always use Java 21 in `tools` in Jenkinsfile and have a profile/toolchain for Java 17. Does it sound good, @cdietrich?

cdietrich:
i think we should then use dedicated toolchain in all cases.
i assume this also will require local config correct?
then we at least should document prominently

LorenzoBettini:
@cdietrich I've update the PR.

For Java 21 I'm not using the toolchain: I'd keep the case for the latest version of Java relying on the Java version used for the build. 

Now, I'll have to check the 3 combinations as before.

LorenzoBettini:
2 out of 3 are green; let's see latest+Java21 and if it's green I'll merge

PR #3143, opened 2024-08-10
Fix exclusion for LanguageServer uber jar assembly

HannesWell:
In the maven-dependency-plugin's 'excludeArtifactIds' the artifactId has to
be specified, not the bundle's Symbolic name.
Also artifacts not present anway.

Also exclude all *.java files in the shade plugin configuration to ensure no java source-files are included in any case (the maven-dependencies-plugin also copies some eclipse source-jars).

PR #3144, opened 2024-08-11
[Build] Set up publication of Test results as comment on pull-requests

HannesWell:
As suggested in https://github.com/eclipse/xtext/pull/3139#issuecomment-2280471647.

@cdietrich, @LorenzoBettini unfortunately this is only picked up after it is submitted to the main branch (since I don't have write access to this repo and the new workflow runs when another one is complted).
But as far as I can tell this is the same config as we use in
- m2e:  https://github.com/eclipse-m2e/m2e-core/tree/master/.github/workflows
- eclipse-platform/equinox/jdt/pde (a bit more convoluted): https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/.github/workflows

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -103,3 +103,12 @@ jobs:
         with:
           name: logs-${{ runner.os }}
           path: '**/target/**/*.log'
+  event_file:
+    name: "Upload Event File"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload
+      uses: actions/upload-artifact@834a144ee995460fba8ed112a2fc961b36a5ec5a # v4.3.6
+      with:
+        name: Event File
+        path: ${{ github.event_path }}
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,41 @@
+name: Publish Unit Test Results
+
+on:
+  workflow_run:
+    workflows: ["Build"]
+    types:
+      - completed
+jobs:
+  unit-test-results:
+    name: Publish Test Results
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion != 'skipped'
+    permissions:
+      checks: write
+      pull-requests: write
+      contents: read
+      issues: read
+      actions: read
+
+    steps:
+      - name: Download and Extract Artifacts
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        run: |
+           mkdir -p artifacts && cd artifacts
+           artifacts_url=${{ github.event.workflow_run.artifacts_url }}
+           gh api "$artifacts_url" -q '.artifacts[] | [.name, .archive_download_url] | @tsv' | while read artifact
+           do
+             IFS=$'\t' read name url <<< "$artifact"
+             gh api $url > "$name.zip"
+             unzip -d "$name" "$name.zip"
+           done
+
+      - name: Publish Test Results
+        uses: EnricoMi/publish-unit-test-result-action@567cc7f8dcea3eba5da355f6ebc95663310d8a07 # v2.17.0
+        id: test-results
+        with:
+          commit: ${{ github.event.workflow_run.head_sha }}
+          event_file: artifacts/Event File/event.json
+          event_name: ${{ github.event.workflow_run.event }}
+          files: "artifacts/**/*.xml"


HannesWell:
@cdietrich I assume the simplest would be to submit this and fix problems if they occur?

.github/workflows/maven.yml
```diff
@@ -103,3 +103,12 @@ jobs:
         with:
           name: logs-${{ runner.os }}
           path: '**/target/**/*.log'
+  event_file:
+    name: "Upload Event File"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload
+      uses: actions/upload-artifact@834a144ee995460fba8ed112a2fc961b36a5ec5a # v4.3.6
```
cdietrich, line 111: is there a reason to use an exact tag?

HannesWell: In general that's suggested practice to harden supply-chain security.
Otherwise it would be possible to publish a new artifact under a new tag (which is desired if one just uses the major version).
I can change it to just use `actions/upload-artifact@v4`. Without dependabot being setup it is otherwise tedious to update.

HannesWell: Btw. since you asked in xtext-reference projects, one can configure dependabot to just update workflows:
https://github.com/eclipse-m2e/m2e-core/blob/master/.github/dependabot.yml

And it can also handle these specific commit-ids.

LorenzoBettini (approved):
Let's merge it and see how it goes!

Thanks @HannesWell

cdietrich:
@HannesWell this seems to have broken the build

Run actions/upload-artifact@v4
With the provided path, there will be 978 files uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run

any idea

https://github.com/eclipse/xtext/actions/runs/10334822997/job/28608707561?pr=3143

cdietrich:
this is from your pr 
https://github.com/eclipse/xtext/pull/3143
lets also check mine.

cdietrich:
is it possible the matrix and the build-maven-artifacts clash?
will try to provide a pr

HannesWell:
I have investigated the uploaded artifacts in detail and there indeed test-failures in the runs. One can find them when searching the build-logs for example for `<<< FAILURE`.
I have no clue at the moment why the details are not shown. I have to investigate that further.

LorenzoBettini:
IIRC, that's just an output log, but not a real test failure. Tests reports should not show real failures. Alternatively, we enabled rerun for flaky tests, so that might be a failure in one of the first run failures. If the plug-in is not able to report flakes then it might be useless.

HannesWell:
> IIRC, that's just an output log, but not a real test failure. [...] If the plug-in is not able to report flakes then it might be useless.

The test-report.xml shows failure in it's headers:
```
<testsuite xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:noNamespaceSchemaLocation="https://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd"
version="3.0"
name="org.eclipse.xtend.ide.tests.compiler.QuickDebugSourceInstallingCompilationParticipantTest"
time="1.324" tests="3" errors="2" skipped="0" failures="0">
```
That the actual builds do not fail is because of https://github.com/eclipse/xtext/pull/3148, which I suggested based on the assumption that the test-collect job will fail if there are test-failures. And then one can see immediately from the run results if run-failures are test-failures (only the `Publish test result` job failed) or if there are for example compiler or other errors (the build jobs fail). 
But if one cannot access the details of the test-failures easily it has indeed only a limited advantage. But from all of the Eclipse-SDK projects I know that this workflow works in general. 

But surprisingly in https://github.com/eclipse/xtext/pull/3148#issuecomment-2282909418 the result is as expected and also the annotations are applied as expected when one clicks on the details.

Maybe it just didn't work before because not everything was available then? 🤔
But with the following two PRs we can now check both scenarios, a PR from a committer and one external
- https://github.com/eclipse/xtext/pull/3153
- https://github.com/eclipse/xtext/pull/3154

LorenzoBettini:
I think the build should fail if there are test failures so the Maven option you added is wrong to me. I'd like to see the environment (build) with test failures, not to have a failure only on the test collect build. For example, we know macos is flaky, so its failure could be ignored most of the time. In the current state, it's harder to detect that. At least, that's what I'd find useful. In the current state I don't see benefits from the test report. At least, that's my POV. Unfortunately, I currently don't have time (holiday) to look into that further. Maybe a test report on a per-build matrix basis would be best. If it's too much work, at least a test report for Linux Java 17 and Linux Java 21. 

To me, it's crucial now to remove that Maven option. Personally, I'd have a strong opinion on this point.

HannesWell:
> In the current state I don't see benefits from the test report. At least, that's my POV. Unfortunately, I currently don't have time (holiday) to look into that further. Maybe a test report on a per-build matrix basis would be best. If it's too much work, at least a test report for Linux Java 17 and Linux Java 21.

Actually I thought this would be a quick-win since it works good in the project's I'm involved and I actually also didn't want to spend much time on this. But I have no clear idea, why it's different here. Multiple builds should not be a problem (Eclipse-platform also has a matrix build). Maybe it's that there are really multiple jobs not one job executed in a matrix?
I don't think that pushing branches to the repo directly is a problem since it works in that case.

> I think the build should fail if there are test failures so the Maven option you added is wrong to me. I'd like to see the environment (build) with test failures, not to have a failure only on the test collect build.
[...]
> 
> To me, it's crucial now to remove that Maven option. Personally, I'd have a strong opinion on this point.

In general you can see the originating workflow if the failure annotations work. But of course it's one click more and you don't see it immediately.  
Created https://github.com/eclipse/xtext/pull/3155 to revert that.

LorenzoBettini:
@HannesWell now the test report job always fails when extracting the archives.. I wonder whether our archives are too big..

HannesWell:
The latest executions of the `Publish Unit Test Results` in https://github.com/eclipse/xtext/actions failed `Archive:  test-results-Linux-17.zip` because test-results-Linux-17.zip seems to be present twice and then the unzip command asks if conflicting content should be overwritten.
I have not yet checked why linux-17 results are present twice in the first place.

cdietrich:
Cause there is two Linux builds with two j versions

LorenzoBettini:
But I thought the archives would be different because of the java version used in the name

HannesWell:
> Cause there is two Linux builds with two j versions

Yes, but we have three 'Linux' results: one `test-results-Linux-21.zip` and two `test-results-Linux-17.zip`.

If you search e.g. https://github.com/eclipse/xtext/actions/runs/10422949927/job/28868593443 for `Archive:  test-results-` you can see this. But it's only the linux-17 zip that's duplicated 🤔

cdietrich:
IMG_0018

cdietrich:
Here  I see just 1

HannesWell:
> But why does it show twice here

I'm wondering as well why this happens?

> Here I see just 1

At this job, in the head-line it says 10 artifacts altough in the bottom only two are listed. For example in:
https://github.com/eclipse/xtext/actions/runs/10421428353

What's definitivly wrong that the `Upload event file` job is executed in
in https://github.com/eclipse/xtext/actions/runs/10421796888

I forgot the condition for this one.
But no clue if this causes problems.

LorenzoBettini:
The GitHub event must be published for this action to work properly, if I understand its documentation correctly

HannesWell:
> The GitHub event must be published for this action to work properly, if I understand its documentation correctly

Yes, but only once per event. If you push a branch directly to the repo it is currently uploaded twice:
- https://github.com/eclipse/xtext/pull/3159

cdietrich:
```
curl -H "Accept: application/vnd.github.v3+json" \
     -H "Authorization: token ghp_xxxxxxx" \
     https://api.github.com/repos/eclipse/xtext/actions/runs/10421428353/artifacts | jq -r '.artifacts[] | [.name, .archive_download_url] | @tsv'  | while read artifact 
           do
             IFS=$'\t' read name url <<< "$artifact"
             echo "$name"
           done
``` 

```
Event File
test-results-macOS-17
test-results-Linux-17
test-results-Linux-21
test-results-Windows-17
build-maven-artifacts-test-results-Linux
test-results-Linux-17
test-results-Linux-21
test-results-macOS-17
test-results-Windows-17
```

cdietrich:
```
{
      "id": 1820726158,
      "node_id": "MDg6QXJ0aWZhY3QxODIwNzI2MTU4",
      "name": "test-results-macOS-17",
      "size_in_bytes": 19054144,
      "url": "https://api.github.com/repos/eclipse/xtext/actions/artifacts/1820726158",
      "archive_download_url": "https://api.github.com/repos/eclipse/xtext/actions/artifacts/1820726158/zip",
      "expired": false,
      "created_at": "2024-08-16T16:10:00Z",
      "updated_at": "2024-08-16T16:10:00Z",
      "expires_at": "2024-11-14T15:04:00Z",
      "workflow_run": {
        "id": 10421428353,
        "repository_id": 1553758,
        "head_repository_id": 1553758,
        "head_branch": "cd-mwe-2190-m3",
        "head_sha": "655bfa7fdc070f5c5e6618257cde9f9cebdb2d88"
      }
    },
    {
      "id": 1820312683,
      "node_id": "MDg6QXJ0aWZhY3QxODIwMzEyNjgz",
      "name": "test-results-macOS-17",
      "size_in_bytes": 175945,
      "url": "https://api.github.com/repos/eclipse/xtext/actions/artifacts/1820312683",
      "archive_download_url": "https://api.github.com/repos/eclipse/xtext/actions/artifacts/1820312683/zip",
      "expired": false,
      "created_at": "2024-08-16T14:21:16Z",
      "updated_at": "2024-08-16T14:21:16Z",
      "expires_at": "2024-11-14T14:18:29Z",
      "workflow_run": {
        "id": 10421428353,
        "repository_id": 1553758,
        "head_repository_id": 1553758,
        "head_branch": "cd-mwe-2190-m3",
        "head_sha": "655bfa7fdc070f5c5e6618257cde9f9cebdb2d88"
      }
    },
    ```

HannesWell:
> ```
> Event File
> test-results-macOS-17
> test-results-Linux-17
> test-results-Linux-21
> test-results-Windows-17
> build-maven-artifacts-test-results-Linux
> test-results-Linux-17
> test-results-Linux-21
> test-results-macOS-17
> test-results-Windows-17
> ```

That explains the failure.

But I don't understand why this happens?
Do files with the same name have identical content? 
If yes we can maybe just override in case of conflicts or ignore.

Can you also check a run with a PR from a fork? Maybe this indeed caused from pushing to the repo directly.

HannesWell:
>  "id": 10421428353, in the same
>  

Is that from the single event file?

cdietrich:
workflow run is the same. see json above

cdietrich:
next attempt: https://github.com/eclipse/xtext/pull/3160

HannesWell:
> next attempt: #3160

At least the publication of the test-results seems to work now, but the indication is still not working. Either the indicated test-failures are false (the corresponding build jobs should fail then) or posting the failure annotations does not work.
One reason could be this issue:
- https://github.com/EnricoMi/publish-unit-test-result-action/issues/12

But I don't understand why this also happens for PRs from forks. AFAICT the setup is almost the same like e.g. in m2e and there it works as expected: https://github.com/eclipse-m2e/m2e-core

I have to admit that I'm slowly losing my patience with this.
Do you think it is worth it to investigate further, i.e. do you really want this, or should this be reverted?

cdietrich:
from windows 

find . -name "*.txt" | xargs grep "Failures: [1-9]"
./org.eclipse.xtend.ide.tests/target/surefire-reports/org.eclipse.xtend.ide.tests.builder.WorkspaceScenariosTest.txt:Tests run: 4, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 4.600 s <<< FAILURE! -- in org.eclipse.xtend.ide.tests.builder.WorkspaceScenariosTest
./org.eclipse.xtend.ide.tests/target/surefire-reports/org.eclipse.xtend.ide.tests.compiler.ResolvingCrossReferenceDuringIndexingTest.txt:Tests run: 9, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 10.97 s <<< FAILURE! -- in org.eclipse.xtend.ide.tests.compiler.ResolvingCrossReferenceDuringIndexingTest

cdietrich:
i see some "-Dmaven.test.failure.ignore=true" in the repo. e.g. in `maven-build.sh` and `.github/workflows/maven.yml`

LorenzoBettini:
> i see some `-Dmaven.test.failure.ignore=true` in the repo. e.g. in `maven-build.sh` and `.github/workflows/maven.yml`

@cdietrich it's only in the shell script that we don't use anymore.

@HannesWell I'd tend to revert this action that doesn't seem to work well, at least in our configuration.. in the future I can try to look for alternative solutions.

cdietrich:
2024-08-16T20:35:15.1006049Z 
2024-08-16T20:35:15.1006206Z Results:
2024-08-16T20:35:15.1006459Z 
2024-08-16T20:35:15.1006621Z Flakes: 
2024-08-16T20:35:15.1007477Z org.eclipse.xtend.ide.tests.builder.WorkspaceScenariosTest.testJarWithoutJava
2024-08-16T20:35:15.1016298Z   Run 1: WorkspaceScenariosTest.testJarWithoutJava:111 mypack.MyData cannot be resolved to a type., MyData cannot be resolved., Couldn't find the mandatory library 'org.eclipse.xtext.xbase.lib' 2.8.0 or higher on the project's classpath. expected:<2> but was:<3>
2024-08-16T20:35:15.1018339Z   Run 2: PASS
2024-08-16T20:35:15.1018541Z 
2024-08-16T20:35:15.1019105Z org.eclipse.xtend.ide.tests.compiler.ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef
2024-08-16T20:35:15.1021759Z   Run 1: ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef:261->testResolvingXFunctionTypeRef:447->assertNoErrorsInWorkspace:460->Assert.assertFalse:65->Assert.assertTrue:42->Assert.fail:89 Couldn't find the mandatory library 'org.eclipse.xtext.xbase.lib' 2.8.0 or higher on the project's classpath.
2024-08-16T20:35:15.1023362Z   Run 2: PASS
2024-08-16T20:35:15.1023514Z 
2024-08-16T20:35:15.1024095Z org.eclipse.xtend.ide.tests.compiler.ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef_5
2024-08-16T20:35:15.1026444Z   Run 1: ResolvingCrossReferenceDuringIndexingTest.testResolvingXFunctionTypeRef_5:307->testResolvingXFunctionTypeRef:447->assertNoErrorsInWorkspace:460->Assert.assertFalse:65->Assert.assertTrue:42->Assert.fail:89 myannotation.MyAnnotation cannot be resolved to a type.
2024-08-16T20:35:15.1027829Z   Run 2: PASS
2024-08-16T20:35:15.1027970Z 
2024-08-16T20:35:15.1027976Z 
2024-08-16T20:35:15.1028217Z Tests run: 3120, Failures: 0, Errors: 0, Skipped: 53, Flakes: 3
2024-08-16T20:35:15.1028722Z 
2024-08-16T20:35:16.2797360Z [INFO] All tests passed

maybe there is something wrong with the flaky annotation

cc @szarnekow

cdietrich:
https://github.com/EnricoMi/publish-unit-test-result-action/issues/539

LorenzoBettini:
> [EnricoMi/publish-unit-test-result-action#539](https://github.com/EnricoMi/publish-unit-test-result-action/issues/539)

The issue seems to imply that flaky tests are supported when reported by Maven. So everything should be fine when there are flakes.

cdietrich:
There is still an open pr for that

HannesWell:
> > [EnricoMi/publish-unit-test-result-action#539](https://github.com/EnricoMi/publish-unit-test-result-action/issues/539)
> 
> The issue seems to imply that flaky tests are supported when reported by Maven. So everything should be fine when there are flakes.

Now I understand what you mean with flaky tests. I have to admit I didn't use `surefire.rerunFailingTestsCount` yet respectively wasn't aware of it.
Has one of you already verified (e.g. by inspecting the test-result archives) that this is really the cause? From my understand the [publish-unit-test-result-action](https://github.com/EnricoMi/publish-unit-test-result-action) only operates on them, so the build output should be irrelevant. But I don't know on which files that action operates exactly.

LorenzoBettini:
The log correctly shows flakes reported my surefire: it also tells you which run finally succeeds, or in some cases they finally fail. So I confirm that Maven and Tycho do the right thing. The above issue has been reopened saying that such flakes are currently not supported by the action. It would be enough the PR doesn't report failures in such cases, but I didn't find a way to achieve that.

cdietrich:
Guess we could try to xslt / sed the xml tags for flaky to dev null

HannesWell:
Thank you @LorenzoBettini, with https://github.com/eclipse/xtext/pull/3164, at least the checks are now green again if there is no persistent failure.
When https://github.com/EnricoMi/publish-unit-test-result-action/issues/539 is resolved, #3164 can hopefully be reverted and we get the full benefit from it.

PR #3147, opened 2024-08-11
attempt to fix upload

cdietrich:
cc @HannesWell

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -95,13 +95,13 @@ jobs:
         uses: actions/upload-artifact@v4
         if: ${{ failure() || cancelled() }}
         with:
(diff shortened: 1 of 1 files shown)


cdietrich:
or do we also have to add a run id or something?

cdietrich:
addd no there is a j17 and j21 build on linux

HannesWell:
This could fix the problem, but I have to check (the references) in detail when back home.

HannesWell:
I checked the existing workflows and there is either only have one job or use different names like here. So I think this fix is reasonable.
I just wonder why this didn't happen before, maybe because there were no failures?

Besides this, I think we also need https://github.com/eclipse/xtext/pull/3148 to make the test-publication work.

PR #3154, opened 2024-08-14
Specify m2e lifecycle mapping metadata using processing-instructions

HannesWell:
Using processing-instructions requires much less code and is always defined directly at the targeted execution.
They are a hidden feature of m2e but can be quite convenient:
https://eclipse.dev/m2e/documentation/release-notes-17.html#new-syntax-for-specifying-lifecycle-mapping-metadata

Resent versions of all Tycho Plugins contain embedded m2e-lifecycle metadata to disable their execution in the IDE by default.

The maven-clean goal is not executed by m2e at all, so does not have to be ignored.


szarnekow:
How can I find the failing tests?

LorenzoBettini:
@szarnekow unfortunately this action cannot handle flakes and interpret them as failures. However, in such a case, there's no link to failed tests, because there are none. When there are real failures a link is available.

LorenzoBettini (approved):
I'm fine with the proposed change. WDYT @cdietrich @szarnekow ?

cdietrich:
Can someone pr against xtext/xtext-reference-projects?
Can we hold this off until after the release (wanted to build m3 this afternoon)

cdietrich:
Does this require specific m2e versions

HannesWell:
> Can we hold this off until after the release (wanted to build m3 this afternoon)

For me this is not urgent. I just stumbled up the usage in Xtext and though you could save some LOC here.


> Does this require specific m2e versions

Yes M2E 1.7, which was, if I'm not mistaken, released June 2016.

PR #3155, opened 2024-08-14
[Build] Fail individual builds if test-failures occur

HannesWell:
Even if the Publish-Test-Results job fails in that case too (but it's currently not completely working as expected).

See https://github.com/eclipse/xtext/pull/3144#issuecomment-2289992783.


LorenzoBettini:
The build fails now

HannesWell:
> The build fails now

Yes, but I'm puzzled why this happens. Rebased the changes, maybe this was just a hiccup.

LorenzoBettini:
Thanks, for the moment I'm merging this

PR #3159, opened 2024-08-16
[Build] Run upload event file job always only once

HannesWell:


cdietrich:
Let’s see what happens

cdietrich (approved):

Issue #3161, opened 2024-08-17
Pot. Guava update needed


cdietrich:
https://github.com/eclipse-orbit/orbit-simrel/issues/41

@szarnekow @LorenzoBettini do you have any capacity. I don’t really have

(Also affects mwe)

merks:
What needs to be done?

merks:
Looking at SimRel right now, we already have duplicates because Papyrus has yet to provide an update for this release cycle and it's apparently futile to tell folks to stop including 3rd party bundles in their SimRel features:

![image](https://github.com/user-attachments/assets/4706bf52-df33-4edb-ab67-39ebd4718d72)

In any case, the new 33.3.0.jre version of guava fits into your current version range.

What problem are you anticipating?

cdietrich:
Bump in poms targets new project wizard tests for wizard
Build another mwe milestone consume that in Xtext

create bump in Xtext/xtext-reference-projects to test

cdietrich:
The version in target files is not a range

cdietrich:
Also the Pom and Osgiversion should match

szarnekow:
I think I can look into it this weekend

cdietrich:
building mwe milestone m4 ....

cdietrich:
https://github.com/eclipse/xtext/pull/3163
might take some time to get green ....
when stuff is promoted to maven central

szarnekow:
Thank you, @cdietrich

PR #3162, opened 2024-08-17
Update to guava 30.3.0.jre

merks:
https://github.com/eclipse/xtext/issues/3161


cdietrich:
This misses the wizard. Maybe features and category yml. Hard to check on phone

merks:
Yes, there were a whole bunch of 33.2.1 without a JRE suffix.  I'll monitor the build results.

cdietrich:
Was referring to TargetPltfornProject(xtend|java) and cliWizardIntegratuonTest or something similar

cdietrich:
looks like we have already removed guava from there

merks:
There is no reference to 33.2.1 left.  I searched for all references to guava and see no remaining places with versions also referenced there so I think the bases are covered.

cdietrich:
thx. lets wait for the build to finish. then i will merge both and produce a mwe m4

cdietrich:
thx @merks

szarnekow:
Thank you, @merks

PR #3164, opened 2024-08-18
Update unit-tests.yml with fail_on

LorenzoBettini:
Let's see whether the job does not fail when there are flaky tests


EnricoMi:
This change only takes effect **after** merging to main branch!

PR #3179, opened 2024-08-26
get rid of junit4 container that does not seem to be needed

cdietrich:


cdietrich:
grafik

LorenzoBettini (approved):
If one can still run tests from Eclipse I'm fine with that :)

HannesWell:
> If one can still run tests from Eclipse I'm fine with that :)

Yes. The JUnit-4/5 container just adds the junit-4/5 jars and their dependencies on the classpath with the need to specify it as a (test) dependency of the project using the technology of choice (e.g. Maven or PDE/Tycho).
If it is defined as explicit dependency, the container is not necessary.

LorenzoBettini:
> > If one can still run tests from Eclipse I'm fine with that :)
> 
> Yes. The JUnit-4/5 container just adds the junit-4/5 jars and their dependencies on the classpath with the need to specify it as a (test) dependency of the project using the technology of choice (e.g. Maven or PDE/Tycho). If it is defined as explicit dependency, the container is not necessary.

At least in the past, in my projects and in Xtext projects itself, without the JUnit container it wasn't possible to correctly run all the tests. IIRC, when you have a mixture of JUnit 4 and 5 tests, without the container (JUnit5), JUnit 4 tests were not executed. In fact, in the past, @cdietrich had to split a project into 2 projects (we couldn't use the JUnit container because of a bug of Tycho, I guess resolved, and so we had to separate JUnit 4 from JUnit 5 tests into 2 separate projects).

PR #3180, opened 2024-08-27
another classgraph bump

cdietrich:
also using 0.0.0 as @merks keeps doing updates too often
(our 2.36 release is broken, but i wont have time for a hotfix) :(

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -60,7 +60,7 @@ ASM (9.7)
 
 * License: BSD-3-Clause
 
-classgraph (4.8.174)
+classgraph (4.8.x)
 
 * License: MIT
 * Project: https://github.com/classgraph/classgraph
--- a/org.eclipse.xtext.common.types/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.common.types/META-INF/MANIFEST.MF
@@ -10,7 +10,7 @@ Require-Bundle: org.eclipse.emf.ecore;bundle-version="2.26.0",
  org.eclipse.xtext;bundle-version="2.37.0",
  org.objectweb.asm;bundle-version="[9.7.0,9.8.0)",
  org.eclipse.xtend.lib;bundle-version="2.37.0"
-Import-Package: io.github.classgraph;version="4.8.174",
+Import-Package: io.github.classgraph;version="4.8.175",
  org.apache.commons.logging;version="1.0.4";resolution:=optional,
  org.apache.log4j;version="1.2.24"
 Bundle-RequiredExecutionEnvironment: JavaSE-11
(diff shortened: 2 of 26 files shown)


merks:
Sorry, I didn't consider that this might break something.  I could have tried to back out the changes but now Orbit is also released.  Sorry sorry.

LorenzoBettini:
> also using 0.0.0 as @merks keeps doing updates too often (our 2.36 release is broken, but i wont have time for a hotfix) :(

What do you mean by broken? It cannot be installed?

cdietrich:
no, the new xtext project wizard will produce a non working target platform

mehmet-karaman:
> no, the new xtext project wizard will produce a non working target platform

Maybe there is a previous version of the orbit which we could refer?

cdietrich:
For the release it does not matter
As the orbit url and the version is hard coded

merks:
For future reference, the Platform's RC2 build is at offset -1 which means their contribution is Friday the week before everyone else contributes.  They do the build for that on Wednesday.  The Orbit, EMF, and ECF milestones are due Tuesday/today.  Given this is RC2/final, Orbit, EMF, ECF must (should) provide final content, i.e., release content.   To avoid last minute stress, the Orbit release build was done yesterday.  I could have backed out the version if I were informed sooner.  The update was done by this commit one week ago:

https://github.com/eclipse-orbit/orbit-simrel/commit/036cf7c7c42fdaf0f326a28ac9ce07078199b292

In general updates happen the day after something becomes available and updates generally continue until the Platform's RC2 week.

cdietrich:
this basically destroys the relase flow we did the last 8 years.
we alwayst built releases at RC1

merks:
Unfortunately I'm not aware of  every flow of every project.  It looks like the release was produced 2024-08-25 so if I'd produced Orbit milestones more rapidly toward the end of the release cycle, the Orbit change from 2024-08-21 would have been consumed by your build.  Live and learn.  

One could subscribe to the repo where the changes happen:

https://github.com/eclipse-orbit/orbit-simrel

One might do less hard coding, though I don't know if that's feasible.

One might time releases around Orbit releases if your release content depends on exact versions:

image

image

This pattern has been in place for quite some time.  I suppose the important change is that Obit updates are semi-automatic such that there typically changes right up until RC2, unless someone makes me aware of a change that is problematic so late in the cycle.

cdietrich:
i guess we just need to live with stuff exploding.
or have other people than me building....

LorenzoBettini:
@cdietrich we're still on time to release a fix, aren't we @merks ?
By the way, Maven artefacts are not affected, only p2 stuff: we could even just release another p2 site (the build qualifier will be increased) and skip the Maven deployment (though our Jenkinsfile should be updated in that respect). Otherwise, we fully release 2.36.1.

In general, I had already proposed not to use fixed versions of such dependencies in the target platform ;)

cdietrich:
I don’t think we did a dot release for years 
And I can’t contribute anything

cdietrich:
yes this is cause my solution would be "live with it"

Issue #3182, opened 2024-08-29
Xtend generates invalid code for extending dispatch method


HannesWell:
For an Xtend class that has a dispatch method, which 'extends' a dispatched method of a super-class, the generated code is faulty.
For example for
```
import org.eclipse.xtext.xbase.jvmmodel.AbstractModelInferrer
import org.eclipse.xtext.xbase.jvmmodel.IJvmDeclaredTypeAcceptor

class MyInferrer extends AbstractModelInferrer {
	
	def dispatch void infer(Number element, IJvmDeclaredTypeAcceptor acceptor, boolean isPreIndexingPhase) {
	}
}
```
the following code is generated
```
import java.util.Arrays;
import org.eclipse.xtext.xbase.jvmmodel.AbstractModelInferrer;
import org.eclipse.xtext.xbase.jvmmodel.IJvmDeclaredTypeAcceptor;
import org.eclipse.xtext.xbase.lib.XbaseGenerated;

@SuppressWarnings("all")
public class MyInferrer extends AbstractModelInferrer {
  protected void _infer(final Number element, final IJvmDeclaredTypeAcceptor acceptor, final boolean isPreIndexingPhase) {
  }

  @XbaseGenerated
  public void infer(final Number element, final IJvmDeclaredTypeAcceptor acceptor, final boolean isPreIndexingPhase) {
    _infer(element, acceptor, isPreIndexingPhase);
    return; else {
      throw new IllegalArgumentException("Unhandled parameter types: " +
        Arrays.<Object>asList(element, acceptor, isPreIndexingPhase).toString());
    }
  }
}
```
The line `return; else {` causes the compiler-error `Syntax error on token "else", delete this token.`

cdietrich:
@LorenzoBettini was this introduced with your recent change or is this older

cdietrich:
Maybe the problem is override a method with dispatch that is not a dispatch method in the superclass and thus there are no cases

LorenzoBettini:
> @LorenzoBettini was this introduced with your recent change or is this older

I haven't changed the actual logic of generation.

cdietrich:
hmm

```
class Viech extends AbstractModelInferrer {
	def dispatch void infer(Number element, IJvmDeclaredTypeAcceptor acceptor, boolean isPreIndexingPhase) {
	}
}
```

gives me

```
public class Viech extends AbstractModelInferrer {
[...]

PR #3185, opened 2024-09-03
Only add @XbaseGenerated annotation if available

HannesWell:
Fixes https://github.com/eclipse/xtext/issues/3182


cdietrich:
thx @HannesWell

Issue #3188, opened 2024-09-08
Even more problems with latest jdt


cdietrich:
/home/jenkins/agent/workspace/xtext_main/org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/InMemoryJavaCompiler.java:[207] 
05:21:12  [ERROR] 	this.compilerOptions.inlineJsrBytecode = true;
05:21:12  [ERROR] 	                     ^^^^^^^^^^^^^^^^^
05:21:12  [ERROR] inlineJsrBytecode cannot be resolved or is not a field
05:21:12  [ERROR] 1 problem (1

cdietrich:
@iloveeclipse what is the proper replacement?

iloveeclipse:
Simply remove it, the option is enabled by default.

cdietrich:
of course now i see a ton of test errors i cannot reproduce locally.

cdietrich:
Caused by: java.lang.Error: Unresolved compilation problem: 
	SANITY CHECK: Invalid opcode -88 at pc 162 for stackmap table attribute for method foo()

	at foo.Test.foo(Test.java:7)
	... 59 more

cdietrich:
=> i assume we need to use another ton of reflection :(

cdietrich:
xtext java 11 uses jdt 3.33

iloveeclipse:
> xtext java 11 uses jdt 3.33

You mean 4.33? The option there is **always true** independently what is set. In 4.34 the option is gone.

cdietrich:
jdt core 3.33

iloveeclipse:
OMG

PR #3189, opened 2024-09-08
remove inlineJsrBytecode following upstream removal in jdt

cdietrich:


iloveeclipse:
Fixes https://github.com/eclipse/xtext/issues/3188

LorenzoBettini (approved):
IIRC that's required because we use, in POM, an ancient version of jdt where this wasn't set by default, and in the recent versions the option is not even available anymore, right?

I think we should keep track of these reflection hacks so that we'll remove them when we switch to Java 17 and update to more recent version of jdt in POM as well.

cdietrich:
@LorenzoBettini as my committer status is caught in eclipse fdn bureaucracy can you please merge

Issue #2115, opened 2022-11-12
Java 17 Masterplan Part II


cdietrich:
This is for the left-overs from #1982

- [ ] Guice 5.1.0 update? https://github.com/eclipse/xtext/issues/2056
- [ ] Gradle 7.5.x https://github.com/eclipse/xtext/issues/2079
- [ ] https://github.com/eclipse/xtext/issues/2009
- [ ] bump emf to 2022-03 version
- [ ] guava update? https://github.com/eclipse/xtext/issues/2101
- [ ] deal with all the deprecations
- [ ] org.eclipse.xtext.builder.standalone.compiler.CompilerConfiguration - change default version to 1.8 or even better 11
- [ ] Adapt & Cleanup Wizard and Tests
- [ ] adapt minimum deps for platform/jdt/mwe/junit in manifests
- [ ] jetty for xtext-web/wizard https://github.com/eclipse/xtext-web/issues/374
- [ ] https://github.com/eclipse/xtext-maven/issues/186
- [ ] https://github.com/eclipse/xtext-xtend/issues/1391
- [ ] cleanup switches in code

cdietrich:
see also https://github.com/eclipse/xtext/issues/2686

cdietrich:
Guice and Guava Update done too

LorenzoBettini:
@cdietrich IIRC, the minimal target will be 2024-03, and we will get rid of all the other older target platforms, right?

cdietrich:
Correct

PR #3198, opened 2024-09-17
better workaround for inlineJsrBytecode jdt removal

cdietrich:


org.eclipse.xtext.xbase.testing/src/org/eclipse/xtext/xbase/testing/InMemoryJavaCompiler.java
```diff
@@ -223,7 +224,8 @@ public InMemoryJavaCompiler(ClassLoader parent, JavaVersion javaVersion) {
 	private final static MethodHandle INLINE_JSR_BYTECODE = findInlineJsrBytecode();
 	private static MethodHandle findInlineJsrBytecode() {
 		try {
-			if (JavaCore.getPlugin().getBundle().getVersion().compareTo(new Version(3, 39, 100)) >= 0) {
+			Plugin plugin = JavaCore.getPlugin();
+			if (plugin != null && plugin.getBundle().getVersion().compareTo(new Version(3, 39, 100)) >= 0) {
 				return null;
 			} else {
```
szarnekow, line 230: I'd think this can all be removed. If the setter isn't present, it'll throw an exception and return null anyways

cdietrich: the question is: how can we find places for removal
do you have an idea?

szarnekow: Ah, I understand. You are thinking `getVersion` is a good indicator to look for obsolete code after bumping the minimal TP...

cdietrich: yes

LorenzoBettini: By the way, I added a checkbox in https://github.com/eclipse/xtext/issues/2115 for keeping track of these "hacks" when we increase the minimal version of JDT core.

szarnekow (approved):
LGTM

LorenzoBettini:
There are test failures in Linux..

Issue #3129, opened 2024-08-08
JDT Facade project


LorenzoBettini:
In some parts, we use internal members of JDT or use reflection to perform different method calls according to the version of JDT (see https://github.com/eclipse/xtext/pull/3128 and https://github.com/eclipse/xtext/pull/3124)

We could confine such code in a separate project, acting as a facade for JDT. At least the code using reflection and internal JDT members is not spread over all our projects.

LorenzoBettini:
The project was created and initial hacks were moved there.

I'll leave this open because there are other parts to be moved in the jdt.facade project

PR #3204, opened 2024-09-24
Introduced JDT Facade project

LorenzoBettini:
For #3129 

This is a possible JDT facade project I had in mind.

For the moment, it only includes https://github.com/eclipse/xtext/pull/3189 but I was planning to confine there all, or most, of our usage of JDT internals.

However, I'd first start with reflection hacks.

I'd like to have some initial feedback @cdietrich @szarnekow


LorenzoBettini:
@cdietrich @szarnekow, could we merge this one so that the JDT Facade project is introduced, and then we or I can add other JDT internal calls and hacks to the project?

PR #3205, opened 2024-09-25
replace obsolete reflection usage with original method calls

mehmet-karaman:


cdietrich:
@mehmet-karaman can you please rebase this on the lastest main

mehmet-karaman:
Something looks really odd.. I am going to check that..

cdietrich:
no idea what is happening here neither
do you @LorenzoBettini 
why did the rebase not help?

cdietrich:
@mehmet-karaman can you try to ammend the commit with a new date e.g. `--date=now`

mehmet-karaman:
No, I use egit "Rebase interactive" for rebasing.

cdietrich:
git command would be `git commit --amend --no-edit --date=now`

cdietrich:
seems to have helped. jobs started again. lets see what happens

LorenzoBettini:
I don't understand the original problem: the build had restarted with no failure. Now the build fails with a compilation error: has jdt just removed a field from CompilerOptions?

mehmet-karaman:
https://github.com/eclipse-jdt/eclipse.jdt.core/commit/59bc0bef09d5dc7ced8f7dc1a4e6f91d3108e817 

This one seems to be the commit, which breaks it.

mehmet-karaman:
@iloveeclipse is there any suggestion what we can use instead of CompilerOptions.originalComplianceLevel ?

mehmet-karaman:
the compilation error should be fixed with the other PR:

https://github.com/eclipse/xtext/pull/3206

If the CompilerOptions are obsolete, then logically they are also obselete for xtext.. or isn't it?

cdietrich:
See the other issue from 2 weeks ago

mehmet-karaman:
Since the least java version  was set to 1.8 it should work with older versions.. The addSuppressed method is visible in java 1.8 so it shuld work.

The other changes were regarding the workarounds for eclipse oxygen and older, which isn't supported anymore.. So my viewpoint is that it should work with all supported jdk's and eclipse versions.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/Exceptions.java
```diff
@@ -28,28 +25,4 @@ public static <T> T throwUncheckedException(Throwable e) {
...
 	}
 
-	/**
-	 * Invoke {@code Throwable#addSuppressed(Throwable)} reflectively if it is available.
-	 * 
-	 * It is not available on JRE &lt; 1.7
-	 * 
-	 * @since 2.8
-	 */
-	public static void addSuppressed(Throwable owner, Throwable add) {
```
cdietrich, line 38: @szarnekow should we mark this forRemoval first?

mehmet-karaman: Maybe we can skip this, I can add this method back and mark this with "@Deprecated" .. or what ever xtext standard is.. Which annotation I should add?

cdietrich: `@Deprecated(forRemoval=true)` + javadoc with deprecated

mehmet-karaman:
@cdietrich do you have an idea why there are test failures? My branch is up to date, nothing to rebasese.. The build was successful before I've added this method back, so the failures makes no sensee for me.

cdietrich:
no. i have restarted the job

cdietrich (approved):

PR #3206, opened 2024-09-25
Remove obsolete compiler option assignments

mehmet-karaman:
This should fix the compilation errors caused by removed CompilerOptions in Jdt.

### DIFF
--- a/org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.java
+++ b/org.eclipse.xtext.java/src/org/eclipse/xtext/java/resource/JavaDerivedStateComputer.java
@@ -286,9 +286,21 @@ public class JavaDerivedStateComputer {
 		compilerOptions.sourceLevel = sourceLevel;
 		compilerOptions.produceMethodParameters = true;
 		compilerOptions.produceReferenceInfo = true;
-		compilerOptions.originalSourceLevel = targetLevel;
 		compilerOptions.complianceLevel = sourceLevel;
-		compilerOptions.originalComplianceLevel = targetLevel;
+		if (ORIGINAL_SOURCE_LEVEL != null) {
+			try {
+				ORIGINAL_SOURCE_LEVEL.invoke(compilerOptions, targetLevel);
+			} catch (Throwable e) {
+				// ignore
+			}
+		}
+		if (ORIGINAL_COMPLIANCE_LEVEL != null) {
+			try {
+				ORIGINAL_COMPLIANCE_LEVEL.invoke(compilerOptions, targetLevel);
+			} catch (Throwable e) {
+				// ignore
+			}
+		}
 		if (INLINE_JSR_BYTECODE != null) {
 			try {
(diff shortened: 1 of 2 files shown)


cdietrich:
Which is the minimal jdt version 
This removal requires 
Last time we needed to use reflection

cdietrich:
@mehmet-karaman can we check if there is more stuff that is marked for removal and is in @iloveeclipse delete queue

mehmet-karaman:
oo .. You're right.. I guess currently it compiles against the latest jdt... but it has to be conform until 2022-03, right?

cdietrich:
Yes. Depending if the team finds time to work on the j17 minimum story

mehmet-karaman:
Are there tests which are running on 2022-03? How we can recognize if we break something on older platform versions?

cdietrich:
yes the one on ci.eclipse.org/xtext (continuous-integration/jenkins/pr-head)

cdietrich:
but am not sure how good overage is.

LorenzoBettini:
What does that field mean? Is in general only setting the sourceLevel, without the originalSourceLevel enough (even in the past)?

cdietrich:
https://ci.eclipse.org/xtext/job/xtext/job/PR-3206/1/

Ton of test fails

LorenzoBettini:
> https://ci.eclipse.org/xtext/job/xtext/job/PR-3206/1/
> 
> Ton of test fails

OK, I see.

Another good candidate to be inserted here https://github.com/eclipse/xtext/pull/3204 ;)

mehmet-karaman:
I've contacted Andrey and srikant and they've said that it is obsolete since version 4.33 .. So for all versions before 4.33 we have to do the reflective workaround. I've also asked if there are any CompilerOptions planned for removal, he said that nothing is left. 

He also added that @cdietrich did similar workarounds for something with JSR .. The same has to be done in this case

mehmet-karaman:
I've found the PR which mentions something with JSR .. I am going to mimik this for the two compiler options.. 

https://github.com/eclipse/xtext/pull/3189

cdietrich:
@mehmet-karaman please note we had to rewrite that one 5 times.
please do the same as its currently on main.
see the 

```
if (INLINE_JSR_BYTECODE != null) {
			try {
				INLINE_JSR_BYTECODE.invoke(this.compilerOptions, true);
			} catch (Throwable e) {
				// ignore
			}
		}
```

mehmet-karaman:
Just a sidenote:
While checking the occurences, I've seen that JDT marked the following constant as deprecated: 
JavaCore.COMPILER_CODEGEN_INLINE_JSR_BYTECODE

Might be we also get problems with that in the future..

cdietrich (approved):

mehmet-karaman:
@cdietrich is this build ok now? Changed the state of this PR from draft to open..

cdietrich:
Will check when Jenkins is though and I am back from lunch nd meeting

mehmet-karaman:
:) finally all green

cdietrich:
thx @mehmet-karaman can you rebase the other one

PR #3210, opened 2024-10-02
EcoreUtil2: Introduce getAllContentsOfType() for resources

tfonda-fbk:
Hi, I'd like to propose this patch. Nothing special actually, simply since EcoreUtil2 already offers one variant of some of its methods for EObjects and one for Resources, I think this one could be handy as well. At the moment I'm working around its absence by using `EcoreUtil2.typeSelect(EcoreUtil2.eAllContentsAsList(resource), MyType.class)` extensively in my project, which works fine but looks quite verbose IMHO.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
@@ -189,6 +189,13 @@ public class EcoreUtil2 extends EcoreUtil {
 		return Lists.newArrayList(Iterators.filter(ele.eAllContents(), type));
 	}
 
+	/**
+	 * @since 2.37
+	 */
+	public static <T extends EObject> List<T> getAllContentsOfType(Resource resource, Class<T> type) {
+		return Lists.newArrayList(Iterators.filter(resource.getAllContents(), type));
+	}
+
 	public static <T> List<T> typeSelect(List<?> elements, Class<T> clazz) {
 		return Lists.newArrayList(Iterables.filter(elements, clazz));
 	}


org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
```diff
@@ -189,6 +189,10 @@ public static <T extends EObject> List<T> getAllContentsOfType(EObject ele, Clas
 		return Lists.newArrayList(Iterators.filter(ele.eAllContents(), type));
 	}
 
+	public static <T extends EObject> List<T> getAllContentsOfType(Resource resource, Class<T> type) {
```
cdietrich, line 192: am not sure if we want this. @szarnekow @LorenzoBettini wdyt

@tfonda-fbk can you please add a `@since 2.37`

LorenzoBettini:
It should be relatively innocent to have this. There's no test added in this PR, but I see that the current EcoreUtil2Test doesn't test similar methods either.

I was wondering: Why not simply implement this method in terms of `typeSelect`?

I mean something like

```java
return typeSelect(resource.getAllContents(), type);
```

tfonda-fbk:
> There's no test added in this PR, but I see that the current EcoreUtil2Test doesn't test similar methods either.

Yes, this is exactly why I have not added a test for this method.

> I was wondering: Why not simply implement this method in terms of `typeSelect`?

This was my initial idea, but then I went this way for the sake of uniformity wrt the EObject version of this method.

org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
szarnekow: Not a fan but I've also no hard feelings about the new method.

szarnekow (approved):

LorenzoBettini:
Thanks @tfonda-fbk for the contribution!

Could you please mention this new method in https://github.com/eclipse/xtext-website/blob/main/xtext-website/_posts/releasenotes/2024-11-25-version-2-37-0.md ?

PR #3211, opened 2024-10-02
fix for flaky editor sync test #2385

mehmet-karaman:
added a sleep before assert.. Tested on redhat 9 .. works as before.. tested on mac sonoma 14.6.1 .. works as before.. Needs observation

### DIFF
--- a/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
+++ b/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Objects;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.jobs.Job;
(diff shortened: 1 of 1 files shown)


mehmet-karaman:
@LorenzoBettini I've added a sleep of 250ms .. There was an unused and deprecated method, but for my case its not a build process, so i really needed sleeping thread. I am not sure if this helps..

LorenzoBettini:
IMHO, adding sleeping in tests is very bad design ;)

I don't understand the deprecated method you mention.

Maybe the problem is that you change the file with Java API and you don't refresh the corresponding eclipse resource?

mehmet-karaman:
Thats the sense of that syncronize :) It must recognize that the file was changed and change the content of the editor.

mehmet-karaman:
Is it ok, to add polling and sleep (5-10ms) to minimize the sleeping? For example wait in total 200ms but poll all 5ms and check if the assert is true.. if after 200ms the value isn't true, fail the test? pollingAssertThat()?

LorenzoBettini:
sleeping is not real synchronization ;)

If I understand the test and the implemented functionality correctly, it should be enough to trigger a refresh on the corresponding Eclipse resource, before doing the assert.

LorenzoBettini:
@mehmet-karaman I'm closing the PR and reopen it so that it builds against the new main branch where there's a fix for tests of web projects.

mehmet-karaman:
Ah maybe you meant "syncUtil.waitForReconciler(editor);" .. I've also removed that.. same reasong as waitForBuild..

LorenzoBettini:
To me the PR looks good. @szarnekow wdyt?

mehmet-karaman:
Its still flaky.. I guess on macosx its a little bit slower with the refresh, so I am going to add another wait for refresh job.. I hope that fixes the test problem for macosx.

LorenzoBettini:
@mehmet-karaman flaky on your machine or here on the CI? If it's on the CI, are you sure the flaky test is the one mentioned here?

mehmet-karaman:
The last macos build failed on the same test: https://github.com/eclipse/xtext/actions/runs/11215846276

LorenzoBettini:
Isn't refresh synchronous? I mean, when it returns the resource should be already refreshed, shouldn't it?

mehmet-karaman:
I am out of home for this week and i don't have access to my private mac machine. On Linux it works locally and on remote build fine. But on mac remote build it had a problem, which seems to work with the latest change.

The next possible time to test this for me on mac is on Saturday evening.

mehmet-karaman:
Could test it on mac. The refresh is synchronous (states also in the comments that it is long running). Removed the unnecessary refresh job wait after the refreshLocal-method call..

LorenzoBettini:
Thanks @mehmet-karaman

PR #3214, opened 2024-10-03
Bump Guava and remove junit strict version

### LINKED ISSUE #3213: Guava and junit.org need updates
Orbit added new versions for guava (3.33.1) and org.junit (4.13.2.v20240929-1000) so our builds fail since the TP cannot be resolved anymore.

I'd like to propose the same strategy we adopted for classgraph: remove the strict version in our target files and rely on dependabot for updating our BOM.

@cdietrich @szarnekow wdyt?

LorenzoBettini:
Closes #3213

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -70,7 +70,7 @@ Google Gson (2.9.x)
 
 * License: Apache License, 2.0
 
-Google Guava (33.3.0)
+Google Guava (33.3.1)
 
 * License: Apache License 2.0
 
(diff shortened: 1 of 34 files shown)


cdietrich:
Can you also open a Xtext reference projects pr

LorenzoBettini:
> Can you also open a Xtext reference projects pr

I can't promise I can do that today, even because I always forget what to do in that repository...

LorenzoBettini:
I'm merging this one and created https://github.com/xtext/xtext-reference-projects/pull/451

PR #3218, opened 2024-10-08
Remove strict version of Guava in some places

LorenzoBettini:
- from the target platform
- in NOTICE replaced patch version with `x`
- removed redundant version from an Xtext Maven plugin IT

NOTE: in other places, MANIFEST, feature and category, possible version bumps must still be done manually

### DIFF
--- a/org.eclipse.xtend.maven.plugin/src/test/resources/it/pom.xml
+++ b/org.eclipse.xtend.maven.plugin/src/test/resources/it/pom.xml
@@ -14,16 +14,15 @@
 	<dependencyManagement>
 		<dependencies>
 			<dependency>
-				<groupId>com.google.guava</groupId>
-				<artifactId>guava</artifactId>
-				<version>33.3.1-jre</version>
-			</dependency>
-			<dependency>
 				<groupId>org.eclipse.xtext</groupId>
 				<artifactId>xtext-dev-bom</artifactId>
 				<version>${xtextBOMVersion}</version>
 				<type>pom</type>
 				<scope>import</scope>
+			</dependency>
+			<dependency>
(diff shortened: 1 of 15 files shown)


LorenzoBettini:
@cdietrich I manually stopped the "push" builds and restarted the Jenkins build because it was failing due to connections to Maven repository.

Otherwise, the build is green

cdietrich:
grafik

org.eclipse.xtend.maven.plugin/src/test/resources/it/pom.xml
```diff
@@ -13,18 +13,17 @@
...
-				<version>33.3.1-jre</version>
-			</dependency>
 			<dependency>
 				<groupId>org.eclipse.xtext</groupId>
 				<artifactId>xtext-dev-bom</artifactId>
 				<version>${xtextBOMVersion}</version>
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
```
cdietrich, line 23: do we know this works?

cdietrich (approved):

LorenzoBettini:
> grafik

Yes, I manually restarted the build in Jenkins

org.eclipse.xtend.maven.plugin/src/test/resources/it/pom.xml
LorenzoBettini: @cdietrich I've just noted that dependency is in the dependency management section, so its version would NOT be taken from the dev BOM.

However, that POM is used as the parent POM of the Xtend Maven Plugin IT projects, and none of such projects uses guava as a dependency. So, that is useless and I'm going to remove it and update the PR.

PR #3223, opened 2024-10-14
Introduce a bundle with dependencies for running MWE2 workflow

### LINKED ISSUE #3208: Introduce a bundle with dependencies for running MWE2 workflow
Currently, our wizard generates a MANIFEST for the DSL project with `additional.bundles` like this:

```
additional.bundles = org.eclipse.xtext.xbase,\
                     org.eclipse.xtext.common.types,\
                     org.eclipse.xtext.xtext.generator,\
                     org.eclipse.emf.codegen.ecore,\
                     org.eclipse.emf.mwe.utils,\
                     org.eclipse.emf.mwe2.launch,\
                     org.eclipse.emf.mwe2.lib,\
                     org.objectweb.asm,\
                     org.apache.commons.commons-logging,\
                     org.apache.log4j
```

which are needed only for running the MWE2 workflow; in fact, we use `additional.bundles` instead of required bundles: they are not needed at run-time.

Keeping this list up-to-date becomes a burden for the users if we (Xtext) need to update them, as it happened with `commons-logging`: the user cannot benefit from any automatic update. The users must be notified and manually update the list.

Moreover, the plain Maven project our wizard generates add such dependencies as real dependencies of the DSL project, while it would be enough to have them in the `exec-maven-plugin` configuration.

Why don't we introduce a bundle, say `org.eclipse.xtext.generator.dependencies` that requires all of the above bundles and simply generate something like

```
additional.bundles = org.eclipse.xtext.generator.dependencies
```

This would allow us to keep the dependencies properly updated and the users won't have to do anything special when updating to a new version of Xtext. Similarly, we could simply add that dependency in the `exec-maven-plugin` configuration of a plain Maven project. (NOTE: `xtext-antlr-generator` should still be added in the POM because we cannot rely that to be present in an Eclipse TP).

By the way, having such bundle as part of Xtext SDK would have also saved us from explicitly add `commons-lang` in the SDK feature as done in https://github.com/eclipse/xtext/pull/3203. What's in there would be forced to be installed and issues like https://github.com/eclipse/xtext/issues/3192 wouldn't have happened.

Unless I miss something, maintaining such a bundle with proper dependencies cannot be harder than maintaining such a list in the wizard. For the users, the new bundle would be very helpful.

wdt @cdietrich @szarnekow

LorenzoBettini:
Closes #3208 

A few notes:

* the new bundle replaces the dependencies we used to generate as additional.bundles in build.properties; it has to re-export such dependencies so that the mwe2 workflow finds what it needs to complete the execution
* the examples have been updated accordingly
* other build.properties, e.g., some of our test projects containing languages, have NOT been updated because the additional.bundles were less, so I found it useless
* I haven't updated the POM files generated from the wizard, for plain Maven projects and for Maven/Tycho projects, because I seem to understand that is not necessary:
    - mwe2 dependencies must be available to make the mwe2 file usable in the IDE, so in a plain Maven project they must be dependencies of the project, not of the exec-maven-plugin configuration, as they are, and in an Eclipse plug-in project, additional.bundles are enough
    - Maven/Tycho uses the additional.bundles in the current exec-maven-plugin configuration that we generate with the wizard, the only additional dependency of the exec-maven-plugin is the custom antlr generator that must be a Maven dependency, since we cannot deploy it in an Eclipse update site due to licence problems
    - summarizing, the new bundle would not make sense as a Maven dependency of the exec-maven-plugin

Maybe, for the above reasons, a better name could be `org.eclipse.xtext.xtext.generator.additional.bundles` for this new bundle, in case the current name does not sound correct. It would not even be required to deploy it as a Maven artifact, but it doesn't cost us anything to do that.

### DIFF
--- a/org.eclipse.xtext.xtext.generator.dependencies/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.xtext.generator.dependencies/META-INF/MANIFEST.MF
@@ -14,7 +14,7 @@ Require-Bundle: de.itemis.xtext.antlr;bundle-version="2.0.0";resolution:=optiona
  org.eclipse.emf.mwe.utils;bundle-version="1.13.0";visibility:=reexport,
  org.eclipse.emf.mwe2.launch;bundle-version="2.19.0";visibility:=reexport,
  org.eclipse.emf.mwe2.lib;bundle-version="2.19.0";visibility:=reexport,
- org.objectweb.asm;bundle-version="[9.7.0,9.8.0)";visibility:=reexport,
+ org.objectweb.asm;bundle-version="[9.7.1,9.8.0)";visibility:=reexport,
  org.apache.commons.commons-logging;bundle-version="1.3.4";visibility:=reexport,
  org.apache.log4j;bundle-version="1.2.24";visibility:=reexport
 Automatic-Module-Name: org.eclipse.xtext.xtext.generator.dependencies


org.eclipse.xtext.full.releng/pom.xml
```diff
@@ -33,6 +33,7 @@
 		<module>../org.eclipse.xtext.ide</module>
 		<module>../org.eclipse.xtext.testing</module>
 		<module>../org.eclipse.xtext.xtext.ide</module>
+		<module>../org.eclipse.xtext.generator.dependencies</module>
```
szarnekow, line 36: Should this be `xtext.xtext.generator.dependencies` ?

szarnekow: `org.eclipse.xtext.xtext.generator.additional.bundles` would be more correct but even longer ...

szarnekow: Yes, please

szarnekow: `org.eclipse.xtext.xtext.generator.dependencies` that is

LorenzoBettini: @szarnekow rename done.

LorenzoBettini:
I rebased. When the build finishes I'll merge

PR #3233, opened 2024-10-29
Java 17 as minimal version, 2024-03 as minimal platform, updated BOM and minimal dependencies

LorenzoBettini:
Moreover,

- the wizards use Java 17 as the minimal version
- in the BOM I used properties when feasible
- explanation for https://github.com/eclipse/xtext/commit/ce2c1717d4a809a8eb24d21302e2527e35f1f38d: now that Java 17 is enabled for byte code, `strictfp` tests had to be adapted (I was wrong about failure only in Java 21: `strictfp` has been removed from byte code since Java 17); Concerning `org.eclipse.xtext.common.types.access.impl.AbstractTypeProviderTest.testMethods_publicStrictFpMethod_01()`, I removed it, because in this test project we also use project with sources and 1.8 compilation level. I preferred to keep things as they are, and simply removed this test that wouldn't make sense anymore, anyway.


cdietrich:
i assume the following is planned for later

- javaversion enum
- wizard tests
- other adaptations, default values / references of JavaVersion.11 usages, isAtLeast calls etc
- reflection switches against older tps eg m2e

LorenzoBettini:
> i assume the following is planned for later
> 
> * javaversion enum

Do you mean remove old versions?

> * wizard tests

That's already implemented in this PR

> * other adaptations, default values / references of JavaVersion.11 usages, isAtLeast calls etc

If you mean remove old tests, then yes, it'll be in another PR

> * reflection switches against older tps eg m2e

Yes for switches. I don't understand m2e

cdietrich:
javaversion enum
here we deprecate stuff

LorenzoBettini:
> javaversion enum here we deprecate stuff
> 
> m2e: org.eclipse.xtext.m2e.XtextProjectConfigurator.call(T, String, String)

Sorry, I don't understand

cdietrich:
regarding wizard.
still see
grafik

cdietrich:
i guess we also need to check the java 8 removal to find out what else we did back then

cdietrich:
@HannesWell might be able to explain the m2e stuff or we need to search the old pr
there is different m2e in current eclipses and the old minimal target platform
so we do reflection crap to deal with it

grafik

LorenzoBettini:
> regarding wizard. still see grafik

The two projects are slightly different, concerning some configurations, and the first one uses Java 17 now anyway.

LorenzoBettini:
> am blind ?!? grafik

Then I got confused by another one. I'll remove the second one.

cdietrich:
regaerding emf. this reminds me of the last emf updated.
there we also had a cascade in fragments genmodels etc
regarding minimal emf version.
https://github.com/eclipse/xtext-core/pull/2033/files

LorenzoBettini:
> regaerding emf. this reminds me of the last emf updated. there we also had a cascade in fragments genmodels etc regarding minimal emf version. https://github.com/eclipse/xtext-core/pull/2033/files
> 
> am not sure if we should do everything in one go given the time constraints

I'd postpone that, and we'll also have to regenerate the languages

cdietrich:
@iloveeclipse @mehmet-karaman i wonder if you could test this branch in your env

iloveeclipse:
> i wonder if you could test this branch in your env

@mehmet-karaman : if you can change our Xtext build to use this branch instead of master, I could try to build platform with that on next Monday and run tests.

@cdietrich : I'm out of PC till monday.

HannesWell:
> @HannesWell might be able to explain the m2e stuff or we need to search the old pr
> there is different m2e in current eclipses and the old minimal target platform
> so we do reflection crap to deal with it

The changes are
- https://github.com/eclipse/xtext/commit/39d68197317339bf2f1be9abbc5d1a51e34edefb
- https://github.com/eclipse/xtext/commit/6b0c7ff925a3f32cc31ab7e0240ddf3b85b7b50c
- https://github.com/eclipse/xtext/commit/6c061e8d60c052d9fc31b5c425b35da2b3da4e5f
- https://github.com/eclipse/xtext/commit/f9181789cebdeda2c8dbf352efac5b4b1dc02dd2

So if you use m2e-2.0+ in your minimal target now you can just call `ProjectConfigurationRequest.mavenProjectFacade()` and `ProjectConfigurationRequest.mavenProject()` and remove the reflection code.

LorenzoBettini:
@cdietrich I've updated the project wizard tests https://github.com/eclipse/xtext/pull/3233/commits/5fdf15df02fe101ad5ee672deca7fd0a30df7b81

I removed the duplicate `mavenTychoP2J17` and renamed another one that uses JUnit 5 (that was the project I mentioned above, concerning a slightly different configuration).

cdietrich:
thx @LorenzoBettini 
there still seem to be files e.g. launches, manifest.mf_gen mentioning 
JavaSE-11

cdietrich:
is it ok when i push the cleanup of TargetPlatformProject?

cdietrich:
```
-                                       <AB>IF config.javaVersion.isAtLeast(JavaVersion.JAVA17)<BB>
-                                               <repository location="https://download.eclipse.org/releases/2024-12"/>
-                                       <AB>ELSE<BB>
-                                               <repository location="https://download.eclipse.org/releases/2023-03"/>
-                                       <AB>ENDIF<BB>
+                                       <repository location="https://download.eclipse.org/releases/2024-12"/>
```

LorenzoBettini:
> ```
> -                                       <AB>IF config.javaVersion.isAtLeast(JavaVersion.JAVA17)<BB>
> -                                               <repository location="https://download.eclipse.org/releases/2024-12"/>
> -                                       <AB>ELSE<BB>
> -                                               <repository location="https://download.eclipse.org/releases/2023-03"/>
> -                                       <AB>ENDIF<BB>
> +                                       <repository location="https://download.eclipse.org/releases/2024-12"/>
> ```

Nice catch! Yes, that needs to be updated as well.

cdietrich:
@LorenzoBettini i have a commit i can just push on your branch

LorenzoBettini:
> @LorenzoBettini i have a commit i can just push on your branch

Please go right ahead

cdietrich:
done

LorenzoBettini:
> My proposal is to bring this in. Update the reference projects. Check the mwe and xpect builds and once these are green build a m2 milestone
> 
> wdyt @szarnekow @LorenzoBettini

I agree!

cdietrich:
@szarnekow are you available this week?

iloveeclipse:
> i wonder if you could test this branch in your env

We've tested and saw no issues in our automated tests.

cdietrich (approved):

cdietrich:
i fear we are running out of time. and @szarnekow does not seem to be reachable ...

LorenzoBettini:
In case, I can build the milestone this week

cdietrich:
Next week is m3 week. And thius we should (tm) be feature complete by then

LorenzoBettini:
So we can build a milestone this week?

cdietrich:
I really had hoped for Sebastian to have a look. If we want to bring it in and considering open points and my availability 2nd half of this week it needs to be asap

cdietrich:
To check
- org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JREContainerProvider.java
- org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/model/ManifestAccess.xtend

org.eclipse.xtend.ide/META-INF/MANIFEST.MF
```diff
@@ -15,15 +15,15 @@ Require-Bundle: org.eclipse.xtend.core;bundle-version="2.37.0",
  org.eclipse.xtext.xbase.ui;bundle-version="2.37.0",
  org.eclipse.xtend.ide.common;bundle-version="2.37.0";visibility:=reexport,
  org.eclipse.ui;bundle-version="3.201.0",
- org.eclipse.jdt.core;bundle-version="3.29.0",
+ org.eclipse.jdt.core;bundle-version="3.37.0",
  org.eclipse.ui.ide;bundle-version="3.13.1",
  org.eclipse.ui.console;bundle-version="3.11.100",
- org.eclipse.core.filesystem;bundle-version="1.9.300",
+ org.eclipse.core.filesystem;bundle-version="1.10.300",
  org.eclipse.jdt.ui;bundle-version="3.26.0",
```
szarnekow, line 22: We should try to be more consistent with the require bundle dependencies, e.g. jdt.ui 3.26 is unlikely to match jdt.core 3.37

org.eclipse.xtext.common.types.eclipse.tests/META-INF/MANIFEST.MF
```diff
@@ -4,18 +4,18 @@ Bundle-Name: %pluginName
...
+ org.eclipse.core.resources;bundle-version="3.20.100",
+ org.eclipse.jdt.core;bundle-version="3.37.0",
+ org.eclipse.core.runtime;bundle-version="3.31.0",
+ org.eclipse.jdt.launching;bundle-version="3.21.100",
  org.eclipse.xtext.common.types.ui;bundle-version="2.37.0",
- org.eclipse.emf.common;bundle-version="2.24.0",
+ org.eclipse.emf.common;bundle-version="2.30.0",
  org.eclipse.xtext.ui;bundle-version="2.37.0",
  org.eclipse.xtext.ui.shared;bundle-version="2.37.0",
  org.eclipse.ui.editors;bundle-version="3.14.300",
```
szarnekow, line 21: Do you know whether there exists a utility that would update all require-bundle versions to match a selected target platform?

merks: Not that I know of, no.

org.eclipse.xtend.ide/META-INF/MANIFEST.MF
LorenzoBettini: @szarnekow I only updated the dependencies when there was a matching one in the dev BOM, IIRC.

In general, I'm OK with keeping in sync minimal versions for the same dependencies that we have in the BOM.
For the other ones, I'd like to avoid such a massive maintenance burden...

szarnekow:
Minor pending things
- org.eclipse.xtext.util.JavaRuntimeVersion.determineJavaVersion(String) should default to 17 and not 11
- org.eclipse.xtext.ui.util.JREContainerProvider.PREFERRED_BREE
- org.eclipse.xtext.xtext.generator.model.ManifestAccess.getContent()

cdietrich:
created https://github.com/eclipse/xtext/issues/3247 for the emf update

org.eclipse.xtext.common.types.eclipse.tests/META-INF/MANIFEST.MF
HannesWell: It would indeed be nice. E.g. API-Tools could suggest to use the current version from the TP as lower bound if the bundle has changed it's version.
Or if no internals are used at all API-Tools could even try to be smart and compute the lower bound from `@since` tags.

szarnekow (approved):

PR #3234, opened 2024-10-30
remove no longer needed reflections

cdietrich:


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -770,17 +760,10 @@ public ISchedulingRule getRule(int kind, Map<String, String> args) {
...
-				requestProjectRebuildMethod.invoke(this, true);
-			} catch (Exception e) {
-				log.error("something went wrong in triggerRequestProjectRebuild", e);
-				needRebuild();
-			}
-		} else {
+		try {
+			requestProjectRebuild(true);
+		} catch (Exception e) {
+			log.error("something went wrong in triggerRequestProjectRebuild", e);
```
cdietrich, line 766: am not sure if we want to keep this
@szarnekow wdyt

LorenzoBettini (approved):
Thanks for cleaning up these parts!

PR #3256, opened 2024-11-26
Bugfix for Hover which is visible above other application windows. 

mehmet-karaman:
eclipse.platform.ui https://github.com/eclipse-platform/eclipse.platform.ui/issues/2534

This patch is fixing the problem, where the HoverManagers listeners are deregistered but the hover is still visibile on top. This happens when the hover was sticky eclipse was minimized and put back on foreground after it. In that case the listeners are already deregistered and there is no chance to react at all. So for this case the visibility has to be set explicitly when the close is going to stop.

### DIFF
--- a/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/hover/XbaseInformationControl.java
+++ b/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/hover/XbaseInformationControl.java
@@ -514,8 +514,6 @@ public class XbaseInformationControl extends AbstractInformationControl implemen
 	@Override
 	public void setVisible(boolean visible) {
 		Shell shell = getShell();
-		if (shell.isVisible() == visible)
-			return;
 
 		if (!visible) {
 			super.setVisible(false);
@@ -523,6 +521,9 @@ public class XbaseInformationControl extends AbstractInformationControl implemen
 			startDisposeTimeout(shell.getDisplay());
 			return;
 		}
+		
+		if (shell.isVisible() == visible)
+			return;
 
 		/*
 		 * The Browser widget flickers when made visible while it is not completely loaded.


cdietrich:
can you please rebase

mehmet-karaman:
Can't stop the build.. It seems that my "rebase interactive" added some unwanted commits.. I will wait until build is finished and push the corrected branch in force mode.

mehmet-karaman:
@cdietrich this is a small bugfix, do you have time to check this?

cdietrich:
Sry I am totally overworked in my day job and regard my spare time : The xtext infra is broken (see open issues ) and we have to deal with other broken tests and more infra changes (might also affect mwe and xpect )

mybe Andrey can do a review and @LorenzoBettini also can have a look

mehmet-karaman:
There is a video where I show it: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2534 ... There are also the steps which reproduce this bug.. (The eclipse.platform.ui part of this bugfix is already merged)

LorenzoBettini:
I'll try to have a look today. The steps are described above, if I understand correctly.

mehmet-karaman:
Open an Xbase Editor, Hover over an identifier and wait for the hover.. Double click on it, minimize eclipse and bring eclipse back to foreground. The hover is now kind of "orphaned" and you can bring other application windows and the sticky dialog is still in foreground.

mehmet-karaman:
> I'll try to have a look today. The steps are described above, if I understand correctly.

Yes, similar, but with Xbase editor. (Video shows the same problem with javadoc hover)

LorenzoBettini:
@mehmet-karaman I tried with an Xtend editor and applied your patch: the problem unfortunately is still there. By setting a breakpoint in the method, the breakpoint is never hit, so maybe this doesn't fix it in Xtend?

LorenzoBettini:
I tried also with an Xbase editor (not Xtend): the problem is still there and the breakpoint is never hit...

LorenzoBettini:
@mehmet-karaman sorry I was using the hover for Javadoc, while you were mentioning hover for members. With members the breakpoint is hit, but I'm afraid I confirm that your patch does NOT solve the problem: after resuming the window the sticky popup is still there.

Unless I missed what you meant...

LorenzoBettini:
I'm on mac at the moment, that's where I made the experiment

LorenzoBettini:
Or should I also rely on the patched platform.ui?
I seem to understand that https://github.com/eclipse-platform/eclipse.platform.ui/issues/2534 is not fixed yet.

mehmet-karaman:
The PR was merged, so there is nothing to do in this issue anymore.. 

PR: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2540

mehmet-karaman:
> Or should I also rely on the patched platform.ui? I seem to understand that [eclipse-platform/eclipse.platform.ui#2534](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2534) is not fixed yet.

The PR for that issue was merged, I've also closed that ticket. 

@LorenzoBettini 
To test this fix you would need the latest changes of platform.ui:
The super implementation of XbaseInformationControl (the platform.ui part org.eclipse.jface.text.AbstractInformationControl) is also necessary for the fix..  My change for the Xbase part was just to be sure that the super.setVisible() will be called before exiting for the false case.

mehmet-karaman:
@LorenzoBettini  You could rely on the Platform.ui fix, as it is tested and reviewed by @iloveeclipse

mehmet-karaman:
@LorenzoBettini 
You could test it with the latest integration builds:
https://download.eclipse.org/eclipse/downloads/
 
If you would test this today you can use directly that one (I Builds are stored for 7 days on, after that the link will become invalid):
https://download.eclipse.org/eclipse/downloads/drops4/I20241201-1800/

LorenzoBettini:
@mehmet-karaman in my Xtext development workspace, where we have the patched `org.eclipse.jface.text
Bundle-Version: 3.26.100.v20241127-1134` I started a new Eclipse instance.

The problem is not there anymore **even without** your patch: if I have a focused hover popup, hide the window, and then restore it, the popup is still there, but clicking outside it closes the popup.

So what would your patch for Xtext do in addition to the one in jfact.text?

mehmet-karaman:
@LorenzoBettini 
Did you tested it with the Xbase editor or Xtext Editor? My patch fixes the problem of the Xbase editor.

LorenzoBettini:
> @LorenzoBettini Did you tested it with the Xbase editor or Xtext Editor? My patch fixes the problem of the Xbase editor.

With an Xbase editor: both the Domainmodel and Xtend.
I cannot reproduce the initial problem anymore.

mehmet-karaman:
Let me verify this again, if its fixed already I will close this PR.

mehmet-karaman:
I've recorded the two states.

Without Patch:

https://github.com/user-attachments/assets/fc940c1b-b7dd-4367-bf82-51ee5d5fd947

With Patch:


https://github.com/user-attachments/assets/61011943-9cc3-4fa9-a0d9-c76d1f26e457

LorenzoBettini:
@mehmet-karaman 

Additional tests: I tried without the patch in these environments (since I started to suspect that the problem is DE dependent):

1. GNOME Wayland
2. GNOME X11
3. KDE Wayland
4. KDE X11

I confirm that the problem (without your patch) can be reproduced ONLY in KDE X11, which I suspect is the environment you used in the video.
Applying the patch fixes the problem in KDE X11 and doesn't seem to break the other environments.

You should have been a bit more specific on the environment you used for your tests ;)

thanks for the contribution.

PR #3261, opened 2024-11-27
Cleanup tests using Java records

### LINKED ISSUE #3258: Cleanup tests using Java records
When Java 17 PR is merged, tests using Java records, based on assuming Java 21, should be updated NOT to check for Java 21 since records have been present since Java 17. Also remove dep on draw2d just for referring to a Java record: we can have a Java record as part of our testdata.

LorenzoBettini:
Closes #3258

Besides what's described in the issue, I also added Java records to ITs of our Maven plugins, since now we required Java 17. The ITs related to Java 21 are still useful because in those ITs we generate Java 21 bytecode.


LorenzoBettini:
@szarnekow I rebased on top of main.
Is it OK to merge?

LorenzoBettini:
@cdietrich @szarnekow can I merge?

cdietrich (approved):

PR #3262, opened 2024-11-27
Remove duplicate TemporaryFolders

### LINKED ISSUE #1072: Consider getting rid of "TemporaryFolder" classes
In several places we have copies of `org.junit.rules.TemporaryFolder`. 

The JavaDoc of our copies says

```
/**
 * Basically copied from {@link org.junit.rules.TemporaryFolder}
 * which looks different in different junit versions and does not reset
 * the private field {@link #folder} on the {@link #after()} event.
 * 
 * Another change is that the folder is only created on demand.
 * 
 * The TemporaryFolder Rule allows creation of files and folders that are
 * guaranteed to be deleted when the test method finishes (whether it passes or
 * fails):
 **/
```

We should evaluate if the mentioned reasons still justify having copies of that class. This involves determining what's the oldest JUnit version we need to be compatible with.

LorenzoBettini:
Closes #1072

- Added `TemporaryFolder` in `.xtext.testing`
- The existing one in `.xbase.testing` is now deprecated and extends the new one
- The other copies have been removed and code adapted to use the new one


szarnekow (approved):
Nice one

PR #3268, opened 2024-11-28
updated Oomph setup for the new organization

LorenzoBettini:


LorenzoBettini:
@merks should the Oomph catalog be updated according to the new Xtext organization? (`eclipse/xtext` -> `eclipse-xtext/xtext`)?

I also noted that once I modified the setup file, rerunning setup tasks was still using the old GitHub repository: I had to manually modified that in the dialog. I guess that's expected because it was in some cache, right?

cdietrich (approved):

merks:
Yes, the catalog should be updated.  Generally github.com redirects URLs but it looks like not for rawcontent...

The https://ci.eclipse.org/oomph/job/setup-archiver job has already picked up the changes:

image

So the server should be serving that up shortly.

PR #3270, opened 2024-11-28
adapt more links

cdietrich:


org.eclipse.xtend.maven.parent/pom.xml
```diff
@@ -18,9 +18,9 @@
 	<description>Xtend is a little language that compiles into idiomatic Java source code. You can use any existing Java library seamlessly from Xtend (and vice-versa). The compiled output is readable and pretty-printed, and tends to run as fast or faster than the equivalent handwritten Java code. It's the CoffeeScript for Java.</description>
 
 	<scm>
-		<connection>scm:git:https://github.com/eclipse/xtext-xtend.git</connection>
-		<developerConnection>scm:git:https://github.com/eclipse/xtext-xtend.git</developerConnection>
-		<url>https://github.com/eclipse/xtext-xtend.git</url>
+		<connection>scm:git:https://github.com/eclipse-xtext/xtext.git</connection>
```
LorenzoBettini, line 21: These 3 elements are not required: they are inherited from the parent pom.

LorenzoBettini (changes requested):

org.eclipse.xtend.maven.parent/pom.xml
LorenzoBettini: I mean the 3 XML tags you modified

cdietrich: Hmmmhave no longer pc Access can you adjust? We did we have 3

LorenzoBettini: I can do that tomorrow, but I don't understand your answer

LorenzoBettini: In the dev bom is ok: that doesn't inherit from the parent pom.

LorenzoBettini:
@cdietrich I made the changes, removing the XML sections and I also fixed the URL of the Xtend website

LorenzoBettini (approved):

PR #3275, opened 2024-12-02
Testfailure fix for MacOS (DirtyStateEditorSupportIntegrationTest)

mehmet-karaman:
This patch adds a delay (waitForReconciler-method call before checking the document content) and it should fix the synchronization problem which causes flaky tests under mac os.

### DIFF
--- a/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
+++ b/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/editor/DirtyStateEditorSupportIntegrationTest.java
@@ -134,7 +134,7 @@ public class DirtyStateEditorSupportIntegrationTest extends AbstractEditorTest {
 
 		Display.getDefault().readAndDispatch();
 		
-		assertNotEquals(document.get(), "");
+		assertNotEquals("", document.get());
 		assertTrue(editor.getDocumentProvider() instanceof FileDocumentProvider);
 		
 		FileDocumentProvider fileDocumentProvider = (FileDocumentProvider)editor.getDocumentProvider();
@@ -152,7 +152,10 @@ public class DirtyStateEditorSupportIntegrationTest extends AbstractEditorTest {
 		Job.getJobManager().join(ResourcesPlugin.FAMILY_AUTO_REFRESH, null);
 		syncUtil.yieldToQueuedDisplayJobs(new NullProgressMonitor());
 		
-		assertEquals(document.get(), "");
+		// This was added to avoid flakyness for testing on mac os
+		syncUtil.waitForReconciler(editor);
+		
+		assertEquals("", document.get());
 	}
 
 	/**


mehmet-karaman:
@LorenzoBettini I hope this helps for the flaky tests..

LorenzoBettini:
@mehmet-karaman looks like it fixes the problem, and for sure fixes the arguments passed to assert methods.
Let's merge it and see whether that test has been fixed for good.

LorenzoBettini:
https://github.com/eclipse-xtext/xtext/runs/33937371956

@mehmet-karaman it falls again..

mehmet-karaman:
Need to reproduce this myself and fix this instead of trial and error approach..

memin20:
image

Yesterday I've spend the whole day with experimenting on mac. It runs successful in a junit plugin test, but fails on maven build. I've added several debug outputs tried to understand the state of the object and when the change happens, but it seems that there is no way without trying to understand with debugging in tycho surefire.

LorenzoBettini:
If it fails in the Maven build constantly, maybe an explicit refresh?

memin20:
explicit Refresh of the IFile didn't helped.. It still fails.. There must be a Eclipse Job, which is scheduled but not executed..

LorenzoBettini:
Does it systematically fail?

memin20:
Mostly they fail, but sometimes they are succesful.. 

image

`Checking Eclipse Jobs
Flushing ui.
more work todo...
Animation start
Flushing ui.
Decoration Calculation
Flushing ui.
Reporting encoding changes.
Flushing ui.
Initializing Quick Diff
Flushing ui.
XtextReconcilerJob
Flushing ui.
Open Blocked Dialog
Flushing ui.
Building
Flushing ui.
Flushing content description cache.
Flushing ui.
Update Job
Flushing ui.
Compacting resource model
Flushing ui.
Workbench Auto-Save Job
Flushing ui.
Periodic workspace save.
Ending of Eclipse Jobs checks`

LorenzoBettini:
Just a few thoughts:

- you don't wait for each job, if I understand correctly 
- maybe that loop should be done right before the last assert

LorenzoBettini:
Maybe unrelated, but please when posting snippets use the markdown syntax instead of posting screenshots of code ;)

LorenzoBettini:
> * you don't wait for each job

By "wait" I mean "join".

memin20:
It causes a deadlock, because these threads are waiting for the ui thread.. So I have to do a kind of polling. I will remove that break check that again.

memin20:
I've executed the maven build several times and it seems the polling helped, the test is always successful and not flaky anymore. 

The asserts have to wait until all scheduled jobs are done (except the periodic jobs). I am going to clean up the code, execute that build several time and if its still ok.. i will create a new PR :) 

Thank your for your patience @LorenzoBettini

memin20:
The solution still fails sometimes, depending on the state of my mac.. I will try to run the junit plugin test with the same arguments as the maven build is using.. i must be able to debug this..

LorenzoBettini:
@memin20 please don't waste too much time on that. Given all your experiments it might also be a lost cause: there must be severe synchronization issues in the platform itself and we don't seem to be able to do anything about that..

LorenzoBettini:
I'd agree. @szarnekow wdyt?

PR #3307, opened 2024-12-30
set jre to 21

cdietrich:


LorenzoBettini:
@cdietrich maybe we could update Tycho as well to 4.0.11?

cdietrich:
Am not sure what we do in wizard about it. Maybe at diffrrrnt tp for different Java versions as we did in the past

cdietrich:
https://github.com/eclipse-xtext/xtext/pull/3309 (proposal)

PR #3309, opened 2024-12-30
java 17/21 switch in tp

cdietrich:


LorenzoBettini (approved):
I'm not a big fun of such solutions that were a burden to maintain in the past...
however, I can't come up with better ones.

If one uses 2025-03 without using m2e in the target platform there shouldn't be a problem without the Java version specification.

cdietrich:
Yes but I fear others will follow soon

PR #3322, opened 2025-01-28
Remove repeated calls to INode#getGrammarElement

rubenporras:
Since some of the implementations need to do some small computations to return the grammar element it is preferable to save the value in a local variable.


rubenporras:
for https://github.com/eclipse-xtext/xtext/issues/3323

szarnekow (approved):
LGTM

rubenporras:
thanks

PR #3329, opened 2025-01-30
Lazy creation of the Set resolving

rubenporras:
By doing that, when resources already linked are loaded from storage or when no cyclic resolution takes place, no
memory needs to be allocated for the set.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
@@ -120,11 +120,26 @@ public class LazyLinkingResource extends XtextResource {
 		if (isEagerLinking())
 			EcoreUtil.resolveAll(this);
 	}
-
+	
+	/**
+	 * The method can be overridden to avoid creation of a Set object for resources loaded from storage 
+	 * or when no cyclic resolution takes place.
+	 * 
+	 * @since 2.39
+	 * 
+	 */
+	protected LinkedHashSet<Triple<EObject, EReference, INode>> getResolvingSet(boolean initial) {
+		if (initial) {
+			return Sets.newLinkedHashSet();
+		} else {
+			return resolving;
+		}
+	}
+	
 	/**
 	 * @since 2.4
 	 */
-	protected LinkedHashSet<Triple<EObject, EReference, INode>> resolving = Sets.newLinkedHashSet();
+	protected LinkedHashSet<Triple<EObject, EReference, INode>> resolving = getResolvingSet(true);
 
 	/**
 	 * resolves any lazy cross references in this resource, adding Issues for unresolvable elements to this resource.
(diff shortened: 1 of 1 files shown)


rubenporras:
The error on MacOs on QuickDebugSourceInstallingCompilationParticipantTest looks unrelated to my change, is it maybe a flaky test?

I would like to rerun the job but I think I do not have the rights to do so. Could someone rerun it?

LorenzoBettini:
> The error on MacOs on QuickDebugSourceInstallingCompilationParticipantTest looks unrelated to my change, is it maybe a flaky test?
> 
> I would like to rerun the job but I think I do not have the rights to do so. Could someone rerun it?

Indeed that's unrelated and it's a flaky test. I created a separate PR for that https://github.com/eclipse-xtext/xtext/pull/3331 

when it gets merged we can restart the failing job, though I don't think that's a blocker for this PR.
Let's wait for @szarnekow

rubenporras:
I was not totally sure it was unrelated, so I thought I would rather ask. Thanks for confirming it.

LorenzoBettini:
I'm closing and reopening to trigger rebuild against the new main branch

rubenporras:
@szarnekow , could you take a look at this PR as well?

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
```diff
@@ -124,7 +124,7 @@ protected void doLinking() {
 	/**
 	 * @since 2.4
 	 */
-	protected LinkedHashSet<Triple<EObject, EReference, INode>> resolving = Sets.newLinkedHashSet();
+	protected LinkedHashSet<Triple<EObject, EReference, INode>> resolving;
```
szarnekow, line 127: This is a bit annoying since it never was nullable and now would become nullable.

szarnekow: Can you add a method instead?
```suggestion
	protected LinkedHashSet<Triple<EObject, EReference, INode>> resolving = createResolvingSet(true);
	
	protected LinkedHashSet<Triple<EObject, EReference, INode>> createResolvingSet(boolean initial) {
	  ..
	}
```

Your code could override that method and return null if initial is true.
Access to the resolving set can subsequently be null-guarded and `createResolvingSet(false)` can be use to make the instance. 

Thoughts?

rubenporras: That works for me, I have changed the PR. What do you think of it now?

szarnekow (approved):

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
szarnekow: Looks good.

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinkingResource.java
```diff
@@ -120,11 +120,24 @@ protected void doLinking() {
 		if (isEagerLinking())
 			EcoreUtil.resolveAll(this);
 	}
-
+	
+	/**
+	 * The method can be overridden to avoid creation of a Set object for resources loaded from storage 
+	 * or when no cyclic resolution takes place.
+	 * @since 2.38
```
szarnekow, line 127: Unfortunately it collided with the release. Should be 2.39 :/

rubenporras: I understand. I have changed it to 2.39.

PR #3331, opened 2025-01-31
Fixed a few other flaky tests

LorenzoBettini:
This should remove further flakyness

* `QuickDebugSourceInstallingCompilationParticipantTest`: in the end, it looks like the problem was that the generated `.class` Eclipse resource was not fully synchronized
* Some tests did not properly reset the autobuild state


szarnekow (approved):
Reasonable. Let’s try this

PR #3333, opened 2025-02-03
updated generation of AbstractLanguageServerTest

cdietrich:


cdietrich:
@LorenzoBettini @szarnekow is this file dirty for you too?

LorenzoBettini:
@cdietrich That happens when you use Java 21, I guess because of the different hierarchy structure.

Unfortunately, we should use Java 21 to run Eclipse and most of the UI tests. However, we target Java 17, so such dirty files should be reverted instead of committed. At least, that's what I always do.

I guess it's also safe to commit them; in the end, it's up to us.

cdietrich:
i see, so i will try on different pc with older tp....

LorenzoBettini:
> i see, so i will try on different pc with older tp....

no need to: it only gets regenerated in clean builds; if you revert it, it will stay the same

Issue #3339, opened 2025-02-07
NPE caused by null defaultImage in XtextEditorTickUpdater (updateEditorImage)


mehmet-karaman:
If the defaultImage is null and the updateEditorImage is triggered in a worker thread, it throws the following NPE.
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.graphics.Image.getBounds()" because "baseImage" is null
	at org.eclipse.jface.viewers.DecorationOverlayIcon.<init>(DecorationOverlayIcon.java:84)
	at org.eclipse.jface.viewers.DecorationOverlayIcon.<init>(DecorationOverlayIcon.java:98)
	at org.eclipse.xtext.ui.editor.XtextEditorErrorTickUpdater.updateEditorImage(XtextEditorErrorTickUpdater.java:100)
	at org.eclipse.xtext.ui.editor.XtextEditorErrorTickUpdater.modelChanged(XtextEditorErrorTickUpdater.java:195)
	at org.eclipse.jface.text.source.AnnotationModel.fireModelChanged(AnnotationModel.java:594)
	at org.eclipse.jface.text.source.AnnotationModel$InternalModelListener.modelChanged(AnnotationModel.java:256)
	at org.eclipse.ui.internal.texteditor.quickdiff.DocumentLineDiffer.fireModelChanged(DocumentLineDiffer.java:1543)
	at org.eclipse.ui.internal.texteditor.quickdiff.DocumentLineDiffer.fireModelChanged(DocumentLineDiffer.java:1526)
	at org.eclipse.ui.internal.texteditor.quickdiff.DocumentLineDiffer$2.run(DocumentLineDiffer.java:713)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

```

LorenzoBettini:
Fixed by https://github.com/eclipse-xtext/xtext/pull/3340

PR #3340, opened 2025-02-07
> #3339 NPE caused by null defaultImage in XtextEditorTickUpdater (updateEditorImage)

mehmet-karaman:
Set fields to volatile which could possible be accessed by multiple threads. Used local variables to store the value and added check with local variable.

### DIFF
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/editor/OverrideIndicatorModelListener.java
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/editor/OverrideIndicatorModelListener.java
@@ -62,7 +62,7 @@ public class OverrideIndicatorModelListener extends NullImpl implements IXtextMo
 	public static final String JOB_NAME = "Override Indicator Updater";
 	private static ISchedulingRule SCHEDULING_RULE = SchedulingRuleFactory.INSTANCE.newSequence();
 
-	private XtextEditor xtextEditor;
+	private volatile XtextEditor xtextEditor;
 	private Set<Annotation> overrideIndicatorAnnotations = Sets.newHashSet();
 
 	private Job currentJob;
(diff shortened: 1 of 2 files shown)


cdietrich:
can you also have a look at the subclasses

mehmet-karaman:
Found one more case, where this could happen.. OverrideIndicatorModelListener uses the editor, which is unset and its not volatile.. I've also used it as local variable to avoid setting to null afterwards..

cdietrich:
no i mean the xtend one. do we need the local stuff you do there too

mehmet-karaman:
Oh now I looked again, the changes in the "OverrideIndicatorModelListener" are not necessary, because the field is accessed only in the UI thread.. So no worker thread should execute this code..

The problem happens only if the UI thread calls the beforeDispose, meanwhile a worker thread accesses the fields which are set by the beforeDispose method to null. If you check the stacktrace in the ticket, you will see that the defaultImage is accessed by the worker thread, where the UI thread already set this defaultImage to null..

mehmet-karaman:
The changes in runInUIThread .. Are also obsolete, because as it says it is executed in the UI thread..

iloveeclipse (approved):

mehmet-karaman:
I've done a rebase.

iloveeclipse:
@mehmet-karaman : could you please check what is wrong with your commit metadata (autor mail seem not match one expected by ECA)?

mehmet-karaman:
eclipsefdn/eca problem seems to be fixed.

mehmet-karaman:
@cdietrich are we done here?

cdietrich:
i have no time to try

LorenzoBettini:
I'm not very familiar with this part of the code base.
If that's related to a bug, I can try to reproduce the bug as long as I'm given the detailed steps and verify that this PR fixes the bug.
Otherwise, @szarnekow may judge better than I do.
In general, I think the PR is OK.

iloveeclipse:
Originally this bug was reported during our automated tests and from the [stack trace](https://github.com/eclipse-xtext/xtext/issues/3339) it was almost obvious why that happened, since the test opens/closes Xtext editor and the background job simply runs after editor was already disposed.

I haven't tried, but to reproduce one should create breakpoint at the job code (`XtextEditorErrorTickUpdater.modelChanged()`) that runs while changing some text in Xtext based editor, change the text, close the editor and let the job continue. If that doesn't sow the problem, the text should be changed outside of the editor so we have a resource change. 

IMHO the code was buggy and the fix is more or less trivial.

mehmet-karaman:
I've recorded the behavior which was described by @iloveeclipse .. I hope it makes it clearer to understand the problem. 

(I could just save the video as 480p because 10mb was the maximum upload size for github, i hope its still readable) 

https://github.com/user-attachments/assets/fa85f966-e0a1-4935-9a18-926d4a858351

LorenzoBettini:
@mehmet-karaman thanks for the detailed explanation.
I think we can merge.

PR #3344, opened 2025-02-10
Fix the Issues query string and update version

tfonda-fbk:
Otherwise no issues are found.


LorenzoBettini:
@cdietrich do we want to keep this link always up-to-date? Or simply remove it?

cdietrich:
i usually update it every 3 months, but i was too busy with other things the last half year

PR #3348, opened 2025-02-12
Lazy creation of ensureCacheInitialized

rubenporras:
By doing that, when resources already linked are loaded from storage, no memory needs to be allocated for the cache.


rubenporras:
Also for https://github.com/eclipse-xtext/xtext/issues/3317

rubenporras:
I see the errors "Failed to resolve target definition file:/home/runner/work/xtext/xtext/org.eclipse.xtext.common.types/../xtext-latest.target: Could not find "com.google.gson/2.11.0" in the repositories of the current location". Is the target broken?

cdietrich:
maybe something new shows up in orbit

cdietrich:
https://github.com/eclipse-orbit/orbit-simrel/commit/352d5183df803a485f11c6a6f212d5c6eae79b2c

cdietrich:
WIP at https://github.com/eclipse-xtext/xtext/pull/3349

rubenporras:
@szarnekow , could you take a look at this PR.

As an example, XbaseLazyLinker is one class that never uses the cache, so it would profit from lazy initialization. Also all resources which are loaded from binary would profit as well.

Thanks

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinker.java
```diff
@@ -208,11 +201,17 @@ protected EObject createProxy(EObject obj, INode node, EReference eRef) {
 		return proxy;
 	}
 
+	private void ensureCacheInitialized() {
```
szarnekow, line 204: I think we can also mark the LazyLinker itself as an `@Singleton` which would reduce the number of objects, too.

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinker.java
```diff
@@ -208,11 +201,17 @@ protected EObject createProxy(EObject obj, INode node, EReference eRef) {
...
+		if (instantiableSubTypes == null) {
+			 instantiableSubTypes = new SimpleCache<EClass, EClass>(this::findInstantiableCompatible);
+		} 
+	}
 	/**
 	 * @since 2.7
 	 */
 	protected EClass getProxyType(EObject obj, EReference eRef) {
 		EClass referenceType = ecoreGenericsUtil.getReferenceType(eRef, obj.eClass());
+		ensureCacheInitialized();
```
szarnekow, line 214: Maybe check first whether `referenceType` is abstract or an interface and if not, use it directly?

rubenporras: Thanks, that is a good idea. I have adapted the commit accordingly.

org.eclipse.xtext/src/org/eclipse/xtext/linking/lazy/LazyLinker.java
rubenporras: This is a good idea. If you do not mind, I would like to do that however in another PR, after Simrel 2025-03 is released, i think making the class a `@Singleton`  could have side-effects if the class or subclasses are not thread-safe.

szarnekow (approved):

PR #3354, opened 2025-02-17
Flaky QuickDebugSourceInstallingCompilationParticipantTest

### LINKED ISSUE #3350: QuickDebugSourceInstallingCompilationParticipantTest is too often flaky on macOS
No matter how hard I try, this test often fails on GitHub Actions macOS, which is arm.

I tried many things, including refreshing after the build and also separating the Xtext Build from the Java build, https://github.com/LorenzoBettini/xtext/commit/4b3a8e3c5d2395cc9b371b64ec21acaac74a42bd

On macOS, it fails often.

I can only think of 2 solutions

- disable this test on macOS
- use our `@Flaky` annotation at least for this test

@szarnekow @cdietrich ?

LorenzoBettini:
Closes #3350

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
@@ -68,7 +68,7 @@ public class TraceForStorageProvider extends AbstractTraceForURIProvider<IFile, 
 		}
 
 		@Override
-		public long getModificationstamp() {
+		public long getModificationStamp() {
 			return file.getModificationStamp();
 		}
 
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
@@ -68,7 +68,7 @@ public class TraceForStorageProvider extends AbstractTraceForURIProvider<IFile, 
 		}
 
 		@Override
-		public long getTimestamp() {
+		public long getModificationstamp() {
 			return file.getModificationStamp();
 		}
 
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
@@ -69,7 +69,7 @@ public class TraceForStorageProvider extends AbstractTraceForURIProvider<IFile, 
 
 		@Override
 		public long getTimestamp() {
-			return file.getLocalTimeStamp();
+			return file.getModificationStamp();
 		}
 
 		@Override
--- a/org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/internal/AbstractTraceForURIProvider.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/internal/AbstractTraceForURIProvider.java
@@ -54,7 +54,7 @@ public abstract class AbstractTraceForURIProvider<SomeFile, Trace extends Abstra
 	
 	protected interface PersistedTrace {
 		AbsoluteURI getPath();
-		long getModificationstamp();
+		long getModificationStamp();
 		InputStream openStream() throws IOException;
 		boolean exists();
 	}
@@ -69,7 +69,7 @@ public abstract class AbstractTraceForURIProvider<SomeFile, Trace extends Abstra
 
 			public CacheKey(PersistedTrace source) {
 				this.path = source.getPath();
-				this.timestamp = source.getModificationstamp();
+				this.timestamp = source.getModificationStamp();
 			}
 
 			@Override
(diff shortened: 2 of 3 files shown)


cdietrich (approved):

LorenzoBettini:
@cdietrich it still fails.. I'm not planning to work on that any longer.. I'd go back to my original alternatives mentioned in the issue

cdietrich:
hmmm running on battery makes it fail for me again .... :(

cdietrich:
and the more logging i add the less i can reprodcue

cdietrich:
could track it down to org.eclipse.xtext.generator.trace.LineMappingProvider.getLineMapping(AbstractTraceRegion)
here null will be returned.

cdietrich:
of course now i can no longer reproduce at all ....

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
```diff
@@ -69,7 +69,7 @@ public AbsoluteURI getPath() {
 
 		@Override
 		public long getTimestamp() {
-			return file.getLocalTimeStamp();
+			return file.getModificationStamp();
```
szarnekow, line 72: Hmm I'm not sure.

LorenzoBettini: @szarnekow I haven't tried, but the Javadoc of `getLocalTimeStamp`, https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fcore%2Fresources%2FIResource.html&anchor=getLocalTimeStamp(), also says

> Note that due to varying file system timing granularities, this value is not guaranteed to change every time the file is modified. For a more reliable indication of whether the file has changed, use [getModificationStamp()](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/resources/IResource.html#getModificationStamp()).

It sounds like the method is quite unreliable...

cdietrich:
i guess a sleep of 1s would also help. but am not sure if other tests might be affected too

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
cdietrich: i guess a sleep of 1s would also help. but am not sure if other tests might be affected too

LorenzoBettini: > Hmm I'm not sure.

@szarnekow what do you think is wrong with `getModificationStamp`?

szarnekow: It's not a timestamp and thus *should* not be used to implement `getTimestamp`

It might be fine to rename getTimestamp to getModification stamp

LorenzoBettini: > It's not a timestamp and thus _should_ not be used to implement `getTimestamp`
> 
> It might be fine to rename getTimestamp to getModification stamp

@szarnekow OK, I'll do that later.

That would be API breakage but I guess for that wouldn't be a problem because it's an internal type, would it?

LorenzoBettini: @szarnekow renamed it (https://github.com/eclipse-xtext/xtext/pull/3354/commits/7cc6b0d715b354580438c2f0fc0cafd07b627b80)

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
```diff
@@ -68,8 +68,8 @@ public AbsoluteURI getPath() {
 		}
 
 		@Override
-		public long getTimestamp() {
-			return file.getLocalTimeStamp();
+		public long getModificationstamp() {
```
szarnekow, line 71:
```suggestion
		public long getModificationStamp() {
```

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/internal/AbstractTraceForURIProvider.java
```diff
@@ -69,7 +69,7 @@ private static class CacheKey {
 
 			public CacheKey(PersistedTrace source) {
 				this.path = source.getPath();
-				this.timestamp = source.getTimestamp();
+				this.timestamp = source.getModificationstamp();
```
szarnekow, line 72:
```suggestion
				this.timestamp = source.getModificationStamp();
```

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/internal/AbstractTraceForURIProvider.java
```diff
@@ -54,7 +54,7 @@ public abstract class AbstractTraceForURIProvider<SomeFile, Trace extends Abstra
 	
 	protected interface PersistedTrace {
 		AbsoluteURI getPath();
-		long getTimestamp();
+		long getModificationstamp();
```
szarnekow, line 57:
```suggestion
		long getModificationStamp();
```

szarnekow (approved):

PR #3365, opened 2025-02-25
junit 5.12 workarounds for wizard

cdietrich:


LorenzoBettini (approved):
the target platforms don't require the fix, do they?

cdietrich:
i dont know. no builds failing yet

PR #3366, opened 2025-02-25
update to tycho 4.0.12

cdietrich:


LorenzoBettini:
@cdietrich sorry I was for reviewing.
Thanks for fixing that.

Issue #3337, opened 2025-02-05
Generated textmate grammar highlights punctuation characters as invalid


iTob191:
A textmate grammar generated by the [TextMateHighlightingFragment](https://github.com/eclipse-xtext/xtext/blob/2e919f42729871126793f1a1874c78d28dcc8b88/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/textmate/TextMateHighlightingFragment.java) does not contain a rule for punctuation. As a result, any punctuation characters are picked up by the `invalid` rule and highlighted accordingly.

I've looked at the code and basically every keyword that does not match `\w+` is ignored [here](https://github.com/eclipse-xtext/xtext/blob/2e919f42729871126793f1a1874c78d28dcc8b88/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/textmate/TextMateGrammar.java#L176) and will then be picked up by the `invalid` rule.

cdietrich:
@iTob191 can you work on a fix?

iTob191:
I'm not really sure how to best resolve this.

1. What textmate scope to use? The [naming conventions](https://macromates.com/manual/en/language_grammars#naming_conventions) aren't really clear regarding punctuation characters. VS Code uses different scopes for different punctuation characters, for example for typescript: `keyword.operator.ternary.ts`, `punctuation.separator.comma.ts`, `meta.brace.square.ts`, ...
2. Creating a punctuation rule for everything that is not matched [here](https://github.com/eclipse-xtext/xtext/blob/2e919f42729871126793f1a1874c78d28dcc8b88/org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/textmate/TextMateGrammar.java#L176) would be a simple solution to prevent valid tokens to be highlighted as errors. But it feels wrong to classify everything that does not match `\w+` as punctuation.

For my personal use case, this simple solution with a scope named `punctuation.languagename` would be sufficient. And I guess it would be sufficient for most languages and certainly an improvement over the current situation.

PR #3377, opened 2025-03-02
Add support for punctuation in generated textmate grammars [#3337]

iTob191:
I've implemented the change as outlined in https://github.com/eclipse-xtext/xtext/issues/3337#issuecomment-2671948388.


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/textmate/TextMateGrammar.java
```diff
@@ -173,15 +176,54 @@ protected TextMateRule getKeywordControlRule(Grammar grammar, boolean ignoreCase
 		matchBuilder.append("\\b(");
 		List<String> allKeywords = GrammarUtil.getAllKeywords(grammar)
 				.stream()
-				.filter(s->s.matches("\\w+"))
+				.filter(s -> s.matches("\\w+"))
 				.sorted(Comparator.naturalOrder())
 				.collect(Collectors.toList());
-		matchBuilder.append(Joiner.on("|").join(allKeywords));
+		Joiner.on("|").appendTo(matchBuilder, allKeywords);
```
szarnekow, line 182: Appreciated attention to detail

org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/textmate/TextMateGrammar.java
```diff
@@ -173,15 +176,54 @@ protected TextMateRule getKeywordControlRule(Grammar grammar, boolean ignoreCase
...
 		return result;
 	}
 
+	protected TextMateRule getPunctuationRule(Grammar grammar, boolean ignoreCase) {
+		StringBuilder matchBuilder = new StringBuilder();
+		if (ignoreCase) {
+			matchBuilder.append("(?i)");
+		}
+		matchBuilder.append("(");
+		List<String> allKeywords = GrammarUtil.getAllKeywords(grammar)
```
szarnekow, line 196: It seems as if we could use the same implementation for bother regex rules. The only meaningful difference seems to be the `s.matches("\\w+")` vs its negation and the name of the match rule.
Regular keywords should still work if we funnel them through `escapeAndAddWordBoundaries`

Would that work?

iTob191: That works and I've updated the PR accordingly. Only downside is that the generated regex for the `keyword.control` rule is less optimized: `(\\ba\\b|\\bb\\b|\\bc\\b)` instead of `\\b(a|b|c)\\b`

iTob191:
After extending my grammar with Xbase support which brings in a lot more punctuation characters, I've noticed that this PR breaks highlighting of comments. The problem is that Xbase introduces `/` as a keyword which is then picked up by the punctuation rule and prevents other rules further down (like the comment rules) from being chosen. To resolve this, I have pushed an updated version in which the punctuation rule is moved to the bottom.

PR #3378, opened 2025-03-03
Improvements for EcoreUtil2

tfonda-fbk:
A few minor fixes, cleanups and modernisations for EcoreUtil2.

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
@@ -580,7 +580,7 @@ public class EcoreUtil2 extends EcoreUtil {
 		if (ref == null) {
 			return null;
 		}
-		String result = exernalFormCache.get(ref);
+		String result = externalFormCache.get(ref);
 		if (result == null) {
 			EClass class1 = ref.getEContainingClass();
 			if (class1 == null) { // some references may be contained in an EAnnotation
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
```diff
@@ -378,13 +382,21 @@ private static List<EClass> getSortedCommonCompatibleTypeCandidates(EClass class
...
 
-		Collections.sort(result, new EClassTypeHierarchyComparator());
+		result.sort((class1, class2) -> {
+			if (class1.getEAllSuperTypes().contains(class2)) {
+				return -1;
+			}
+			if (class2.getEAllSuperTypes().contains(class1)) {
+				return 1;
+			}
+			return 0;
```
szarnekow, line 392: I've a nagging feeling that we should make the comparator exhaustive by comparing also the name and ePackage nsUri, if the containment checks in EAllSuperTypes are not conclusive.

szarnekow (approved):

org.eclipse.xtext/src/org/eclipse/xtext/EcoreUtil2.java
tfonda-fbk: @szarnekow Okay, thanks for the feedback! I haven't got the time to look into this right now, but I should be able to do it in a few days.

tfonda-fbk: @szarnekow I wasn't entirely sure what an EPackage nsUri was, but as I've seen it's just a String, I've extended the comparator with string comparisons for both the EClass names and EPackage nsUris as you requested. Please let me know if any adjustment is required. Thanks!

szarnekow: Yes, exactly.

szarnekow (approved):
It looks good.

PR #3403, opened 2025-04-09
bootstrap 2.39.0.M1

cdietrich:


LorenzoBettini:
Swtbot tests fail in Linux and windows, but not on Jenkins.. maybe the timeout must be increased

cdietrich:
@LorenzoBettini do you have any idea what the problem is with the tests

cdietrich:
ok. so we will have to adjust the tests in xtext-reference-projects too...

cdietrich:
created https://github.com/eclipse-xtext/xtext/issues/3404

cdietrich:
maybe i can find some time weekend nor next week.
until then: blindflug

LorenzoBettini:
Maybe using some kind of SWTBot matcher should be enough: something like `startsWith("Preferences")`

LorenzoBettini:
At least locally, this https://github.com/eclipse-xtext/xtext/pull/3405 fixes the problem for me.

PR #3409, opened 2025-04-15
gson update. guava update. closes #3408

### LINKED ISSUE #3408: The current gson version is no longer in current orbit

cdietrich:


cdietrich:
looks like there is yet another new guava :(

PR #3411, opened 2025-04-16
Add scrolling to AnnotationWithQuickFixesHover

MKamelEJ:
Add scrolling to the hover in order to allow displaying of larger text in the error hover

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AnnotationWithQuickFixesHover.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AnnotationWithQuickFixesHover.java
@@ -293,7 +293,7 @@ public class AnnotationWithQuickFixesHover extends AbstractProblemHover {
 
 		private void createAnnotationInformation(Composite parent, final Annotation annotation) {
 			Composite composite= new Composite(parent, SWT.NONE);
-			composite.setLayoutData(new GridData(SWT.FILL, SWT.TOP, true, false));
+			composite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
 			GridLayout layout= new GridLayout(2, false);
 			layout.marginHeight= 2;
 			layout.marginWidth= 2;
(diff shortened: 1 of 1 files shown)


cdietrich:
@mehmet-karaman @iloveeclipse can you please also check this with your codebase

cdietrich:
on mac i have scrollbars without this patch but do not with this patch

mehmet-karaman:
I am going to check that.

mehmet-karaman:
I am not sure what you that change does.. On my environment it works like usual.. Can you send a screenshot what you mean they disappear? In hover without F2 i don't see the scrollbars if I press F2 and resize the hover i am able to see the scrollbars.. Maybe it differs between linux and mac.

My environment:
OS: redhat 9.2
Version: 2025-06 (4.36)
Build id: I20250415-1800

LorenzoBettini:
But to use the scrollbar (after the change), you have to focus on the popup, do you?
When you focus, the scrollbar will appear anyway, won't it?

cdietrich:
no even on focus i dont see any


https://github.com/user-attachments/assets/ca95aa3d-bd0f-44eb-93e8-7a4db3185236

cdietrich:
on main: 


https://github.com/user-attachments/assets/be3800d5-48a4-4e72-aef1-2fa643f8d93c

mehmet-karaman:
> even if it wasn't necessary..

Yes, If i click on the Toolbar, it behaves like I've pushed the F2 key.. But in my case I even didn't needed a Scrollbar.. The original space was enough.

LorenzoBettini:
So if I understand correctly, the PR does not provide additional utility and it breaks the existing behavior at least on mac.

mehmet-karaman:
Why was it necessary to provide this fix? Was there a bug regarding this? I've never felt like it was defect.. Could it be that on windows it behaves different?

Abozekry:
I tried and it seems for large annotation text, user friendly scrolling is needed
Screenshot_before_For_Github
Screenshot_after_For_Github

cdietrich:
so i guess we need a splution that works for both the area with the message and the area with the quickfixes

szarnekow:
@Abozekry @MKamelEJ how shall we proceed with this? Do you plan to refine the PR?

PR #3413, opened 2025-04-24
catch NoClassDefFoundError

### LINKED ISSUE #3381: JDT dependency added in SyncUtil and potentially crashing languages that do not include JDT
I have tested v. 2.38 as part of Eclipse 2025-03 and a dependency to JDT has been introduced in `SyncUtil` (see here: https://github.com/eclipse-xtext/xtext/commit/2f2fe5d60e523ec4b9c5d9a12e96d172833499ea#commitcomment-153274022), which will probably make any DSL not dependent from JDT crash during refactoring operations. It did so for the GAML language (see https://github.com/gama-platform/gama), where JDT has always been treated as an optional dependency. 

Hopefully (thanks @LorenzoBettini)  a workaround can be implemented until this is fixed. It consists in subclassing `SyncUtil` like below and binding this new implementation (`		binder.bind(SyncUtil.class).to(GamlSyncUtil.class);`) in the `UIModule` of your language. 

E.g. 

```java
public class GamlSyncUtil extends SyncUtil {

	/** The workspace. */
	@Inject (optional = true) private IWorkspace workspace;

	/**
	 * Wait for build.
	 *
	 * @param monitor
	 *            the monitor
	 */
	@Override
	public void waitForBuild(final IProgressMonitor monitor) {
		try {
			// waitForJdtIndex(monitor); -- commented out
			workspace.build(IncrementalProjectBuilder.INCREMENTAL_BUILD, monitor);
		} catch (CoreException e) {
			throw new OperationCanceledException(e.getMessage());
		}
	}

}
```

LorenzoBettini:
when jdt.core is not available

Closes #3381

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/ui/SyncUtil.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/ui/SyncUtil.java
@@ -158,7 +158,12 @@ public class SyncUtil {
 
 	@SuppressWarnings("restriction")
 	public void waitForJdtIndex(IProgressMonitor monitor) {
-		JavaModelManager.getIndexManager().waitForIndex(true, monitor);
+		try {
+			JavaModelManager.getIndexManager().waitForIndex(true, monitor);
+		} catch (NoClassDefFoundError e) {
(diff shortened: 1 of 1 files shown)


cdietrich:
guess we need to wait until the total eclipse infra collapse has resolved

PR #3433, opened 2025-05-28
bootstrap 2.39.0

LorenzoBettini:


cdietrich:
@LorenzoBettini should we prepare 2.40 first?

Issue #3438, opened 2025-06-02
Non blocking validation processing


mehmet-karaman:
Hi, 

I've a question on the following two methods: 
org.eclipse.xtext.validation.AbstractDeclarativeValidator.internalValidate(...) 
AbstractInjectableValidator.validate(...)

Do they have to be final? The synchronize is blocking in our case, which we would like to override with our own functionality. Is there an excuse why its final? I mean for the most of the cases it makes sense to keep the synchronized, but we would like to adjust this on our own risk. Thank you in advance.

Best regards
Mehmet

cdietrich:
@mehmet-karaman which validation callstack are you talking about, and which part is blocking
the collect methods should be done once only

mehmet-karaman:
We will shift this problem for a few days. I guess we need a working experiment, maybe it can improve the performance for other xtext users too.. 

The performance bottleneck in our case seems to be in org.eclipse.xtext.validation.AbstractDeclarativeValidator.methodsForType which we could try to fix when the callee wasn't final.. But first we will do some experiments and profile it again to check if it helps..

szarnekow:
> The synchronize is blocking in our case, which we would like to override with our own functionality. Is there an excuse why its final?

What would be different in your own impl? Can you double-check that your validators are actual singletons?

mehmet-karaman:
> > The synchronize is blocking in our case, which we would like to override with our own functionality. Is there an excuse why its final?
> 
> What would be different in your own impl? Can you double-check that your validators are actual singletons?

Our validator is also a singleton, but we have a huge workspace where the validation takes much time. Currently we experiment with the methodsForType cache (try to optimize or replace with optimized implementation), if we have results I will share it here again.

cdietrich:
but the cache filling should be constant, if you have 1 or 100000 files

mehmet-karaman:
We had time to investigate the bottleneck and we found a way to avoid this. To remember: we have a workspace with millions of eobjects and we are building the workspace in parallel with many threads (we have also many cpu cores to benefit from that).

The problem we have is that the validation cache (methodsForType) is initialized on demand, but if we execute a bunch of threads they are locked in the readWriteLock (methodsForType.get().. ). If the type or subtype wasn't asked before, its aquiring the lock and blocks all of the reading threads.

What we basically can do is just initialize the methodsForType cache once for all our EClasses in our package and then start all  threads for validation.

[...]

Issue #3398, opened 2025-04-05
Deal with sunset of OSSRH


cdietrich:
https://central.sonatype.org/news/20250326_ossrh_sunset/

LorenzoBettini:
I seem to understand that we need a new token (the eclipse foundation should provide that, shouldn't it)

cdietrich:
I don’t know and I don’t have time to search unsearchable gitlab or reach out etc

cdietrich:
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5913

cdietrich:
https://www.eclipse.org/lists/cross-project-issues-dev/msg20050.html

cdietrich:
unfortunately my time is still very very limited

LorenzoBettini:
@cdietrich @szarnekow 

Some updates concerning the actual release process.

I migrated my own groupId and my own artifacts to the new Central repository.

Concerning the actual procedure, it's just a matter of using the new plugin:

```xml
<plugin>
    <groupId>org.sonatype.central</groupId>
    <artifactId>central-publishing-maven-plugin</artifactId>
    <version>0.7.0</version>
    <extensions>true</extensions>
    <configuration>
        <publishingServerId>central</publishingServerId>
    </configuration>
</plugin>
```

And it will take care of everything.

`<distributionManagement>` section must be removed, since the plugin already deploys to the new Centra by default.

In that configuration, artifacts are not effectively published: Once deployed, the deployment must be released from here <https://central.sonatype.com/publishing/deployments>.
That's quite easy.
For sure, much easier than the old web interface ;)

From what I understand, this cannot be consumed as a staging repository, but we never used that feature anyway.

The plugin above can also be configured to do the release automatically, in case.

Unlike the previous deployment, this plugin seems to do the actual uploading in the end in bulk.

The plugin also handles SNAPSHOT deployments; again, there's no need to configure the distributionManagement.

For consuming the snapshots from the new central, this is the needed section:

```xml
	<repositories>
		<repository>
			<name>Central Portal Snapshots</name>
			<id>central-portal-snapshots</id>
			<url>https://central.sonatype.com/repository/maven-snapshots/</url>
			<releases>
				<enabled>false</enabled>
			</releases>
			<snapshots>
				<enabled>true</enabled>
			</snapshots>
		</repository>
	</repositories>
```

[...]

PR #3442, opened 2025-06-05
Non blocking validation processing - #3438

mehmet-karaman:
- Added initialization for methodForType cache in the validator.
- Implemented a test case to approve the methods are exposed/available.


org.eclipse.xtext/src/org/eclipse/xtext/validation/AbstractDeclarativeValidator.java
```diff
@@ -333,13 +326,33 @@ protected final boolean internalValidate(EClass class1, EObject object, Diagnost
...
+				if (checkMethods == null) {
+					Set<MethodWrapper> checkMethods = Sets.newLinkedHashSet();
+					checkMethods.addAll(collectMethods(getClass()));
+					this.checkMethods = checkMethods;
+				}
+			}
+		}
+	}
+	
+	protected List<MethodWrapper> initializeMethodsForType(Class<?> clazz) {
```
iloveeclipse: This is unused and duplicates `getMethodsForType()`. Did you missed some change to commit?

mehmet-karaman: Yes .. there is an amend commit missing.. going to fix that.. and the commit message..

iloveeclipse (approved):
I'm not the code owner and can't say whether it would be safe to expose the code in question. 

At least some comments on the exposed methods could be added to explain the purpose and **why** they are exposed at all. 

Should they be overridden, could they be overridden or should they be only overridden in some extra cases...

Beside this the PR seem not to change any behavior, and if it can be useful, why not?

LorenzoBettini:
"since" annotations should also be added.

mehmet-karaman:
Does this PR have some problems or is this waiting until xtext is released?

cdietrich:
as you committer you can merge appoved prs.
but there is the big  migration for maven
so i dont know if the build with deployment still works
see
https://github.com/eclipse-xtext/xtext/issues/3398

mehmet-karaman:
I am not committer for the xtext repository (only for xtend I have committer rights). Thanks for the hint, i will ping again if #3398 is done.

cdietrich:
xtend and xtext are the same repo.you mean xpect.
will talk to the team if it makes sense to nomminate you (if it is ok for you)
as i dont know what the current status of the migration is foundation wise ill keep this one open for now

mehmet-karaman:
Sorry, I meant xpect :).. Sure, I would be happy to be also a commiter for the xtext repository.

org.eclipse.xtext/src/org/eclipse/xtext/validation/AbstractDeclarativeValidator.java
```diff
@@ -333,13 +326,49 @@ protected final boolean internalValidate(EClass class1, EObject object, Diagnost
...
+	/**
+	 * This method can be called to initialize the cache. The initialization of the cache will 
+	 * improve the validation performance, for the cases where the validation is executed in many threads.
+	 * The caching of non initialized types are blocking all other reading threads. After initialization
+	 * the readWriteLock won't block any accessing threads.
+	 * 
+	 * To initialize the cache, this method has to be called with every Class (and its subtypes) which is handled
+	 * by this validator.
+	 * 
+	 * If its called internally it doesn't have to be overriden/exposed.
```
szarnekow:
```suggestion
	 * If it's called internally it doesn't have to be overriden/exposed. 
```

org.eclipse.xtext/src/org/eclipse/xtext/validation/AbstractDeclarativeValidator.java
```diff
@@ -316,15 +316,8 @@ protected CancelIndicator getCancelIndicator() {
...
-		if (checkMethods == null) {
-			synchronized (this) {
-				if (checkMethods == null) {
-					Set<MethodWrapper> checkMethods = Sets.newLinkedHashSet();
-					checkMethods.addAll(collectMethods(getClass()));
-					this.checkMethods = checkMethods;
-				}
-			}
-		}
+		initializeCheckMethods();
```
szarnekow: This invocation can be moved into getMethodsForType, I suppose.
This would allow for a replacement of the simple cache that does not have the init call for each internalValidate

org.eclipse.xtext/src/org/eclipse/xtext/validation/AbstractDeclarativeValidator.java
```diff
@@ -333,13 +326,49 @@ protected final boolean internalValidate(EClass class1, EObject object, Diagnost
...
+	 * 
+	 * To initialize the cache, this method has to be called with every Class (and its subtypes) which is handled
+	 * by this validator.
+	 * 
+	 * If its called internally it doesn't have to be overriden/exposed. 
+	 * If its wanted to be called outside of the Validator it can be overriden and exposed.
+	 * 
+	 * @since 2.37
+	 */
+	protected List<MethodWrapper> getMethodsForType(Class<?> object) {
```
szarnekow: I'd probably move the EObject.getClass() into this impl. That would allow for querying MethodWrappers by EObject/EClass instead of Java Class object. Potentially one could build a faster lookup based in the int EClass ID.

szarnekow: Since the new API returns MethodWrapper, it's probably best to make the MethodWrapper constructor accessible, too.

org.eclipse.xtext/src/org/eclipse/xtext/validation/AbstractDeclarativeValidator.java
mehmet-karaman: For the replacement the cache field has to be also replaceable/exposed.. or not?

org.eclipse.xtext/src/org/eclipse/xtext/validation/AbstractDeclarativeValidator.java
mehmet-karaman: Sounds good.. I will replace Class<?> by eobject and gather the class inside of this method..

mehmet-karaman:
> A few more thoughts. What do you think about the ideas?

Looks good to me.. I like solutions which help more users than only me :)

mehmet-karaman:
Rebase done.  @cdietrich Can we merge this PR?

cdietrich:
we first should fix the maven deployment

org.eclipse.xtext/src/org/eclipse/xtext/validation/AbstractDeclarativeValidator.java
```diff
@@ -333,13 +324,50 @@ protected final boolean internalValidate(EClass class1, EObject object, Diagnost
...
+	protected List<MethodWrapper> getMethodsForType(EObject object) {
+		initializeCheckMethods();
+		
+		return methodsForType.get(object.getClass());
+	}
+
+	/**
+	 * Initializes the check method wrappers.
+	 *  
+	 * @since 2.37
```
cdietrich: 2.40

org.eclipse.xtext/src/org/eclipse/xtext/validation/AbstractDeclarativeValidator.java
```diff
@@ -333,13 +324,50 @@ protected final boolean internalValidate(EClass class1, EObject object, Diagnost
...
+	 * The caching of non initialized types are blocking all other reading threads. After initialization
+	 * the readWriteLock won't block any accessing threads.
+	 * 
+	 * To initialize the cache, this method has to be called with every Class (and its subtypes) which is handled
+	 * by this validator.
+	 * 
+	 * If it's called internally it doesn't have to be overriden/exposed. 
+	 * If it's wanted to be called outside of the Validator it can be overriden and exposed.
+	 * 
+	 * @since 2.37
```
cdietrich: 2.40

LorenzoBettini:
@cdietrich shall we merge this before M0?

PR #3444, opened 2025-06-16
Publish to new Maven Central

### LINKED ISSUE #3398: Deal with sunset of OSSRH

LorenzoBettini:
Closes #3398 

- added missing `<name>` for deployable artifacts
- removed stale information (e.g., distribution management)
- added also the new Maven Central snapshot repository (besides the old one)
- updated the wizard
- exclude artifacts that we don't want to deploy (the only exceptions are the features with the same artifact id as the bundles that must be deployed, see https://github.com/eclipse-xtext/xtext/issues/3398#issuecomment-2970520389)

**WARNING**: I have the feeling that the main plugin configuration must be done in the aggregator project, `full.releng`, but I'm not sure about that. The documentation is not clear about that, and in my projects the parent is also the aggregator, while it's not the case in this project.
I'm afraid the only way to check that is to merge this PR and see which snapshot artifacts are deployed from JIRO.
For this, we should merge this PR ASAP.


org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
```diff
@@ -457,6 +457,17 @@ class ParentProjectDescriptor extends ProjectDescriptor {
 						</repository>
```
cdietrich, line 460: whaout about other places e.g. line 98, 450, 473

org.eclipse.xtext.full.releng/pom.xml
```diff
@@ -162,4 +162,246 @@
...
+				<plugins>
+					<plugin>
+						<groupId>org.sonatype.central</groupId>
+						<artifactId>central-publishing-maven-plugin</artifactId>
+						<version>0.7.0</version>
+						<extensions>true</extensions>
+						<configuration>
+							<publishingServerId>central</publishingServerId>
+							<excludeArtifacts>
+								<excludeArtifact>
```
cdietrich, line 181: wonder how to check this list below

cdietrich:
you mentioned a problem with duplicate artifact ids, is this resovled?

org.eclipse.xtext.full.releng/pom.xml
```diff
@@ -162,4 +162,246 @@
...
+									org.eclipse.xtend.lib.tests
+								</excludeArtifact>
+								<excludeArtifact>
+									org.eclipse.xtend.m2e
+								</excludeArtifact>
+								<excludeArtifact>
+									org.eclipse.xtend.sdk
+								</excludeArtifact>
+								<excludeArtifact>
+									org.eclipse.xtend.standalone
```
cdietrich, line 212: wired name @szarnekow

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
LorenzoBettini: Concerning gradle, I had missed that one. Done that in https://github.com/eclipse-xtext/xtext/pull/3444/commits/6842ac913eb7d4c0e3e5189dbde2be05782bb4ff

Concerning the disabling of snapshot repositories (line 450 https://github.com/eclipse-xtext/xtext/blob/47c9c9d6bd98b502707c23538c5df091f071a009/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend#L450) that was required due to some artifacts/plugins; that's not required for the new Central snapshot repository, because the new snapshot repository is not dragged in automatically by the culprit plugins/dependencies.

Concerning line 473: as in other parts, including our POM, I enabled the new snapshot repository but also keep the old one because not every snapshot artifacts will be immediately migrated to the new one.

org.eclipse.xtext.full.releng/pom.xml
LorenzoBettini: As I said above:

> I'm afraid the only way to check that is to merge this PR and see which snapshot artifacts are deployed from JIRO.
For this, we should merge this PR ASAP.

After all, they're just snapshots

LorenzoBettini:
> you mentioned a problem with duplicate artifact ids, is this resovled?

No, not yet. For the time being, we'll deploy also 3 artifacts related to features. Only 3 of them are affected by the problem: if we skipped them, we'd skip also the homonimous bundles.

org.eclipse.xtext.full.releng/pom.xml
LorenzoBettini: Did you mean "weird"?
That bundle is now empty and hasn't been deployed to Maven for a long time.

LorenzoBettini:
@cdietrich if you don't mind, I'll merge this PR and see what happens in Jenkins.

cdietrich:
yes
but i am still confused about the wizards
see

https://github.com/xtext/xtext-reference-projects/commit/3f83f81157611d15f67cead3a290d5f5ff583998

org.eclipse.xtext.full.releng/pom.xml
cdietrich: its called standlone

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
cdietrich: https://github.com/xtext/xtext-reference-projects/commit/3f83f81157611d15f67cead3a290d5f5ff583998

there i update more files as you did in the test expectation

LorenzoBettini: The wizard tests never use snapshots so expectations didn't need an update. If that's what you meant.

org.eclipse.xtext.full.releng/pom.xml
LorenzoBettini: IIRC, that was used long ago before the maven plug-in. It's now empty and contains only dependencies. I think it was also deprecated.

LorenzoBettini:
It fails with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:3.0.0:deploy (default-deploy) on project xtext-parent: Deployment failed: repository element was not specified in the POM inside distributionManagement element or in -DaltDeploymentRepository=id::url parameter
```

So it does not pick the new plugin correctly.

I'll have to look at that.

I'll do further experiments, but I'll have to merge to master then.

cdietrich:
@LorenzoBettini in the root i see
central-publishing-maven-plugin
in the pluginMgmt only and not in the build

how does the full deploy parent interit stuff to that one?

LorenzoBettini:
> @LorenzoBettini in the root i see central-publishing-maven-plugin in the pluginMgmt only and not in the build
> 
> how does the full deploy parent interit stuff to that one?

That indeed was my mistake and my misunderstanding.

I thought that with pluginManagement section, when the plugin kicked in from the releng it would work; again, my misunderstanding due to the poor documentation.

I've merged https://github.com/eclipse-xtext/xtext/pull/3446 and it goes better, but now there are problems from the Maven Central side I'm afraid.

PR #3446, opened 2025-06-16
Publish to new Maven Central (2nd attempt)

LorenzoBettini:


LorenzoBettini:
Now the new plugin kicks in correctly, and it seems to skip the artifacts as requested.

It fails in the end with

```
[ERROR] Failed to execute goal org.sonatype.central:central-publishing-maven-plugin:0.7.0:publish (injected-central-publishing) on project org.eclipse.xtext.p2repository: org.apache.maven.artifact.deployer.ArtifactDeploymentException: Failed to deploy artifacts: Could not find artifact org.eclipse.xtext:org.eclipse.xtext.purexbase:xml.asc:p2artifacts:2.40.0-20250616.191026-1 in central (https://central.sonatype.com/repository/maven-snapshots/)
```

But even from my local machine I cannot access snapshots from that URL.
I guess they have a few problems right now.
Even https://central.sonatype.com/service/rest/repository/browse/maven-snapshots gives a 404.

Guess we have to try again tomorrow.

PR #3448, opened 2025-06-19
bump to clipse.org/releases/2025-09

cdietrich:


org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/pom.xml
```diff
@@ -242,6 +242,17 @@
 			<releases><enabled>false</enabled></releases>
 			<snapshots><enabled>true</enabled></snapshots>
 		</repository>
+		<repository>
```
cdietrich, line 245: @LorenzoBettini i found this one was missing in your orignal pr
see my comment on the ref projects pr

LorenzoBettini: @cdietrich 

It was missing because I thought snapshots were disabled in domainmodel.

Indeed they are a few lines above: https://github.com/eclipse-xtext/xtext/blob/f1d94c3351725b4b0e8ead8ec5c4c0af1bbdd8d5/org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/pom.xml#L215
where we have:

```xml
		<repository>
			<id>ossrh-snapshots</id>
			<name>ossrh-snapshots</name>
			<releases>
				<enabled>false</enabled>
			</releases>
			<snapshots>
				<enabled>false</enabled>
			</snapshots>
			<url>http://oss.sonatype.org/content/repositories/snapshots</url>
		</repository>
```

I had seen that part and thought snapshots were disabled.

So there's still a mistake: either they are enabled and the above part must be removed, or they are disabled so a few lines below they must be disabled as well (and the additional central snapshots are not needed).

cdietrich: Am Even more confused we have it multiple times with different ids maybe Claude Bad references Somewhere

LorenzoBettini: "Claude"?

Anyway, yes, that's the mistake. I seem to understand that in the Xtext git repository domainmodel is without snapshots enabled. Snapshots for domainmodel are used only in Xtext reference projects, at least that was my understanding.

Note that when they are disabled above there's a comment (by me): we were explicitly disabling because of a bad behavior of the old exec-maven-plugin.

cdietrich: Smartphone autocorrect. Cause . We had to disable some repos . Don’t remember why . Commit with false false might explain. GitHub blames an smartphones bad idea

cdietrich: the comment indeed says the false false pairs are there to prevent some transitive mess (i assume by id)

PR #3454, opened 2025-06-23
Parent POM for features that must be published as Maven artifacts

LorenzoBettini:


LorenzoBettini:
I also specified in the description that it is not meant to be used as a Maven dependency.

I'd like to merge this right away and try again with a dry run deployment.

LorenzoBettini:
@cdietrich I'm merging and try another dry-run

cdietrich (approved):

PR #3458, opened 2025-06-24
updated for the org.eclipse.xtext groupId

### LINKED ISSUE #3455: Ref Project Wizard Test Problems
xtextProject_webIntegration_withMaven_plain
fails with the newest snapshot

cc @LorenzoBettini

LorenzoBettini:
Closes #3455

Unfortunately, I cannot put 2.40.0 in the version range, because we're still use SNAPSHOT

### DIFF
--- a/org.eclipse.xtend.m2e/lifecycle-mapping-metadata.xml
+++ b/org.eclipse.xtend.m2e/lifecycle-mapping-metadata.xml
@@ -2,7 +2,7 @@
   <pluginExecutions>
     <pluginExecution>
       <pluginExecutionFilter>
-        <groupId>org.eclipse.xtend</groupId>
+        <groupId>org.eclipse.xtext</groupId>
         <artifactId>xtend-maven-plugin</artifactId>
         <versionRange>[2.2.0,)</versionRange>
         <goals>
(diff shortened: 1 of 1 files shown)


org.eclipse.xtend.m2e/lifecycle-mapping-metadata.xml
```diff
@@ -2,7 +2,7 @@
   <pluginExecutions>
     <pluginExecution>
       <pluginExecutionFilter>
-        <groupId>org.eclipse.xtend</groupId>
+        <groupId>org.eclipse.xtext</groupId>
         <artifactId>xtend-maven-plugin</artifactId>
```
cdietrich, line 6: @HannesWell
i wonder if we should add two entries here, one with the old and one with the new groupId
would this work in backwards comp. way?

HannesWell: > @HannesWell i wonder if we should add two entries here, one with the old and one with the new groupId would this work in backwards comp. way?

That should probably work, although I'm not absolutely sure if there is an error if the plugin is not found.
But you can check that very quickly when you implement this.

LorenzoBettini:
Thank you @HannesWell . I tried as you said and it does not complain about the old plugin not found.

@cdietrich I've updated the PR accordingly. I also added the two other plugins' goals for exclusion (`xtend-test-install-debug-info` and `xtend-install-debug-info`), to avoid warnings on the POM.

szarnekow (approved):
Let's give it a try.

PR #3465, opened 2025-07-03
Allow usage of Read/Write Request with an own RequestManager

rubenporras:


rubenporras:
For https://github.com/eclipse-xtext/xtext/issues/3456

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.java
```diff
@@ -55,25 +53,17 @@ public void shutdown() {
 		parallel.shutdown();
 		cancel();
 	}
-
```
cdietrich, line 69: @szarnekow now do we deal with these removals api wise?

szarnekow: I'm wondering the same

szarnekow: Question: Do you think it'd help to keep the overloaded constructors in the ReadRequest and WriteRequest with a RequestManager instead of an AbstractRequestManager?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/AbstractRequestManager.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+	
+	protected final OperationCanceledManager getOperationCanceledManager() {
+		return operationCanceledManager;
+	}
+
+	/**
+	 * Check if the given throwable is an indicator for a cancellation.
+	 */
+	public boolean isCancelException(Throwable t) {
```
szarnekow: Preferably protected.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.java
rubenporras: Are you thinking about an API removal because we break with this PR binary compatibility?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/IRequestManager.java
```diff
@@ -0,0 +1,36 @@
...
+import org.eclipse.xtext.xbase.lib.Functions.Function0;
+import org.eclipse.xtext.xbase.lib.Functions.Function1;
+import org.eclipse.xtext.xbase.lib.Functions.Function2;
+
+/**
+ * @author Rub�n Porras Campo - Initial Contribution and API
+ * 
+ * @since 2.40
+ */
+public interface IRequestManager {
```
szarnekow, line 23: Now we should start using IRequestManager instead of RequestManager in LanguageServerImpl?

Should this interface be annotated with `@ImplementedBy(AbstractRequestManager.class)`

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/AbstractRequestManager.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.CompletionException;
+
+import org.eclipse.xtext.service.OperationCanceledManager;
+
+/**
+ * @author Rub�n Porras Campo - Initial Contribution and API
+ *
+ * @since 2.40
+ */
+public abstract class AbstractRequestManager implements IRequestManager {
```
szarnekow, line 20: Should this class be annotated with @ImplementedBy(RequestManager.class)

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/IRequestManager.java
rubenporras: I can do that in the next iteration of the PR

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.java
szarnekow: Yes. I think we need to make a judgement call here. I believe it's ok to break binary compatibility for the hierarchy around AbstractRequest

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/IRequestManager.java
rubenporras: You are right that there should be a binding (to the `RequestManager`, not the `AbstractRequestManager`, the later cannot be injected). I prefer to put the binding in the module and that is part of my updated PR.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.java
rubenporras: Are you any further with this?

szarnekow: We had to postpone the committer call to Friday this week. I'll get back to you

szarnekow: We are fine with this binary incompatibility. Adopters will need to make a minor adjustment to their code if they are already using a custom implementation in rare cases. That’s ok.

rubenporras: Hi @szarnekow ,

Is there anything to do on this PR? I think I have already addressed all your comments in the last iteration of the PR on the 14th of July.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ServerModule.java
```diff
@@ -32,5 +34,6 @@ protected void configure() {
 		bind(IMultiRootWorkspaceConfigFactory.class).to(MultiRootWorkspaceConfigFactory.class);
 		bind(IProjectDescriptionFactory.class).to(DefaultProjectDescriptionFactory.class);
 		bind(IContainer.Manager.class).to(ProjectDescriptionBasedContainerManager.class);
+		bind(IRequestManager.class).to(RequestManager.class);
```
szarnekow, line 37: Smol refinement:
Please bind IRequestManager to the new AbstractRequestManager and that one to RequestManager.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/concurrent/RequestManager.java
szarnekow: One point remaining about the Guice bindings.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/ServerModule.java
rubenporras: @szarnekow: may I ask what is the advantage of binding the `AbstractRequestManager` and that one to `RequestManager` compared to the most straightforward version of directly binding the `RequestManager`?

szarnekow: It avoids an unfortunately common pitfall. Instead of injecting the interface in a hierarchy Class -> AbstractClass -> Interface, people sometimes accidentally inject the base class.

That being said, it would lead to a runtime error since the base class is abstract in this case. So I'll merge as is.

rubenporras: Thanks for the explanation.

PR #3477, opened 2025-07-28
Demote error to warning in TextRegionAccessBuildingSequencer and AbstractDeclarativeFormatter

tfonda-fbk:
Just a proposal for a very minor improvement. This really looks like a warning rather than an error - it warns the user that things *might* go south later on, but it does not prevent the rest of the method from being executed anyway. Do you agree?


cdietrich:
am not sure about this.
what about the 2nd place

github.com/eclipse-xtext/xtext/commit/9158fdbed0784aa44839500a093be8f0380febe4

tfonda-fbk:
> am not sure about this. what about the 2nd place
> 
> github.com/eclipse-xtext/xtext/commit/9158fdbed0784aa44839500a093be8f0380febe4

@cdietrich for uniformity I'd demote the message to warning even in the other place, since it has no immediate impact on the execution of the method it is contained in.

cdietrich:
please do so

tfonda-fbk:
@cdietrich done!

szarnekow (approved):

PR #3479, opened 2025-07-29
Fixed broken isAffected computation for bulk-changes

szarnekow:


cdietrich:
@mehmet-karaman can you check this is your env too

cdietrich:
@szarnekow did you check if the namesareunique stuff still works

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/builder/XtendUIResourceDescriptionManager.java
```diff
@@ -70,7 +70,7 @@ public boolean isAffected(Collection<Delta> deltas, IResourceDescription candida
 						descriptionIsContained = containers.get(i).hasResourceDescription(uri);
 					}
 					if (!descriptionIsContained && !isProjectDependency(uri, candidate.getURI(), checkedProjects)) {
-						return false;
```
cdietrich, line 73: do you remember we have this slightly different impl at all

szarnekow:
> @szarnekow did you check if the namesareunique stuff still works

Yes, it does. It'll call the method in line 165

org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/builder/XtendUIResourceDescriptionManager.java
szarnekow: In Xtend we have to be affected by Java changes in related projects. The type URIs on their own do not play well with the regular container manager stuff. That's why it's a custom solution.

mehmet-karaman:
@cdietrich The Xtext tests are still runing in my environment (nearby two hours now) and there is no failure regarding that change.. I guess it will take a few hours more. I will write again when all tests are executed.

mehmet-karaman:
Executed all component tests in our environment and no failures or errors.

cdietrich (approved):

PR #3483, opened 2025-08-14
check the cancel indicator before creating the hover

rubenporras:
this can happen as the createContext method may take long while resolving cross-references, and in this time the client might have cancelled the request.


rubenporras:
for https://github.com/eclipse-xtext/xtext/issues/3482

szarnekow (approved):

PR #3527, opened 2025-10-10
Update Xtext to use ASM 9.9.0

iloveeclipse:
Fixes https://github.com/eclipse-xtext/xtext/issues/3525


iloveeclipse:
I have no idea what is going on, the commit has wrong author, wrong target.

PR #3530, opened 2025-10-13
Update Xtext to use ASM 9.9.0

merks:
- Restrict the junit version in the target platform to 5.x dependencies.

Fixes https://github.com/eclipse-xtext/xtext/issues/3525

See also https://github.com/eclipse-xtext/xtext/pull/3528


cdietrich:
also misses change in org.eclipse.xtext.dev-bom/pom.xml

cdietrich:
/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TargetPlatformProject.xtend
also has the 0.0.0

so where is the bad one coming from?
or is the problem that 2025-12+I-builds has a mixture of both

merks:
My workspace is dirty including that file you mention.

image

I only added the one *.target file change to see if that fixes the jenkins build.  Depending on that we can try to move forward.

merks:
I think the failing test has already passed in the Jenkins build:

image

I think you are saying that those ranges also need to be here:

image

Of course the other *.target files need this change too:

For https://github.com/eclipse-xtext/xtext/pull/3530#issuecomment-3396730290 I think you are saying these version should match exactly the current version is the current latest Orbit aggregation?

Please correct if I got any of that wrong.

merks:
> so where is the bad one coming from?
> or is the problem that 2025-12+I-builds has a mixture of both

I answered that already here in https://github.com/eclipse-xtext/xtext/pull/3528#issuecomment-3396713807

https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/table.html

image

LorenzoBettini:
> My workspace is dirty including that file you mention.

@merks unfortunately some of the dirty files are related to the Xtend compiler and are not to be considered.
We still haven't found a way to avoid that.

merks:
The GHA failures look unrelated to asm and junit changes. Perhaps a change in jdt or the platform? The Jenkins build was successful.

cdietrich:
dont know whats broken/changed in platform / jdt that could break AutoEditTest
somehow an enter does no longer auto insert a star

merks:
So is this a blocker to merging this PR?  It seems to me orthogonal and something that can be dealt with later, time/resource permitting.

cdietrich:
i dont have time or infrastructure to investigate ..

cdietrich:
@LorenzoBettini @szarnekow wdyt?

cdietrich:
have started https://ci.eclipse.org/xtext/job/xtext/job/PR-3530/4/ do see proper test results.
these logs from github are unreadanle and i cannot find the link to the rest report for gh actions anymore

cdietrich:
it looks like its broken at least since the 9th
https://ci.eclipse.org/xtext/job/xtext/job/main/2450/

update: 8th
https://ci.eclipse.org/xtext/job/xtext/job/cd_tycho50/935/

LorenzoBettini:
This is the original bug related to those failing tests: https://bugs.eclipse.org/bugs/show_bug.cgi?id=453205

Can we ask someone from platform?

cdietrich:
looking at platform.ui log did not spot anything obvious

merks:
FYI, test results are here:

image

https://github.com/eclipse-xtext/xtext/pull/3530/checks?check_run_id=52627697324

image

It seems relatively readable if somewhat hard to find.

cdietrich:
maybe https://github.com/eclipse-platform/eclipse.platform.ui/commit/1848058a52102b5deff39427365e56fe882656b3

cdietrich:
created https://github.com/eclipse-xtext/xtext/issues/3531

cdietrich:
thx @iloveeclipse @merks 

@iloveeclipse do you know about the platform change around projection viewer?

releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -33,12 +33,12 @@
   </annotation>
   <setupTask
       xsi:type="jdt:JRETask"
-      version="JavaSE-17"
-      location="${jre.location-17}"/>
+      version="JavaSE-21"
```
LorenzoBettini, line 36: Sorry for being late.

I noted this change in the Oomph configuration.

I personally like it and always use Java 21 in my workspace.

However, I seem to recall that @cdietrich or @szarnekow preferred to stick with Java 17 which is the minimal requirement.

Maybe I don't remember correctly, but just to be sure...

cdietrich: i keep n laptops with different setups to be able to reproduce problems with different tps

cdietrich: what is the effect of that version property if you select 2024-03?

LorenzoBettini: it shouldn't hurt but I don't know...

LorenzoBettini: However, if the Oomph setup only requires Java 17, one is entitled to specify only a Java 17 JRE, and then, when using one of the latest TPs, most tests won't work and you won't be able to start a runtime workspace.

merks:
Given that large parts of the Platform itself are now at Java 21, I'm doubtful that Java 17 launching is workable when using a latest target platform.

Of course one can always added a 17 and set it as the default, for for "latest" I think it needs to be this way.

cdietrich:
no its not
therefor in the old tps/workspaces i use 17 and in the new one 21

LorenzoBettini:
> Given that large parts of the Platform itself are now at Java 21, I'm doubtful that Java 17 launching is workable when using a latest target platform.
> 
> Of course one can always added a 17 and set it as the default, for for "latest" I think it needs to be this way.

I agree.

PR #3547, opened 2025-11-13
Fixed compile problem revealed by latest JDT

### LINKED ISSUE #3542: Eclipse shows Errors with latest platform
Description	Resource	Path	Location	Type
Name clash: The method getTraceToSource(PlatformResource) of type IPlatformSpecificTraceProvider<PlatformResource,Trace> has the same erasure as getTraceToSource(SomeFile) of type AbstractTraceForURIProvider<SomeFile,Trace> but does not override it	TraceForStorageProvider.java	/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace	line 40	Java Problem

szarnekow:
Closes #3542

### DIFF
--- a/org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/internal/AbstractTraceForURIProvider.java
+++ b/org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/internal/AbstractTraceForURIProvider.java
@@ -186,16 +186,16 @@ public abstract class AbstractTraceForURIProvider<SomeFile, Trace extends Abstra
 	@Override
 	public Trace getTraceToSource(AbsoluteURI absoluteDerivedResource, IProjectConfig project) {
 		final SomeFile generatedFile = asFile(absoluteDerivedResource, project);
-		return getTraceToSource(generatedFile);
+		return doGetTraceToSource(generatedFile);
 	}
 	
 	@Override
 	public Trace getTraceToSource(SourceRelativeURI srcRelativeDerivedResource, IProjectConfig project) {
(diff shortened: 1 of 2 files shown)


szarnekow:
> I guess that since annotations aren't required because they're internal, are they?

Yes, I'd say so.

org.eclipse.xtext/src/org/eclipse/xtext/generator/trace/internal/AbstractTraceForURIProvider.java
```diff
@@ -186,16 +186,16 @@ protected SomeFile asFile(SourceRelativeURI srcRelativeDerivedResource, IProject
...
 	
 	@Override
 	public Trace getTraceToSource(SourceRelativeURI srcRelativeDerivedResource, IProjectConfig project) {
 		final SomeFile generatedFile = asFile(srcRelativeDerivedResource, project);
-		return getTraceToSource(generatedFile);
+		return doGetTraceToSource(generatedFile);
 	}
 	
-	public Trace getTraceToSource(final SomeFile generatedFile) {
+	protected Trace doGetTraceToSource(final SomeFile generatedFile) {
```
cdietrich, line 198: wait this always was meant to be a util and never called from outside.
did not grasp that when staring at the code.

szarnekow: Took me a while to realize that, too
I was considering a public overload in the subclass for IFile, but didn't find a usage

Issue #3558, opened 2025-11-19
org.eclipse.xtext.ide.server.Document goes over the document text twice to compute two offsets for a given lsp4j range


rubenporras:
This can be avoided if when finding the offset for the start of the lsp4j range, we store the data about the document position and reuse it during the search for the second offset.

This should improve performance when the operations are done on big documents.

PR #3559, opened 2025-11-19
Optimize range operations to avoid repeated content traversal

rubenporras:
for #3558

### DIFF
--- a/org.eclipse.xtend.core/src/org/eclipse/xtend/core/parser/FlexTokenRegionProvider.java
+++ b/org.eclipse.xtend.core/src/org/eclipse/xtend/core/parser/FlexTokenRegionProvider.java
@@ -36,7 +36,7 @@ public class FlexTokenRegionProvider extends TokenRegionProvider {
 			int nextStart = flexer.getTokenLength();
 			
 			final int regionStartOffset = region.getOffset();
-			final int regionEnd = regionStartOffset + region.getLength();
+			final int regionEnd = region.getEndOffset();
 			while (token != Token.EOF && nextStart <= regionStartOffset) {
 				prevStart = nextStart;
 				token = flexer.advance();
--- a/org.eclipse.xtend.ide.tests/src/org/eclipse/xtend/ide/tests/refactoring/XtendExpressionUtilTest.xtend
+++ b/org.eclipse.xtend.ide.tests/src/org/eclipse/xtend/ide/tests/refactoring/XtendExpressionUtilTest.xtend
@@ -93,7 +93,7 @@ class XtendExpressionUtilTest extends AbstractXtendTestCase {
 		val selectedExpression = util.findSelectedExpression(expression.eResource as XtextResource, 
 			new TextSelection(selectionOffset, selectionLength))
 		val selectedRegion = locationInFileProvider.getFullTextRegion(selectedExpression)
-		assertEquals(expectedSelection, cleanedModel.substring(selectedRegion.offset, selectedRegion.offset + selectedRegion.length))
+		assertEquals(expectedSelection, cleanedModel.substring(selectedRegion.offset, selectedRegion.endOffset))
 	}
 
 	def protected parse(CharSequence string)  {
--- a/org.eclipse.xtend.ide.tests/xtend-gen/org/eclipse/xtend/ide/tests/refactoring/XtendExpressionUtilTest.java
+++ b/org.eclipse.xtend.ide.tests/xtend-gen/org/eclipse/xtend/ide/tests/refactoring/XtendExpressionUtilTest.java
@@ -103,11 +103,7 @@ public class XtendExpressionUtilTest extends AbstractXtendTestCase {
     TextSelection _textSelection = new TextSelection(selectionOffset, selectionLength);
     final XExpression selectedExpression = this.util.findSelectedExpression(((XtextResource) _eResource), _textSelection);
     final ITextRegion selectedRegion = this.locationInFileProvider.getFullTextRegion(selectedExpression);
-    int _offset = selectedRegion.getOffset();
-    int _offset_1 = selectedRegion.getOffset();
-    int _length = selectedRegion.getLength();
-    int _plus = (_offset_1 + _length);
-    Assert.assertEquals(expectedSelection, cleanedModel.substring(_offset, _plus));
+    Assert.assertEquals(expectedSelection, cleanedModel.substring(selectedRegion.getOffset(), selectedRegion.getEndOffset()));
   }
 
   protected XtendFile parse(final CharSequence string) {
--- a/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/contentassist/XtendProposalProvider.java
+++ b/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/contentassist/XtendProposalProvider.java
@@ -510,11 +510,11 @@ public class XtendProposalProvider extends AbstractXtendProposalProvider {
 		INode node = context.getCurrentNode();
 		ITextRegion textRegion = node.getTextRegion();
 		int offset = textRegion.getOffset();
-		int length = textRegion.getLength();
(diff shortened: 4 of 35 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.java
```diff
@@ -38,16 +39,38 @@ public Document(Integer version, String contents, boolean printSourceOnError) {
 		this.contents = contents;
 		this.printSourceOnError = printSourceOnError;
 	}
+	
+	public record Range(int startOffset, int endOffset) {
```
cdietrich: since?

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.java
```diff
@@ -38,16 +39,38 @@ public Document(Integer version, String contents, boolean printSourceOnError) {
...
+			return endOffset - startOffset;
+		}
+	};
+
+	private record PositionWithDocumentOffset(Position position, int offset) {};
+	
+	/**
+	 * @since 2.41
+	 */
+	public Document.Range getRange(org.eclipse.lsp4j.Range range) {
```
cdietrich: wonder if we could have some tests

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.java
cdietrich: we also already have

- ITextRegion (well offset + length)

but we only partially use it in xtext.ide but e.g. Formatting Service does already use it

rubenporras: done

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.java
rubenporras: done

cdietrich:
@rubenporras can you have a look at the compile errors?

rubenporras:
> @rubenporras can you have a look at the compile errors?

Sorry, I forgot to push some of the files. I have modified many more files in to use the new interface instead of always calling `getOffset() + getLength()` but I think it is better to leave those for a new commit to keep this one small.

If you prefer, I can push that to this PR as a second commit.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/ITextRegion.java
```diff
@@ -17,6 +17,13 @@ public interface ITextRegion {
 
 	int getLength();
 
+	/**
+	 * @since 2.41
+	 */
+	default int getEndOffset() {
```
cdietrich, line 23: i am not sure if i like this @szarnekow wdyt

rubenporras: If it helps for the discussion, I have added a second commit to the PR with the simplification of the several consumers of `ITextRegion`.

szarnekow: I think I do like it. In fact I missed this quite often :)

Maybe add a hint as JavaDoc that the endOffset is exclusive and the offset is inclusive.

cdietrich:
@rubenporras there is still compile errors

rubenporras:
@cdietrich, there is still a failure in the macos build

  RenameStrategyTest.testOverrideIndicatorAnnotationAfterFileRename:102->Assert.assertNotNull:723->Assert.assertNotNull:713->Assert.assertTrue:42->Assert.fail:87

I had a quick look and I do not see a clear relation between that failure an my PR. Do you think it can be a flaky test? If so, could you rerun the build to verify?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/ITextRegion.java
```diff
@@ -17,11 +17,43 @@ public interface ITextRegion {
...
+	};
+
 	ITextRegion merge(ITextRegion region);
 	
 	boolean contains(ITextRegion other);
 	
 	boolean contains(int offset);
 
-	static ITextRegion EMPTY_REGION = ITextRegionWithLineInformation.EMPTY_REGION;
+	static ITextRegion EMPTY_REGION = new ITextRegion() {
```
rubenporras, line 33: @cdietrich , @szarnekow , this is a bit tricky. I need to remove the dependency to ITextRegionWithLineInformation because such a circular dependency does not work with the default method. There is an exception and `EMPTY_REGION` is set to `null`. 
I think the modification is fine, but if you are worried about possible consumers using `==` and not `equals` I would not know how to implement `getEndOffset`, and I would fallback to not do it.

cdietrich:
er aslo have some flaky test am not sure if this is one of them

org.eclipse.xtext.util/src/org/eclipse/xtext/util/ITextRegion.java
cdietrich: what do you mean by circular dependency?

rubenporras: > what do you mean by circular dependency?

the static constant `EMPTY_REGION` of type `ITextRegion ` references another static constant `ITextRegionWithLineInformation.EMPTY_REGION` which has a type that extends `ITextRegion` itself.

It looks like this does not work after adding the method with a default implementation.

rubenporras:
> er aslo have some flaky test am not sure if this is one of them

it looks like it was indeed a flaky one.

org.eclipse.xtext.ide/META-INF/MANIFEST.MF
```diff
@@ -13,7 +13,8 @@ Require-Bundle: org.eclipse.xtext;bundle-version="2.41.0";visibility:=reexport,
  org.eclipse.equinox.common;bundle-version="3.19.0",
  org.eclipse.lsp4j;bundle-version="[0.24.0,0.25.0)";resolution:=optional,
  org.eclipse.lsp4j.jsonrpc;bundle-version="[0.24.0,0.25.0)";resolution:=optional,
- org.eclipse.emf.ecore.change;bundle-version="2.16.0"
+ org.eclipse.emf.ecore.change;bundle-version="2.16.0",
+ org.eclipse.xtext.util
```
rubenporras: @cdietrich , do you want to me to pin the bundle version?

cdietrich: cant look into this today @szarnekow @LorenzoBettini can one of you guys continue reviewing

rubenporras: I think probably you will want the pinning, so I have done it.

rubenporras:
From my point of view, the PR is ready to be merged if you are happy with it.

org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/Document.java
```diff
@@ -38,16 +42,31 @@ public Document(Integer version, String contents, boolean printSourceOnError) {
...
+
+	private PositionWithDocumentOffset getPostionWithDocumentOffset(Position toSearch, PositionWithDocumentOffset start) {
 		int l = contents.length();
 		char NL = '\n';
-		int line = 0;
-		int column = 0;
-		for (int i = 0; i < l; i++) {
+		int line = start.position.getLine();
+		int column = start.position.getCharacter();
+		for (int i = start.offset; i < l; i++) {
```
szarnekow, line 66: I wonder whether we can use the intrinsified `String.indexOf()` and some math to speed this up even further. Replacing our loop here might be beneficial.

Follow-up work, though.

PR #3572, opened 2025-12-02
remove x-internal for TextMate fragment

asheiduk:
The TextMate generator fragment was introduced in 2.35. But to use it without provoking warnings in our code the `x-internal` marking should be removed. No other fragment seems to have this marker.


cdietrich:
am not sure if @szarnekow had some reasoning behind this

cdietrich:
@asheiduk can you please check your ECA status

asheiduk:
@cdietrich Updated

szarnekow:
> am not sure if @szarnekow had some reasoning behind this

The reason was probably that it landed short before a release, and we didn't get a ton of feedback yet so no updates have been made.

If it works reasonably well, I'm absolutely fine with making it "public"

asheiduk:
@szarnekow 
Regarding the output: The resulting file does not work when trying to install it into Eclipse: Comparing with the example from 
https://eclipse.dev/Xtext/documentation/340_lsp_support.html at least the section `fileTypes` is missing.

cdietrich:
The eclipse help needs regular regeneration and I have no time to check where I still have a workspace with a working Jekyll stuff

asheiduk:
@cdietrich : +Confused+ The Help is correct, the text generated by the fragment is missing `fileTypes`. Without that key I failed to install the TextMate grammar into Eclipse. With that key added manually it worked.

cdietrich:
Ahhh am confused too . How do you register it to be eclipse ?

asheiduk:
@cdietrich As explained in https://eclipse.dev/Xtext/documentation/340_lsp_support.html using only these steps:
- Define a content type/extension mapping as as explained in step 1)
- then import the generated/modified Json file as explained in step 4)

Before that I registered the fragment according as explained here: https://github.com/eclipse-xtext/xtext/issues/1329#issuecomment-2518670529

cdietrich:
I wonder when this broke. Need to check the old example if I came find it

cdietrich:
nope here i had it from the beginning:
https://github.com/cdietrich/mydsl-lsp4e/blob/cdb0fe757262d105f05bd7eecdff836adf15deed/mydsl.lsp4e/syntaxes/mydsl.tmLanguage.json#L5

asheiduk:
@cdietrich Regarding CLA: I have updated the CLA, checked there the link to my github account. what is missing?

cdietrich:
i requested a revalidate.
what about the missing feature?

cdietrich:
ping @asheiduk

asheiduk:
Thanks for pinging...

I have tried extending `TextMateGrammar` to support `fileType`. This is not possible because in `TextMateGrammar.init()` a new instance is created and this new instance is used for writing the actual file. There is no way the add an "copy additional data" step. So I wrote a new `TextMateHighlightingFragment2` and `TextMateGrammar2` with a more extensible API. The rest of the existing code can be reused. 

One weakness of the existing code and the modified code is the handling of operators (defined here loosly as keywords without alphanumeric characters). The existing code ignores them. I have tried to add them but dropped the idea because contrary to alphanumeric keywords operators can be adjacent without any space (think of `a*-b`). But without any delimiter there is a conflict between Division (`/`) and Line-Comment `// foobar`. The order of rules matter although this only makes a difference when two rules match the same text. Because the existing code writes the rule with the keywords first, I attempted to add another rule for the operators *roughly* like this after that rule:

```
{
  match: "\+|-|\*|/|\(|\)", // and other operators while having fun quoting them
  name: "keyword.control.mydsl"
}
```
The automatically generated rules from the `Terminals` grammar follow after that.

But then a line comment is interpreted as a "double-division" with trailing garbage. So I've decided to skip the operators -- these are matched now by the rule `ANY_OTHER` in `Terminals` and the TextMate grammar maps this to `invalid.mydsl`. This isn't perfect but looked actepable in IntelliJ Community Edition using the Redhat plugin.

But last week someone used the "Light"-mode of IntelliJ and the `invalid` rule renders very badly: Red background with black text. This reveals that not only the operators are matched by the `invalid` rule but also all whitespaces. This should be a problem with the existing code also! The `WS` rule in `Terminals` seems to be ignored.

**TL;DR:** 
- Simply removing the flag is not an option any more.
- The existing generated TextMate Grammar is not usable:
    - missing `fileType`
    - inappropriate handling of `WS` and `ANY_OTHER`
    - missing operator support

I've intended to upload a CLA-clean redo of the results but I don't acutally see a clean and automatic solution.

cdietrich:
So we can postpone and merge this one anyway ?

asheiduk:
Postpone: yes, merge: not advised

PR #3579, opened 2025-12-22
fix: be more lenient in xtend hover test

cdietrich:
the new jdt has a newline before the close code tag


LorenzoBettini:
I'd agree with making our tests more lenient and less dependent on small details like these.

PR #3600, opened 2026-02-13
Migrate to maven-wrapper source and update to Maven 3.9.12

HannesWell:
Having only source files checked in makes it easier to very the correctness of the content.

Files were generated by running
`mvn wrapper:wrapper -Dtype=script -Dmaven=3.9.12`


LorenzoBettini:
@HannesWell I think you forgot to give the mwnw Unix script the execution bit

LorenzoBettini:
I guess you can also remove the java source since we can assume that curl is installed.

HannesWell:
> @HannesWell I think you forgot to give the mwnw Unix script the execution bit

Indeed. Fixed it (not trivial on Windows 😅)


> I guess you can also remove the java source since we can assume that curl is installed.

Sure, even better. Switched to the `script` type as described in
https://maven.apache.org/tools/wrapper/#Usage_with_or_without_Binary_JAR

LorenzoBettini:
@cdietrich I'm fine with the PR, WDYT?

cdietrich:
i still see red builds

LorenzoBettini:
> i still see red builds

I restarted the failing builds but those are related to the usual flaky tests

PR #3601, opened 2026-02-13
[Build] Leverage select-jdk-toolchain to simplify toolchain spec

HannesWell:
Leverage the new [JDK Toolchain discovery mechanism](https://maven.apache.org/plugins/maven-toolchains-plugin/toolchains/jdk-discovery.html) of the latest `maven-toolchains-plugin` and use it's `select-jdk-toolchain`.

This avoids the need to maintain the `toolchains.xml`.

The new goal selects the JDK toolchain based on a set of constraints and the only one used here is a corresponding [`version`](https://maven.apache.org/plugins/maven-toolchains-plugin/select-jdk-toolchain-mojo.html#version) constraint if one of the `strict-jdk-X` profiles is invoked.
It then searches for example for environment variables named `JAVA_XX_HOME` pointing to JDKs (used in for Jenkins) or the Maven `toolchains.xml` if one is available (used in Github actions and generated by the [`setup-java`](https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#modifying-maven-toolchains) action).
It also searches known locations from popular installers, but this is not used in the Jenkins or GH CI.

To all JDKs found the mentioned constraints are applied and the matching first one is used as tool.

If no constraint is specified, the default case for Xtext, the selection is dormant and just the running JDK is used.

PR #3603, opened 2026-02-16
Use maven.compiler.release instead of source plus target

HannesWell:
The use of the `--release` option of the Java compiler implies a corresponding `--source` and `--target` value and is even more powerful: https://docs.oracle.com/en/java/javase/21/docs/specs/man/javac.html#option-release

### DIFF
--- a/org.eclipse.xtend.maven.archetype/src/main/resources/archetype-resources/pom.xml
+++ b/org.eclipse.xtend.maven.archetype/src/main/resources/archetype-resources/pom.xml
@@ -10,8 +10,7 @@
 	<properties>
 		<xtend.version>${project.version}</xtend.version>
 		<project.build.sourceEncoding>${sourceEncoding}</project.build.sourceEncoding>
-		<maven.compiler.source>${javaVersion}</maven.compiler.source>
-		<maven.compiler.target>${javaVersion}</maven.compiler.target>
+		<maven.compiler.release>${javaVersion}</maven.compiler.release>
 	</properties>
 	<dependencyManagement>
 		<dependencies>
--- a/org.eclipse.xtend.maven.parent/pom.xml
+++ b/org.eclipse.xtend.maven.parent/pom.xml
@@ -22,8 +22,7 @@
 
 		<project.build.sourceEncoding>ISO-8859-1</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>ISO-8859-1</project.reporting.outputEncoding>
-		<maven.compiler.source>21</maven.compiler.source>
-		<maven.compiler.target>21</maven.compiler.target>
+		<maven.compiler.release>21</maven.compiler.release>
 		<maven.javadoc.failOnError>false</maven.javadoc.failOnError>
 		<maven.javadoc.opts></maven.javadoc.opts>
 	</properties>
--- a/org.eclipse.xtend.maven.plugin/src/test/resources/it/compile/simple-java25/pom.xml
+++ b/org.eclipse.xtend.maven.plugin/src/test/resources/it/compile/simple-java25/pom.xml
@@ -10,8 +10,7 @@
 	</parent>
 	<artifactId>simple-java25</artifactId>
 	<properties>
-		<maven.compiler.source>25</maven.compiler.source>
-		<maven.compiler.target>25</maven.compiler.target>
+		<maven.compiler.release>21</maven.compiler.release>
 	</properties>
 	<build>
 		<plugins>
--- a/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/wizard/WizardConfigurationTest.java
+++ b/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/wizard/WizardConfigurationTest.java
@@ -75,7 +75,7 @@ public class WizardConfigurationTest {
 		assertTrue(config.getP2Project().pom().getContent().contains("eclipse-repository"));
 		assertTrue(config.getParentProject().pom().getContent().contains("tycho"));
 	}
-	
+
 	@Test
 	public void p2ProjectsEnablesSourceGenerationWithTychoWhenMavenBuiltIsEnabled() {
 		config.getUiProject().setEnabled(true);
@@ -351,8 +351,7 @@ public class WizardConfigurationTest {
 	@Test
 	public void allBuildSystemsUseJava21() {
 		String parentPom = config.getParentProject().pom().getContent();
-		assertTrue(parentPom.contains("<maven.compiler.source>21</maven.compiler.source>"));
-		assertTrue(parentPom.contains("<maven.compiler.target>21</maven.compiler.target>"));
+		assertTrue(parentPom.contains("<maven.compiler.release>21</maven.compiler.release>"));
 		String parentGradle = config.getParentProject().buildGradle().getContent();
 		assertTrue(parentGradle.contains("sourceCompatibility = JavaVersion.VERSION_21"));
 		assertTrue(parentGradle.contains("targetCompatibility = JavaVersion.VERSION_21"));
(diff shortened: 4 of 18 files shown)


cdietrich:
can you import the reference projects onto xtext dev workspace and execute the launch config to regenerate them.
with that we can verify it works as expected

cdietrich:
https://github.com/xtext/xtext-reference-projects/blob/master/launch/all-refprojects.launch

HannesWell:
> can you import the reference projects onto xtext dev workspace and execute the launch config to regenerate them. with that we can verify it works as expected
> https://github.com/xtext/xtext-reference-projects/blob/master/launch/all-refprojects.launch

Done in
- https://github.com/xtext/xtext-reference-projects/pull/528

cdietrich:
thx. wonder how gradle handles these
https://github.com/xtext/xtext-reference-projects/blob/05e0a200617bba171e3be91dcfe10e3bc20bb1fd/greetings-gradle/2.42.0/org.xtext.example.mydsl.parent/build.gradle#L36

HannesWell:
> wonder how gradle handles these

How to admit that I my gradle knowledge is almost zero (tried to use/configure it once and failed miserably).
But since the Maven options map 1:1 on the java compiler options mentioned above, I assume there are some similar options for gradle?

cdietrich:
how does your schedule next week tuesday at 9am look like?
maybe we should invite you to our Xtext call

HannesWell:
> > wonder how gradle handles these
> 
> How to admit that I my gradle knowledge is almost zero (tried to use/configure it once and failed miserably). But since the Maven options map 1:1 on the java compiler options mentioned above, I assume there are some similar options for gradle?

According to Gemini AI one can just set
```
java {
    release = 17
}
```
Should I apply this too? 

> how does your schedule next week tuesday at 9am look like?

I could be available.

cdietrich:
https://github.com/eclipse-xtext/xtext/actions/runs/22051341768/job/63710002939?pr=3603
there seems to be test fail. not sure if related

cdietrich:
@szarnekow can you add hannes to the meeting

HannesWell:
> https://github.com/eclipse-xtext/xtext/actions/runs/22051341768/job/63710002939?pr=3603 there seems to be test fail. not sure if related

Since it consistently failed over all build I think it is related. The output was
```
[INFO] --- xtext:2.42.0-SNAPSHOT:generate (generate) @ java-lang-bi-ref ---
[INFO] Encoding: UTF-8
[INFO] Compiler source level: 11
[INFO] Compiler target level: 11
[INFO] Collecting source models.
[INFO] Installing type provider.
[INFO] Generating stubs into /home/jenkins/agent/workspace/xtext_PR-3603/org.eclipse.xtext.maven.plugin/target/local-test-dir/it/generate/java-lang-bi-ref/target/xtext-temp/stubs for 1 resources
[INFO] Compiling stubs located in /home/jenkins/agent/workspace/xtext_PR-3603/org.eclipse.xtext.maven.plugin/target/local-test-dir/it/generate/java-lang-bi-ref/target/xtext-temp/stubs
[INFO] Installing type provider for stubs.
[INFO] Validate and generate.
[INFO] Validating: 'XbaseReferToJava.xbase'
[ERROR] ERROR:JavaRecord cannot be resolved to a type. (file:/home/jenkins/agent/workspace/xtext_PR-3603/org.eclipse.xtext.maven.plugin/target/local-test-dir/it/generate/java-lang-bi-ref/src/XbaseReferToJava.xbase line : 2 column : 5)
[ERROR] ERROR:JavaRecord cannot be resolved. (file:/home/jenkins/agent/workspace/xtext_PR-3603/org.eclipse.xtext.maven.plugin/target/local-test-dir/it/generate/java-lang-bi-ref/src/XbaseReferToJava.xbase line : 2 column : 25)
[WARNING] WARNING:The value of the local variable name is not used (file:/home/jenkins/agent/workspace/xtext_PR-3603/org.eclipse.xtext.maven.plugin/target/local-test-dir/it/generate/java-lang-bi-ref/src/XbaseReferToJava.xbase line : 3 column : 5)
```

It might be related to the fact that the default source/target version of the xtext/xtend compiler mojo were still at Java-11.
I moved it to 17, which I think makes sense as 17 is the current baseline, doesn't it?

cdietrich:
yes but i dont remember what @LorenzoBettini did back then when looking into the record topic

LorenzoBettini:
As I said in https://github.com/eclipse-xtext/xtext/pull/3601#discussion_r2811629004, we need .source and .target because Xtend reads them, but not .release, is that right @szarnekow ?

cdietrich:
there is 

org.eclipse.xtend.maven.AbstractXtendCompilerMojo.javaSourceVersion
org.eclipse.xtext.maven.AbstractXtextGeneratorMojo.compilerSourceLevel

that i could find

HannesWell:
> As I said in [#3601 (comment)](https://github.com/eclipse-xtext/xtext/pull/3601#discussion_r2811629004), we need .source and .target because Xtend reads them, but not .release

Indeed, yes that makes sense and explains the error. Thanks! Before this change the mojos read the source/target values from the corresponding properties and now fall back to their default, which is/was Java-11, which doesn't know records.

But I wonder if there is a specific reason to not read `maven.compiler.release` too and derive the source/target values from it? I can extend this PR and implement it.
For me the only question would be which value takes precedence if all three are set?
As far as I understand the Java compiler doc, release takes precedence over source/target and since the maven-compiler-plugin seems to just translate the parameters to the corresponding compiler flags, I assume it behaves the same.

Then changing the default of source/target/release could be applied in a follow-up (if you want that at all).

LorenzoBettini:
> > As I said in [#3601 (comment)](https://github.com/eclipse-xtext/xtext/pull/3601#discussion_r2811629004), we need .source and .target because Xtend reads them, but not .release
> 
> Indeed, yes that makes sense and explains the error. Thanks! Before this change the mojos read the source/target values from the corresponding properties and now fall back to their default, which is/was Java-11, which doesn't know records.
> 
> But I wonder if there is a specific reason to not read `maven.compiler.release` too and derive the source/target values from it? I can extend this PR and implement it. For me the only question would be which value takes precedence if all three are set? As far as I understand the Java compiler doc, release takes precedence over source/target and since the maven-compiler-plugin seems to just translate the parameters to the corresponding compiler flags, I assume it behaves the same.
> 
> Then changing the default of source/target/release could be applied in a follow-up (if you want that at all).

Probably updating the maven plugins to also read that property would be the right way to go.
I'd like to hear from @szarnekow in that respect.

HannesWell:
> Probably updating the maven plugins to also read that property would be the right way to go.
> I'd like to hear from @szarnekow in that respect.

Great. I already took action and we can discuss details at
- https://github.com/eclipse-xtext/xtext/pull/3607

Since this effectively depends on the latter and it seems Xtext doesn't use SNAPSHOTs of its own Maven plugins it would be good of that could land in the upcoming release

HannesWell:
> It might be related to the fact that the default source/target version of the xtext/xtend compiler mojo were still at Java-11.
> I moved it to 17, which I think makes sense as 17 is the current baseline, doesn't it?

This part is now removed/reverted, since the mojos are now capable of handling `maven.compiler.release`:
- https://github.com/eclipse-xtext/xtext/pull/3607

Lets see if that's already sufficient to make all tests pass or if we (temporarily) have to bootstrap against the nightly build until M1.

cdietrich:
i can build a m0, but do we want to include more stuff

HannesWell:
> i can build a m0, but do we want to include more stuff

That would be the most save way.
The build already seem to pass without the new support used all over the build, but it could be that some implicit changes happen because now the default source/target version is used at some places atm. and would cause undesired differences (that don't break the build).

I cannot tell if something else should be included into M0. From my side there are no further 'urgent' changes.

HannesWell:
Update this to the latest main and resolved all conflicts.
Due to https://github.com/eclipse-xtext/xtext/pull/3655, this will probably already work although M1 with https://github.com/eclipse-xtext/xtext/pull/3607 is not yet available used.
The target, compliance value is atm. not set, but the default has the desired value.

So the result should probably be the same, but I think it would also make sense to await M1.

PR #3604, opened 2026-02-16
Add compiler support for Java-25 and set up a Java-25 build

HannesWell:
This adds only compiler support for Java-25, i.e. it becomes possible to compile projects that target Java-25.
This doesn't add UI support for it yet, but I can try to extend this PR if you want to have this one pass.
A dedicated Java-25 test project is also missing (like the one added for Java-21), but I can also look into that.
As far as I know now new relevant compiler constructs where added respectivly the Java standard library wasn't changed between Java-21 and 25 in a way that affects Xtext.

Java-17 support is maintained with this.

Part of
- https://github.com/eclipse-xtext/xtext/issues/3532

This also includes
- https://github.com/eclipse-xtext/xtext/pull/3601

I could do it without it, the latter but it would become more complex eventually.


.github/workflows/maven.yml
```diff
@@ -14,14 +14,18 @@ jobs:
     strategy:
       matrix: # 'macos-14' is arm64
         os: ['ubuntu-22.04', 'macos-14', 'windows-2022']
-        java: [ 21 ]
+        java: [ 25 ]
```
HannesWell, line 17: In case just minimal support is added for Java-25 at the moment the main build could also stay at Java-21 and just one extra branch for Java-25 could be added.

cdietrich:
what is still unclear to me: how does this work without changes to bom

org.eclipse.xtext.xbase.ui.tests/src/org/eclipse/xtext/xbase/ui/tests/JavaVersionExtendedTest.java
```diff
@@ -26,10 +26,11 @@ public void testToJdtClassFileConstant() {
...
 		assertEquals(ClassFileConstants.JDK11, JavaVersion.JAVA11.toJdtClassFileConstant());
 		assertEquals(ClassFileConstants.JDK17, JavaVersion.JAVA17.toJdtClassFileConstant());
-		// until we use the newer JDT as minimal requirement, we cannot refer to ClassFileConstants.JDK21
+		 assertEquals(ClassFileConstants.JDK21, JavaVersion.JAVA21.toJdtClassFileConstant());
+		// until we use the newer JDT as minimal requirement, we cannot refer to ClassFileConstants.JDK25
 		// or the test won't compile with older target platforms
-		// assertEquals(ClassFileConstants.JDK21, JavaVersion.JAVA21.toJdtClassFileConstant());
-		assertEquals(4259840, JavaVersion.JAVA21.toJdtClassFileConstant());
+		// assertEquals(ClassFileConstants.JDK25, JavaVersion.JAVA25.toJdtClassFileConstant());
+		assertEquals(4521984, JavaVersion.JAVA25.toJdtClassFileConstant());
```
cdietrich, line 33: i assume we need to use the try catch approach here

cdietrich:
as i have feared this will open the complete can
grafik

HannesWell:
> as i have feared this will open the complete can

Indeed.
So I assume you prefer to postpone this too?
For me it wouldn't be that bad, since I found in the meantime, that I can just leave the Xtext/Xtend compiler configuration at Java-21 while moving the runtime and other parts to Java-25 seems to work well. So for me this isn't 'urgent' anymore.
I assume that lower level adaption isn't necessary anyways in Xtext.

cdietrich:
i also assume it wont work at runtime due to bom
even if we would hide it from the wizard

HannesWell:
What bom are you referring to exactly?
The targeted eclipse version must of course support Java-25. But just running with Java-25 could work with quite old releases. Of course if one compiles code at runtime targeting Java-25 one must use Eclipse 2025-12, which is the first version of JDT fully supporting Java-25.

cdietrich:
stuff like the xtext maven plugin uses the bom i have linked above / in the other discussionhttps://github.com/eclipse-xtext/xtext/blob/425a9bee287588415dacc1042ce9e31f4ac7a3fa/org.eclipse.xtext.dev-bom/pom.xml#L273

HannesWell:
> what is still unclear to me: how does this work without changes to bom

In the meantime I added a java-25 profile to the BOM. But I'm not sure if it is actually necessary, because the `jdt.core` version supporting Java-25 still runs on Java-17. So we could just bump that. But I don't know your policy with respect to the versions in the BOM and the minimally supported target SimRel.

Furthermore I added a dedicated compiler and wizard tests for Java-25 and adapted failing tests.
The most difficult one is probably `ImportOrganizerTest.xtend`, which had to be adapted because `java.util.Collections` class became final in (or before) Java-25.

In the current state all Github and Jenkins builds (executed in the m2e JIPP) succeeded in my testing.

Please let me know if something is missing.

org.eclipse.xtext.xbase.ui.tests/src/org/eclipse/xtext/xbase/ui/tests/JavaVersionExtendedTest.java
HannesWell: Why? I don't see anything referenced that's not available in the baseline Java or Eclipse version?

cdietrich: also in the 2024-03 target platform / the one that is also used in maven bom?

cdietrich: in the past we did some reflection in that test instead of 
4521984

HannesWell: > also in the 2024-03 target platform / the one that is also used in maven bom?

It should be yes. Eclipse 2025-12 had complete Java-25 support and since Java-21 was released two years earlier I expect the 21 constants to be available since 2023-12.
And I also ran the Jenkins build in the m2e JIPP with the lowest TP and java-17 and it succeeded.

> in the past we did some reflection in that test instead of

That would work too, but I think the approach to just inline the value (implemented before by @LorenzoBettini) is simpler.
And the later expected code is already available as commented out code.

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/imports/ImportOrganizerTest.xtend
```diff
@@ -2020,21 +2020,21 @@ class ImportOrganizerTest extends AbstractXtendTestCase {
 	// see https://bugs.eclipse.org/bugs/show_bug.cgi?id=399519
 	@Test def testExtensionFeatureCallToStaticLocalFieldFromSuper() {
 		'''
-			import java.util.Collections
```
szarnekow, line 2023: Why did you change this test?

HannesWell: The `Collections` is final in Java-25 and wasn't in Java-21 and before:
- https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/Collections.html
- https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Collections.html

So I searched for another class that's not final and has (directly or inherited) static methods that could be used for this test.
And `ArrayList` seemed a simple replacement where the static methods from the `List` interface could be used (and it was similar).

cdietrich:
@HannesWell can you check what of this is missing on https://github.com/eclipse-xtext/xtext/pull/3638

cdietrich:
So your proposal is to close eyes on eg wizard not working properly on main

org.eclipse.xtext.xbase.ui.tests/src/org/eclipse/xtext/xbase/ui/tests/JavaVersionExtendedTest.java
LorenzoBettini: Yes, IIRC, in the past I had inlined a constant for such a scenario. I did that because someone else had done that in the past.

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/imports/ImportOrganizerTest.xtend
LorenzoBettini: In the past, when supporting Java 21, I had similar problems due to changes in the JDK. I solved them by NOT using JDK types for tests: I crafted a few Java types (stub classes) with the requirements for our tests.

If you need a "class that's not final and has (directly or inherited) static methods" I'd create a stub class in that tests project and use that one in the test.

HannesWell:
> So your proposal is to close eyes on eg wizard not working properly on main

Sorry for the simple question, but what wizard changes are you referring to exactly?
I went through your big PR https://github.com/eclipse-xtext/xtext/pull/3638 and didn't find a change related to the wizard that seems relevant for me for just the addition of Java-25 support. Bumping the minimum to 21, of course requires changes again, but IMO that's better done separately.
Or do you also have not yet applied the required wizard changes for adding Java-25?
Because in https://github.com/eclipse-xtext/xtext/pull/3604#issuecomment-3907078524, we see that the wizard already reads the available constants.
And I've already added the wizard tests for Java-25 and they all passed the last time I checked it.

But I'm happy to add the required wizard changes if you point me to them.

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/imports/ImportOrganizerTest.xtend
HannesWell: That's a good idea.
But since it's working now, I would be fine to leave it as it is for now. However, if you'd prefer to have it changed, I can implement your suggestion.

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaRuntimeVersion.java
```diff
@@ -58,6 +58,10 @@ public static boolean isJava21OrLater() {
 		return JAVA_VERSION >= 21;
 	}
 
+	public static boolean isJava25OrLater() {
+		return JAVA_VERSION >= 21;
```
cdietrich: typo

pom.xml
```diff
@@ -715,6 +715,19 @@
...
+			<id>strict-jdk-25</id>
+			<properties>
+				<toolchain.jdk.version>[25,26)</toolchain.jdk.version>
+				<!--
+					Since the Xtend compiler generates Java code using new API introduced in Java 21,
+					so we must increase the release level or the Java compilation would not find the
+					new API when it compiles the Java code generated by Xtend.
+					see https://github.com/eclipse-xtext/xtext/issues/2686#issuecomment-2038175179
+				-->
+				<maven.compiler.release>21</maven.compiler.release>
```
cdietrich, line 728: intentionally 21?

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaRuntimeVersion.java
HannesWell: Indeed, thanks! Fixed it.

pom.xml
HannesWell: Yes, because the java code of Xtext jar's itself still targets 17 at the moment and needs newer Java API are used when compiling against 21 or higher. This was the SequencedCollection issue.
This can be removed as soon as the baseline is 21 and 17 is dropped.

HannesWell:
The Jenkins changes are again tested with
https://ci.eclipse.org/m2e/job/try-outs-xtext-Hannes/

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/imports/ImportOrganizerTest.xtend
LorenzoBettini: It depends on how much work it requires.
If it's not that much, I'd do it now so we can continue removing our dependencies on JDK, which will likely break in future JDK versions.
If it takes too much time, now that it works we can leave it like that; in the future it might break again ;)

org.eclipse.xtext.xbase.ui.tests/src/org/eclipse/xtext/xbase/ui/tests/JavaVersionExtendedTest.java
cdietrich: can clean this up with my big bump pr in follow up

org.eclipse.xtend.core.tests/src/org/eclipse/xtend/core/tests/imports/ImportOrganizerTest.xtend
cdietrich: we have this problem on a regular base so i am fine with adjusting stuff while keeping the intent of the test

cdietrich:
can you check the failing CliWizardIntegrationTest

HannesWell:
> can you check the failing CliWizardIntegrationTest

It was referencing a milestone MWE repository that was probably set during the release rampdown and that I copied at the time when I created this PR. That should be fixed now.

But this is now blocked by
- https://github.com/eclipse-xtext/xtext/pull/3637#issuecomment-4040132394

cdietrich:
Done

org.eclipse.xtext.util/src/org/eclipse/xtext/util/JavaVersion.java
```diff
@@ -75,7 +75,11 @@ public enum JavaVersion {
 	/**
 	 * Java 21, next LTS version
 	 */
-	JAVA21("Java 21", new String[] {"21", "1.21"}, "JavaSE-21", "-1.21", Constants.JAVA21)
+	JAVA21("Java 21", new String[] {"21", "1.21"}, "JavaSE-21", "-1.21", Constants.JAVA21),
+	/**
+	 * Java 25, next LTS version
```
szarnekow: `next LTS version` is a bit misleading as a comment. Let's just remove that part from the comments on 17, 21, and 25

HannesWell: Makes sense. Done.

HannesWell:
Now all tests pass again and the Jenkins build looks also good:
https://ci.eclipse.org/m2e/job/try-outs-xtext-Hannes/

Do you have more more remarks or need more changes or is this ready?

PR #3606, opened 2026-02-16
Remove M2E builder from all PDE Plug-in/Feature and POM projects

HannesWell:
For standard Plug-in and Feature projects, i.e those are only built by PDE in the IDE or for Maven projects with packaging type POM, the M2E build does not produce any relevant results and sometimes even negatively interferes with the project configuration (i.e. when the 'Maven update' is applied to a project.
To avoid any such negative interference and in order to accelerate the workspace build, this removes the M2E builder where it's not needed.

### DIFF
--- a/.project
+++ b/.project
@@ -5,14 +5,8 @@
 	<projects>
 	</projects>
 	<buildSpec>
-		<buildCommand>
-			<name>org.eclipse.m2e.core.maven2Builder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
 	</buildSpec>
 	<natures>
-		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
 	<filteredResources>
 		<filter>
--- a/org.eclipse.emf.mwe2.language.sdk.dummy/.project
+++ b/org.eclipse.emf.mwe2.language.sdk.dummy/.project
@@ -10,14 +10,8 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
-		<buildCommand>
-			<name>org.eclipse.m2e.core.maven2Builder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
 	</buildSpec>
 	<natures>
-		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 		<nature>org.eclipse.pde.FeatureNature</nature>
 	</natures>
 </projectDescription>
--- a/org.eclipse.xtend.core.tests/.project
+++ b/org.eclipse.xtend.core.tests/.project
@@ -25,14 +25,8 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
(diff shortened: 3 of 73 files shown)


cdietrich:
do know if maven/tycho/m2e ?!? will readd these on a maven update project?

HannesWell:
> do know if maven/tycho/m2e ?!? will readd these on a maven update project?

It shouldn't since these projects are not Maven projects anymore with this change applied. They are just PDE Plug-in or Feature (or plain Eclipse) projects that als happen to contain a `pom.xml` file and that `pom.xml` is not further processed.


They are only re-added if one configures the project as Maven project again (or if one edits the `.project` file):
grafik

LorenzoBettini:
Even for packaging pom projects? IIRC, without that nature you don't have the context menu for running a maven build.

If a maven project depends on the "maven" part of a plugin project, does this change still allow for dependency resolution?

HannesWell:
> Even for packaging pom projects? IIRC, without that nature you don't have the context menu for running a maven build.

That's right that doesn't work then. But keeping it at pom-projects is also much less harm.

> If a maven project depends on the "maven" part of a plugin project, does this change still allow for dependency resolution?

Again, correct. I didn't thought about the need of the 'pure' Maven projects to require Xtext's Plugin-in projects directly.
That would also not work, I did full test where I deleted all Xtext projects from the workspace, cleaned the repo and imported them again and ended up with resolution errors.

In the meantime I also discovered that I had M2E's automatic project update enabled explicitly and overwrote the configuration from the Xtext Oomph setup (must have set that long time ago).
With that disabled the constantly changing configuration stopped and I don't have that trouble anymore.

And given that this causes other problems, I think there is no point in continuing here.

LorenzoBettini:
@HannesWell In this case, I don't like being correct: I wished these changes could have helped because the m2e update behavior is really annoying ;)

HannesWell:
> In this case, I don't like being correct: 

It can be a burden 😅

> I wished these changes could have helped because the m2e update behavior is really annoying ;)

Absolutely (at least in this case)! But I got silence in this regard by making sure the M2E's Maven setting `Automatically update maven projects configuration` is disabled (maybe also the Update on startup):

grafik

In fact the Xtext Oomph setup already disables that option explicitly:
https://github.com/eclipse-xtext/xtext/blob/99a5c2260930ddb7e1b6c4b8622d709da7adbd3d/releng/org.eclipse.xtext.contributor/Xtext.setup#L248-L251

But I had overwritten in my Oomph User setup (don't know anymore why). So I'd recommend to disable it in the preferences and check if the Oomph setup re-actives it again (by running the setup). And if yes, you should check your `User`, `Workspace` and `Installation` setup and remove it there.

LorenzoBettini:
I'm sure that in the Oomph setup the property is indeed disabled.

PR #3607, opened 2026-02-16
Support maven.compiler.release parameter in Xtext/Xtend Maven plugins

HannesWell:
Add support for reading the maven.compiler.release parameter and grant it precedence over a `maven.compiler.source` or `maven.compiler.target` configuration (similar to the Java compiler).

As suggested in
- https://github.com/eclipse-xtext/xtext/pull/3603#issuecomment-3908158432

### DIFF
--- a/org.eclipse.xtend.maven.plugin/src/main/java/org/eclipse/xtend/maven/AbstractXtendCompilerMojo.java
+++ b/org.eclipse.xtend.maven.plugin/src/main/java/org/eclipse/xtend/maven/AbstractXtendCompilerMojo.java
@@ -45,10 +45,23 @@ public abstract class AbstractXtendCompilerMojo extends AbstractXtendMojo {
 	/**
 	 * Create Java Source Code that is compatible to this Java version.
 	 * 
-	 * Supported values: 11, 17 and so forth
+	 * Supported values: 11, 17, 21 and so forth
 	 */
 	@Parameter(property="maven.compiler.source", defaultValue="11")
 	private String javaSourceVersion;
+
+	/**
+	 * Create Java Source Code that targets this Java release.
+	 * <p>
+	 * If set, this implies a {@link #javaSourceVersion} of the same value and any value explicitly configured for the latter is ignored.
+	 * Nevertheless specifying this parameter does not enforce strict Java API checks like the {@code release} option for a Java compiler
+	 * does. Still it is possible to enforce strict Java API checks by specifying the {@code release} option for the compiler-plugin.
+	 * </p>
+	 * 
+	 * Supported values: 11, 17, 21 and so forth
+	 */
+	@Parameter(property = "maven.compiler.release")
+	private String javaReleaseVersion;
 
 	/**
 	 * The current build session instance. This is used for toolchain manager API calls.
@@ -132,8 +145,9 @@ public abstract class AbstractXtendCompilerMojo extends AbstractXtendMojo {
 			return;
 		}
 		String baseDir = project.getBasedir().getAbsolutePath();
-		log.debug("Set Java Compliance Level: " + javaSourceVersion);
-		compiler.setJavaSourceVersion(javaSourceVersion);
+		String sourceVersion = javaReleaseVersion != null && !javaReleaseVersion.isBlank() ? javaReleaseVersion : javaSourceVersion;
+		log.debug("Set Java Compliance Level: " + sourceVersion);
+		compiler.setJavaSourceVersion(sourceVersion);
 		log.debug("Set generateSyntheticSuppressWarnings: " + generateSyntheticSuppressWarnings);
 		compiler.setGenerateSyntheticSuppressWarnings(generateSyntheticSuppressWarnings);
 		log.debug("Set useXbaseGenerated: " + useXbaseGenerated);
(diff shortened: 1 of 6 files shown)


org.eclipse.xtend.maven.plugin/src/test/resources/it/compile/simple-java21/pom.xml
```diff
@@ -10,8 +10,7 @@
 	</parent>
 	<artifactId>simple-java21</artifactId>
 	<properties>
-		<maven.compiler.source>21</maven.compiler.source>
-		<maven.compiler.target>21</maven.compiler.target>
+		<maven.compiler.release>21</maven.compiler.release>
```
HannesWell, line 13: I only adapted the test-cases in this directory because I expect them to be the only ones to 'use' the changed code immediately.

HannesWell:
@cdietrich, @LorenzoBettini, @szarnekow any chance this will land in the upcoming release?

LorenzoBettini:
As for me, the PR is fine!

I think there's an important aspect to clear: the Xtend compiler and also the Xtext maven plugin infrastructure do not really honor the semantics of "release". That is, they don't check that in the Xtend code or in an Xbase DSL code you only use Java API up to the specified release. Am I right, @szarnekow ?

If that's the case, it should be clearly documented in the Javadoc and possibly in the release notes and documentation.

HannesWell:
> I think there's an important aspect to clear: the Xtend compiler and also the Xtext maven plugin infrastructure do not really honor the semantics of "release". That is, they don't check that in the Xtend code or in an Xbase DSL code you only use Java API up to the specified release.

That's my understanding too. At least at the moment it's implemented as basically just an alias for the source respectively  source+target option in the xtend-compiler and xtext-generator mojos.
However as that property also configures the java compiler such issues should be catched when the Java-compiler is invoked.

On the other hand, since the Xtext/Xbase compiler receives type information from JDT and if the JDT configuration of the project is set accordingly, Xtext should just inherit that capability. So I naively assume that it's just a matter of configuring JDT accordingly.

LorenzoBettini:
For me it's fine.
@szarnekow what do you think?

org.eclipse.xtend.maven.plugin/src/main/java/org/eclipse/xtend/maven/AbstractXtendCompilerMojo.java
```diff
@@ -45,11 +45,24 @@ public boolean apply(String filePath) {
...
+	 * Create Java Source Code that targets this Java release.
+	 * <p>
+	 * If set, this implies a {@link #javaSourceVersion} of the same value and any value explicitly configured for the latter is ignored.
+	 * Nevertheless specifying this parameter does not enforce strict Java API checks like the {@code release} option for a Java compiler
+	 * does. Still it is possible to enforce strict Java API checks by specifying the {@code release} option for the compiler-plugin.
+	 * </p>
+	 * 
+	 * Supported values: 11, 17, 21 and so forth
+	 */
+	@Parameter(property = "maven.compiler.release", defaultValue = "11")
```
cdietrich: what about /org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend

cdietrich: should we switch to new style there?

LorenzoBettini: `release` has been available since Java 9, so it might be the case to use the new style

HannesWell: That's also part of https://github.com/eclipse-xtext/xtext/pull/3603.
And since, as far as I understand, the Xtext build requires some bootstrapping and is not built with the Xtext Maven plugins of the same state, I suggest to keep it separated and implement that enhancement in https://github.com/eclipse-xtext/xtext/pull/3603 as follow-up on this one.

HannesWell:
@cdietrich and @LorenzoBettini (and maybe @szarnekow),
is there anything missing or can this be submitted in time for the release?

cdietrich:
i dont have time to think how we could verify this with
- xtext-reference-projects using the maven repo by this build
- figure out (maybe with logs)
that java 11 fallback is not used 
https://ci.eclipse.org/xtext/view/PRs/job/xtext/job/PR-3607/3/artifact/build/maven-repository/

HannesWell:
> * figure out (maybe with logs)
>       that java 11 fallback is not used

That's a very good point. Probably it does indeed use the Java-11 default as the parameter should actually never be null.
I see three options here:
1. Not set a default for the release parameter
2. Consider the release parameter not set if the value is equal to the default
It then will use the source/target parameter as before.
3. Apply the default in code to be able to detect if the value was actually set or not

The first and last option basically have the same out-come: If not set the target/source attributes are used and since their default is 11 too, that default is used again.
So having no default for this alias (what it effectively is), seems the right choice to me.
Will update this PR accordingly.

HannesWell:
Now that Xtext 2.42 is released, I think it would be a good time to submit this.
Do you think this is ready or what would be missing?

cdietrich:
please be aware of https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/7208

LorenzoBettini:
OK, so we have to wait for the signing issue to be fixed and then we can merge.

cdietrich:
looking into the wizard as follow up would be nice. thx @HannesWell

HannesWell:
Thanks for your review, the discussion and submitting this.

> looking into the wizard as follow up would be nice.

Isn't this already included in
- https://github.com/eclipse-xtext/xtext/pull/3603

PR #3611, opened 2026-02-17
Activate Annotation of synthetic members in Xtext's Oomph setup

HannesWell:
Explicitly activate the Xtend compiler preference `Annotate synthetic members with @XbaseGenerated`.
The generated code committed to the repository uses that and if one has disabled once it will generate diffs in the Git worktree.
So the Oomph setup should explicitly activate that Option.

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -342,6 +342,14 @@
           key="/instance/org.eclipse.xtext.builder/schedulingrule"
           value="ALL_XTEXT_PROJECTS"/>
     </setupTask>
+    <setupTask
+        xsi:type="setup:CompoundTask"
+        name="org.eclipse.xtend.core.Xtend">
+      <setupTask
+          xsi:type="setup:PreferenceTask"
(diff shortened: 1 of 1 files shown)


HannesWell:
@cdietrich, @LorenzoBettini is this good to go?

LorenzoBettini:
I'm fine with that, though I haven't manually tested, it's good to ensure it's set. If @cdietrich is ok, I'd merge.

cdietrich:
i cant judge sry. no mental capaity

szarnekow:
Hmm curious. I expected all projects to have project specific settings.

LorenzoBettini:
> Hmm curious. I expected all projects to have project specific settings.

I think so, but I added that option recently and I guess project settings haven't been updated. By default that setting is on.

Maybe we should update all settings accordingly.

HannesWell:
> By default that setting is on.

It is one by default but disabled it in the workspace because when introduced initially I had problems with it for a short time:
- https://github.com/eclipse-xtext/xtext/issues/3182#issuecomment-2323913121

Then I forgot about it and later had differing code. Therefore I thought that nowadays it would be good to ensure its always active.

> Maybe we should update all settings accordingly.

When that is done, this could be reverted, yes.

PR #3612, opened 2026-02-17
Update generated source code

HannesWell:
When running a full clean-build these source files are changed in this way and currently lead to diffs in the Git worktree.


LorenzoBettini:
@HannesWell I know that's annoying, but we have to live with that until we fix a problem in the Xtend compiler (https://github.com/eclipse-xtext/xtext/pull/2995).

For the time being, I'd close that.

HannesWell:
> For the time being, I'd close that.

OK, too bad. I have to admit I didn't fully understood the problem, but I also didn't want to spend too much time on it. :)

LorenzoBettini:
I wasn't able to solve it myself, let's hope @szarnekow will solve that :)

PR #3614, opened 2026-02-18
Add adapter factory for XtextDocument to support URI, IFile, and IPath

raghucssit:
conversions.

Clients like lsp4e and github copilot tries to derive URI and IFile from IDocument. XtextDocument does not implement IAdaptable so direct adaption is not possible. So we introduce adapter factory.

see https://github.com/eclipse-xtext/xtext/issues/3613

### DIFF
--- a/org.eclipse.xtext.ui/plugin.xml
+++ b/org.eclipse.xtext.ui/plugin.xml
@@ -1176,4 +1176,20 @@
          </themeid>	
 	  </stylesheet>	
    </extension>
+ <extension
+       point="org.eclipse.core.runtime.adapters">
+    <factory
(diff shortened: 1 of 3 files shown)


raghucssit:
@iloveeclipse FYI

iloveeclipse:
See discussion on https://github.com/eclipse-xtext/xtext/pull/3615.

cdietrich:
what is the state here? is the additional or as a replacement to the other pr that was merged

iloveeclipse:
This can be closed, https://github.com/eclipse-xtext/xtext/pull/3615 is better.

PR #3638, opened 2026-03-09
Update to Java 25, Newer tps, etc

cdietrich:


cdietrich:
```
 07:00:09  [ERROR] Failed to execute goal org.apache.maven.plugins:maven-invoker-plugin:3.5.0:install (install dependencies in local test repo) on project xtend-archetype: The following artifacts could not be resolved: p2.eclipse.plugin:de.itemis.xtext.antlr:pom:2.1.1.v201405091103 (absent): Could not find artifact p2.eclipse.plugin:de.itemis.xtext.antlr:pom:2.1.1.v201405091103 in lsp4j-from-jenkins 
```

@HannesWell @LorenzoBettini @szarnekow does this ring a bell, e.g with respect to tycho 5?

cdietrich:
attempt: https://github.com/eclipse-xtext/xtext/commit/0d655235693d96854662c2ddc89dc31e6a1d60b5

HannesWell:
> ```
> ent): Could not find artifact p2.eclipse.plugin:de.itemis.xtext.antlr:pom:2.1.1.v201405091103 in lsp4j-from-jenkins 
> ```
> @HannesWell @LorenzoBettini @szarnekow does this ring a bell, e.g with respect to tycho 5?

This seems to be caused by Tycho 5.
I noticed this also in https://github.com/eclipse-xtext/xtext/pull/3602.
But have not yet found a cause. But can investigate.

Regarding this change, for Java-25 I've already created a PR a while ago:

And from my POV would be ready for some time.
I just assumed you would first want to do the unification with respect to using maven.compiler.release, but for that I'm awaiting an M0 deployment:
- https://github.com/eclipse-xtext/xtext/pull/3603#issuecomment-3989327770

This would unlock multiple PRs from my side, which I consider ready besides that.

cdietrich:
@HannesWell meanwhile i have a ton of stuff on this branch. the question is rather what i overlooked
- min java version 21 (is there 17s i did not see)
- min tp to 2025-12 (bom, tp, manifests)
- constant for java 25 + test
- wizard stuff
- tycho

cdietrich:
see also https://github.com/xtext/xtext-reference-projects/pull/534

HannesWell:
> meanwhile i have a ton of stuff on this branch. the question is rather what i overlooked

Yes, but why can't we have the all these many steps separated into smaller changes?
It makes it simpler to follow in the history later. Plus I find it frustrating to have spent a day or two to make the PR ready to just have it discarded later.

> - constant for java 25 + test

This is fully self-contaiend in https://github.com/eclipse-xtext/xtext/pull/3604, which has (almost) zero prerequisites.

cdietrich:
It how if we need to do tycho and platform and rm java 17 and bin java 21 all in one

cdietrich:
So what is your proposal for a split

LorenzoBettini:
> It how if we need to do tycho and platform and rm java 17 and bin java 21 all in one

@cdietrich I'm not sure I understand what you meant.

I'd tend to agree with @HannesWell concerning the small steps, in particular the PRs he's already worked on.

I had already approved the one about the release option in the maven plugins.

Unfortunately, now we have to wait for the eclipse download problems are fixed before going on.

HannesWell:
> It how if we need to do tycho and platform and rm java 17 and bin java 21 all in one

I also believe it would be possible to split it up.
Adding Java-25 seems possible without significant hurdles and Tycho-5 also almost worked separately and I hope that your fix from this makes it succeed completely. And I'm convinced this can all be done separately.
Probably you could even split the drop of older TPs and requiring Java-21/dropping Java-17 since even the oldest TP should support Java-21 released autumn 2023.


> So what is your proposal for a split

My proposal is
1. https://github.com/eclipse-xtext/xtext/pull/3601
2. https://github.com/eclipse-xtext/xtext/pull/3604 (ideally after rebase)

- https://github.com/eclipse-xtext/xtext/pull/3602  could in general be submitted as soon as the build passes.
But then there is the problem with users need toolchains: https://github.com/eclipse-xtext/xtext/pull/3602#issuecomment-3906248449
So maybe this should await the drop of Java-17 respectively Java-21 as minium.
Or we explain to users, that even if their plugins target Java-17 (or generally a Java version below Tycho's runtime requirement) that they have to use a newer JDK for builds.
Does Xtext also generate a Maven build launch configuration? There the requirement of Tycho could be implemented (plus a warning).
We could also leverage the simpler JDK discovery mechanism of the toolchains plugin (https://maven.apache.org/plugins/maven-toolchains-plugin/toolchains/jdk-discovery.html), but this would also require a newer JDK and usually the `maven.compiler.release` option is sufficient. Toolchains are only really interesting for tests nowadays.

- Dropping Java-17 support/requiring Java-21 should be possible at anytime and generally should not have prerequisites.
- Dropping older TPs should be able at any time.
This can lead to the situation where even the oldest targeted Eclipse already requires Java-21 and Xtext still only requires 17 (if not dropped before). That wouldn't be a fundamental problem, but it would make it impossible to run tests against Java-17.
Therefore I think it would be cleaner to do this after dropping Java-17.

- https://github.com/eclipse-xtext/xtext/pull/3603 can wait until M1. Having it now would only save a two lines workaround in https://github.com/eclipse-xtext/xtext/pull/3601.

cdietrich:
yes but
- real / full compiler support for 25 wont work without bom update and if i understand correct tycho 5
- bom update (and dropping old tp) require java 17 drop
- etc

and if we do it separately finding all places will be terribly hard.
if @LorenzoBettini or @szarnekow are jumping in corrdinating i am fine
but my availablility will reduce more 2nd half of march and april

HannesWell:
> * real / full compiler support for 25 wont work without bom update and if i understand correct tycho 5

In https://github.com/eclipse-xtext/xtext/pull/3604 I have updated the bom and added a profile for Java-25, like it was done for Java-21.
And it builds with Java-25 successfully. It hast tests for the wizard with java-25 that pass, I really don't understand what's missing for full/real support: https://github.com/eclipse-xtext/xtext/pull/3604#issuecomment-4035467603

As far as I can tell, at its very bottom, the core of Java-25 support is just adding the constant. But since Xtext is not using Java-25 language features, it basically just to prevent the exception due to failed constants. I'm about to move our product to Java-25 based on todays SimRel and that works fine, when I set/keep the Xtext java-version to 21 to prevent said exception. So at the moment all one needs is JDT supporting a new java version, the availability of the Xtext java version constant plus wizard support in case new projects are created.

full-build.sh
```diff
@@ -47,6 +47,15 @@ MVN_ARGS+=(-PuseJenkinsSnapshots)
 
 MVN_ARGS+=("-Dmaven.home=$(./mvnw --version | grep "Maven home:" | cut -c 13-)")
 
+echo ./mvnw -B -f org.eclipse.xtend.relocated.parent ${MVN_ARGS[@]} $@
```
cdietrich: am not sure if we want to keep

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TargetPlatformProject.xtend
```diff
@@ -85,7 +79,9 @@ class TargetPlatformProject extends ProjectDescriptor {
 				<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
 					<unit id="org.eclipse.xtext.sdk.feature.group" version="0.0.0"/>
 					�IF config.xtextVersion.isSnapshot�
-						<repository location="https://download.eclipse.org/modeling/tmf/xtext/updates/nightly/"/>
+���					/org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/wizard/WizardConfigurationTest.java
```
cdietrich: revert before merge

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/ParentProjectDescriptor.xtend
```diff
@@ -94,8 +94,11 @@ class ParentProjectDescriptor extends ProjectDescriptor {
 					repositories {
 						mavenCentral()
 						�IF config.xtextVersion.isSnapshot�
+���							maven {
+���								url = 'https://central.sonatype.com/repository/maven-snapshots'
+���							}
 							maven {
-								url = 'https://central.sonatype.com/repository/maven-snapshots'
+								url = 'https://ci.eclipse.org/xtext/job/xtext/job/cd-big-bump/lastSuccessfulBuild/artifact/build/maven-repository/'
```
cdietrich: revert before merge

org.eclipse.xtext.tests/src/org/eclipse/xtext/xtext/wizard/WizardConfigurationTest.java
```diff
@@ -208,7 +209,7 @@ public void featureProjectsHaveEclipseBuildProperties() {
 				(AbstractFile it) -> "build.properties".equals(it.getRelativePath())));
 	}
 
-	@Test
+	@Test @Ignore
```
cdietrich: revert before merge

cdietrich:
@szarnekow @LorenzoBettini @HannesWell 

rebased on changes by @HannesWell that are on main now
so this one should be reviewable now from my pov

HannesWell:
In order to simplify the review and later tracing in the history I extracted your commits to move Xtext plugins to Java-21 and to drop support for Java-17, sorted them in to commits and created a separate PR with them:
- https://github.com/eclipse-xtext/xtext/pull/3647

This also includes a few missing clean-ups in the tests I found during review.
Plus updates in `strict-jdk` profiles in the root `pom.xml` that are now possible.

I've made sure to keep the diff to this PRs branch minimal and there are only a handful of merge conflicts (mainly in the Manifests due to closely located updated bundle requirements).
I can create a PR against this branch merging the new main if https://github.com/eclipse-xtext/xtext/pull/3647 is accepted to not cause you extra work due to this extraction.

With that the number of changes in this PR should be significantly reduced where the majority are mechanical changes.

Jenkinsfile
```diff
@@ -27,7 +27,7 @@ pipeline {
   tools {
      // the Java version we use to run the build
      // we force the effective JDK version for compilation/testing through Maven toolchains
-     jdk "temurin-jdk21-latest"
+     jdk "temurin-jdk25-latest"
```
HannesWell: For GH the build runs, by default, with the latest supported Java version.
But since Tycho requires only Java-21, this could stay at Java-21 and back in the days was also at the minimally supported java version.

cdietrich:
@HannesWell do you see further split-off potential?

HannesWell:
> do you see further split-off potential?

Yes, I think the TP updates can be further split off. I have just created a PR with that:
- https://github.com/eclipse-xtext/xtext/pull/3650

Then only the changes for deployment changes and building the relocations should remain.
But I don't know yet, what exactly the goal is and if it's intended to be permanently on master?

Should that also be exported into a dedicated PR so this becomes 'empty' and can be closed (to have a dedicated PR for context)?

cdietrich:
we need to discuss the relocation stuff
i did this in order to be able to test in xtext-reference-project
as gralde and maven still ask for the old corrdinates

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TargetPlatformProject.xtend
```diff
@@ -66,11 +66,7 @@ class TargetPlatformProject extends ProjectDescriptor {
 					<unit id="org.eclipse.pde.feature.group" version="0.0.0"/>
 					<unit id="org.eclipse.draw2d.feature.group" version="0.0.0"/>
 					<unit id="org.eclipse.emf.sdk.feature.group" version="0.0.0"/>
-					�IF config.javaVersion.isAtLeast(JavaVersion.JAVA25)�
```
cdietrich, line 69: @HannesWell why did you have this if/else?
afaik all simrels are still on java 21

Jenkinsfile
```diff
@@ -182,8 +182,8 @@ def selectedTargetPlatform() {
         println("Choosing 'latest' target since this build was triggered by upstream with Java ${javaVersion}")
         return 'latest'
     } else if (isTriggeredByUpstream() && javaVersion>=21) {
-        println("Choosing 'r2025-12' target since this build was triggered by upstream with Java ${javaVersion}")
```
cdietrich: i wonder if this one is ever called.
this upstream stuff still stems from times where we had gradle and repo split

full-build.sh
cdietrich: @LorenzoBettini do ypou remember why we did not add it back then when we introduced the relocations

Jenkinsfile
HannesWell: I don't think so. Otherwise it would have file since the tp file name fragment is wrong for some time (I also didn't notice it).
My proposal is to remove that part as the pipeline can then be significantly simpler:
- https://github.com/eclipse-xtext/xtext/pull/3651

With that I also fix the TP selected for nightly builds, which I oversaw because in I removed the trigger for running the pipeline at the m2e infrastructure.

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TargetPlatformProject.xtend
HannesWell: Good question.
I believe I did this because support for Java-25 is better with 2026-03, although it works already with 2025-03.
Should I remove it and always use 2025-03?

HannesWell:
> we need to discuss the relocation stuff
> i did this in order to be able to test in xtext-reference-project
> as gralde and maven still ask for the old corrdinates

Understand. I've also extracted that into a dedicated PR so a dedicated discussion could happen there:
- https://github.com/eclipse-xtext/xtext/pull/3652

With that only the change of Maven-Central snapshot repositories remains as far as I can see. But from your comments I have the impression that it's a temporary change for testing and this PR is then effectively empty as everything has been extracted?

If that's correct, I suggest to close this and create dedicated follow-up PRs, e.g. if https://github.com/eclipse-xtext/xtext/pull/3638#discussion_r2931112375 requires a change.

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TargetPlatformProject.xtend
cdietrich: i believe we can do the rest with this pr. i will undo the stuff to fullbuild.sh for now and then later on check with lorenzo

full-build.sh
HannesWell: I suggest to discuss that in the dedicated PR at
- https://github.com/eclipse-xtext/xtext/pull/3652

org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TargetPlatformProject.xtend
cdietrich: so ok with the remaineder of this pr? if yes ill merge once build is in, and tomorrow morning update the refefrence projects one as well

HannesWell: Fine for me.
Thanks.

I'll have a look at mwe in the next days and will try to land my open PRs (I assume they need some adjustments now) and apply similar changes there too.

PR #3648, opened 2026-03-12
Fix expected Tycho version in mavenTychoP2J25.parent test

HannesWell:
This was forgotten in the Update to Tycho 5.0.2 in
- https://github.com/eclipse-xtext/xtext/pull/3602
 
because that PR was not rebased on latest main and in the meantime that new case was introduced with


HannesWell:
@cdietrich this should fix the current test failure on main.

PR #3649, opened 2026-03-12
remove download.itemis.com part 1

cdietrich:


HannesWell:
Will this also improve the stability of the build?
Currently I often see build failures because of a itemis p2-repository not being available.

cdietrich:
this is the hope. 
did not hear back from @kthoms on the patch file yet

LorenzoBettini:
> this is the hope. did not hear back from @kthoms on the patch file yet

The patch file could be downloaded from GitHub IIRC

PR #3651, opened 2026-03-13
[Build] Fix nightly build TP and drop unused upstream trigger handling

HannesWell:
Upstream builds that trigger the Xtext Jenkins pipelines do not exist anymore and therefore isTriggeredByUpstream() always returns false. With that assumption the pipline can be simplify and also selectedTargetPlatform() can be removed.
The latter returned an erroneous tp file name fragment anyways if it where called by an upstream job.

Follow-up on
- https://github.com/eclipse-xtext/xtext/pull/3650

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ pipeline {
 
   triggers {
     parameterizedCron(env.BRANCH_NAME == 'main' ? '''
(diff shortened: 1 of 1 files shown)


HannesWell:
This should also fix the failure on the main branch.

PR #3652, opened 2026-03-13
Build also the relocations

HannesWell:
This is extracted from:
- https://github.com/eclipse-xtext/xtext/pull/3638


HannesWell:
Obsolete as per https://github.com/eclipse-xtext/xtext/pull/3638#discussion_r2933088574.

PR #3655, opened 2026-03-13
Set default Java version to 21 in Xtext/Xtend Maven plugins

HannesWell:
Follow-up on
- https://github.com/eclipse-xtext/xtext/pull/3647

and also discussed in
- https://github.com/eclipse-xtext/xtext/pull/3603

### DIFF
--- a/org.eclipse.xtext.maven.plugin/src/test/resources/it/generate/xcore-auto-mapping/sample-auto.emf/model/a.xcore
+++ b/org.eclipse.xtext.maven.plugin/src/test/resources/it/generate/xcore-auto-mapping/sample-auto.emf/model/a.xcore
@@ -7,12 +7,12 @@
 	editPluginID="sample-auto.emf.edit", 
 	forceOverwrite="true", 
 	resource="XMI",
-	complianceLevel="11.0"
+	complianceLevel="21.0"
 )
 package axcore
 
 annotation "http://www.eclipse.org/emf/2002/GenModel" as GenModel
 
 class ModelA {
-	
+
 }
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.maven.plugin/src/test/resources/it/generate/xcore-auto-mapping/sample-auto.emf/model/a.xcore
```diff
@@ -7,12 +7,12 @@
 	editPluginID="sample-auto.emf.edit", 
 	forceOverwrite="true", 
 	resource="XMI",
-	complianceLevel="11.0"
```
cdietrich, line 10: we did not bump emf yet so i dont know if this will work

cdietrich: (is yet another big thing.
see https://github.com/eclipse-xtext/xtext/pull/3293 for last one)

HannesWell: I assumed that's related to the default as it's test data.

But I'll have a look later.

cdietrich: could work anyway as this is quite old https://github.com/eclipse-emf/org.eclipse.emf/commit/823d27818c87aebcddec36707925cccbb5ad0dda

PR #3662, opened 2026-03-17
Clean-up and format target files

HannesWell:
Leverage omitted version specification to get the latest version available in locations.
Use version ranges where versions are necessary, because not always the latest version is intended.

Remove unused `sequenceNumber` attribute.
Additionally format file content.

This is similar to
- https://github.com/eclipse-mwe/mwe/pull/362

### DIFF
--- a/org.eclipse.xtext.tests/testdata/wizard-expectations/full/full.parent/full.target/full.target.target
+++ b/org.eclipse.xtext.tests/testdata/wizard-expectations/full/full.parent/full.target/full.target.target
@@ -1,35 +1,35 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde version="3.8"?>
-<target name="full.target" sequenceNumber="1">
+<target name="full.target">
 	<targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-21"/>
 	<locations>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<unit id="org.eclipse.jdt.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.platform.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.pde.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.draw2d.feature.group" version="0.0.0"/>
-			<unit id="org.eclipse.emf.sdk.feature.group" version="0.0.0"/>
+			<unit id="org.eclipse.jdt.feature.group"/>
+			<unit id="org.eclipse.platform.feature.group"/>
+			<unit id="org.eclipse.pde.feature.group"/>
+			<unit id="org.eclipse.draw2d.feature.group"/>
+			<unit id="org.eclipse.emf.sdk.feature.group"/>
 			<repository location="https://download.eclipse.org/releases/2026-06"/>
 		</location>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<unit id="org.eclipse.emf.mwe2.launcher.feature.group" version="0.0.0"/>
+			<unit id="org.eclipse.emf.mwe2.launcher.feature.group"/>
 			<repository location="https://download.eclipse.org/modeling/emft/mwe/updates/releases/2.25.0/"/>
 		</location>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<unit id="org.eclipse.xtext.sdk.feature.group" version="0.0.0"/>
+			<unit id="org.eclipse.xtext.sdk.feature.group"/>
 			<repository location="https://download.eclipse.org/modeling/tmf/xtext/updates/milestones/"/>
 		</location>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-			<unit id="com.google.gson" version="2.13.2"/>
-			<unit id="com.google.inject" version="7.0.0"/>
-			<unit id="jakarta.inject.jakarta.inject-api" version="2.0.1"/>
-			<unit id="org.antlr.runtime" version="3.2.0.v20230929-1400"/>
-			<unit id="org.junit" version="0.0.0"/>
-			<unit id="org.hamcrest" version="2.2.0"/>
-			<unit id="org.hamcrest.core" version="2.2.0.v20230809-1000"/>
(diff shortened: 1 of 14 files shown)


org.eclipse.xtext.tests/testdata/wizard-expectations/full/full.parent/full.target/full.target.target
```diff
@@ -1,35 +1,35 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde version="3.8"?>
-<target name="full.target" sequenceNumber="1">
+<target name="full.target">
```
LorenzoBettini, line 3: I think it's fine to remove "0.0.0" and have version ranges (though, in that respect, I'd like to hear from @cdietrich because I don't know whether we have a strict version to keep it in sync with our BOM).

Concerning the removal of sequenceNumber: I thought that's used by PDE to understand whether the target platform should be reloaded, e.g., after updating from Git. Isn't that needed anymore?

cdietrich: in the past we had hickups sometimes ....
thus we initially went better safe than sorry

cdietrich: see my comment and eds answer on the mwe pr about sequence number

HannesWell: > I think it's fine to remove "0.0.0" and have version ranges (though, in that respect, I'd like to hear from @cdietrich because I don't know whether we have a strict version to keep it in sync with our BOM).

Since you use, at least for release, p2-repositories that are stable/do not change, the version range just selects from what's available in the specified repositories. In fact, actually the listed p2-repositories are the main definition of the version of artifacts.
And in that case, specifying a version range is only necessary if there are two bundles with the same version (e.g. a version 1.X and 2.Y) or if in the future there could be one and you'd get 'notified' by a resolution failure if the current version (range) is not available anymore.

> Concerning the removal of sequenceNumber: I thought that's used by PDE to understand whether the target platform should be reloaded, e.g., after updating from Git. Isn't that needed anymore?

Yes that's not used by PDE for a long time, not even read anymore for some time and not even updated in Oomph anymore since yesterday:
- https://github.com/eclipse-mwe/mwe/pull/362#discussion_r2944875204

LorenzoBettini: > Yes that's not used by PDE for a long time, not even read anymore for some time and not even updated in Oomph anymore since yesterday:

I wasn't taliking about Oomph: if in Eclipse I update from Git and the target platform has changed, I had the feeling that PDE would resolve the new one only if the sequenceNumber has changed. Alternatively, if I manually change the target file and reload it, I had the impression that PDE did not bother updating the target platform unless the sequenceNumber has changed.

Of course, if you tell me that PDE just ignores that sequenceNumber, we can safely remove it.

HannesWell: > Of course, if you tell me that PDE just ignores that sequenceNumber, we can safely remove it.

Yes PDE ignores that number and it can safely be removed.

LorenzoBettini: Then, that's OK.

But then the target file editor does not even increment that number anymore upon modifying and saving?

HannesWell: > But then the target file editor does not even increment that number anymore upon modifying and saving?

Yes, exactly.

HannesWell:
@cdietrich, @LorenzoBettini can this now be submitted?

LorenzoBettini:
There's a test failure on Jenkins https://ci.eclipse.org/xtext/job/xtext/job/PR-3662/6/testReport/junit/org.eclipse.xtext.ui.tests.editor.hover/ProblemHoverTest/testAnnotations/

Maybe it's not related, but I'll restart the build.

HannesWell:
The latest build has failures again, but I assume they're unrelated.

HannesWell:
@LorenzoBettini, @cdietrich can this be submitted now?

iloveeclipse:
Interesting. In our custom build (based on latest SDK nightly) we see now **two** versions of `org.antlr.runtime` in Xtext build
```
tmf-xtext-Update-2.43.0/plugins/org.antlr.runtime_3.2.0.v20230929-1400.jar
tmf-xtext-Update-2.43.0/plugins/org.antlr.runtime_3.5.3.v20230929-1400.jar
```
Last week build only had `3.2.0.v20230929-1400` version. 
So there is some dependency now that pulls in "latest"?

cdietrich:
cc @HannesWell

iloveeclipse:
Not sure whether the official Xtext build also has two versions, I don't know where are nightly artifacts from Xtext are hosted. Could be of course our internal issue with latest nightly platform we set to build Xtext.

cdietrich:
you can check at https://ci.eclipse.org/xtext/job/xtext/job/main/2877/artifact/build/p2-repository/

and this is what i feared with these "no versions"
osgi does tooo much magic

iloveeclipse:
> you can check at https://ci.eclipse.org/xtext/job/xtext/job/main/2877/artifact/build/p2-repository/
> 
> and this is what i feared with these "no versions" osgi does tooo much magic

Yep, same issue like with out build, two antlr versions:

image

LorenzoBettini:
Then the corresponding version constraint must be reset..

HannesWell:
> Then the corresponding version constraint must be reset..

If you only want version 3.2.0, then yes. My suggestion is to use suitable version ranges (in this case `[3.2, 3.3)`), in order to avoid the need to adjust the version with each update of a repo that has this changed.
If that's what you want, I can create a PR for it. If you want something else I can create a corresponding PR too.

Out of curiosity: Why are using requiring `antlr.runtime` 3.2 and not the latest 3.5?

> and this is what i feared with these "no versions"
> osgi does tooo much magic

Actually there is no magic at all. This can all be explained straight forward:

Both versions are contained in the target at the moment
grafik

Almost all Xtext plugin have their dependency on antlr.runtime locked to 3.2.0, but the antlr.generator has an open ended requirement on antlr.runtime:
grafik

And this pulls in the later version.

HannesWell:
> Out of curiosity: Why are using requiring `antlr.runtime` 3.2 and not the latest 3.5?

This question was also asked, but not really answered in
- https://github.com/eclipse-xtext/xtext/issues/3352

cdietrich:
we want 3.2 and there is obviously noone with time and knowledge to look into newer versions
cause this is also the version we use the generator of.

cdietrich:
@HannesWell @LorenzoBettini can one of you open a pr.
as i indicated last call i dont have time

PR #3698, opened 2026-05-03
Attempt to fix flaky ca test

cdietrich:


LorenzoBettini (approved):
So the problem is solved by all the following:

- explicit wait for PDE
- explicit dependency on PDE core
- explicit start level

right?

cdietrich:
i dont know if this really solve it. 
main was also quite stable the last two days.
i also dont know if all of it is needed or not

PR #3699, opened 2026-05-05
try to fix hover problem

cdietrich:


LorenzoBettini (approved):
Sounds reasonable.
Let's wait for the build.

PR #3724, opened 2026-06-02
feat: drop gwt support

cdietrich:
please check if i overlooked something


org.eclipse.xtext.dev-bom/pom.xml
```diff
@@ -155,11 +155,6 @@
 				<artifactId>error_prone_annotations</artifactId>
 				<version>2.23.0</version>
 			</dependency>
```
szarnekow, line 157: I wonder if we still need these two

szarnekow (approved):

org.eclipse.xtext.dev-bom/pom.xml
cdietrich: i assume they are in the general guava realm

PR #3729, opened 2026-06-05
Inline xtext.common.types.shared.jdt38 into xtext.xbase.ui

### LINKED ISSUE #3718: Inline org.eclipse.xtext.common.types.shared.jdt38 fragment?
The fragment `org.eclipse.xtext.common.types.shared.jdt38` currently contains only two classes of which the `OriginalEditorSelector.java` can be very useful when e.g. opening test results from the JUnit view for a test case that originates in a Xbase based DSL.

Can you tell why this fragment exists and why it's content is not merged into its host?
I see it has some optional UI dependencies, but so does the host.
Was it created because it uses some, back then new, JDT features?
And would it be reasonable and easily possible to merge into the host nowadays?

For background: I spend some time to find out that this fragment is the 'provider' of the functionality I mentioned above and maybe others could be saved from that too.

HannesWell:
Fixes https://github.com/eclipse-xtext/xtext/issues/3718

While implementing this I noticed that just inlining the fragment into its host `org.eclipse.xtext.common.types.shared` and adding the required (optional) dependencies leads to a dependency cycle with `org.eclipse.xtext.xbase.ui` and is therefore not feasible.
Since `OriginalEditorSelector` requires types from `org.eclipse.xtext.xbase.ui`, I found that `org.eclipse.xtext.xbase.ui` is probably the better target.
Placing it there does not require the addition of any new dependency in the target bundle and should not make a difference since `org.eclipse.xtext.xbase.ui` was always a prerequisite to make the `OriginalEditorSelector` work.
And since the `DebugPluginListener` is only referenced by `OriginalEditorSelector` I moved it into `org.eclipse.xtext.xbase.ui` too.

Additionally this fixes some warnings and avoids Guava in the moved classes.

@cdietrich and @szarnekow, please have a look.


cdietrich:
@iloveeclipse how do you solve this is your codebase? do you depend on the fragment + xbase.ui?

iloveeclipse:
> how do you solve this is your codebase? do you depend on the fragment + xbase.ui?

We ignore it :(

HannesWell:
Is there anything missing here or do you want me to adjust something?

I just resolved the conflicts and rebased this on the latest master commit.

cdietrich:
i cannot judge so leave to @szarnekow to find time to review

PR #3730, opened 2026-06-05
Avoid usage of org.antlr.runtime.Token.EOF_TOKEN

HannesWell:
instead check if a Token has the type EOF.

This simplifies a potential future migration to the latest ANTLR 3 release since the EOF_Token was removed.
Furthermore I avoids the assumption that the EOF token is always the same instance.
@szarnekow, you mentioned that the latter assumption lead to problems in the past.
Would it be possible to simplify some code with this change being applied.

This is extracted from 
- https://github.com/eclipse-xtext/xtext/pull/3669

### DIFF
--- a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/internal/Lexer.java
+++ b/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/internal/Lexer.java
@@ -9,6 +9,7 @@
 package org.eclipse.xtext.ide.editor.contentassist.antlr.internal;
 
 import org.antlr.runtime.CharStream;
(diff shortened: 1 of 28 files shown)


org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/contentassist/antlr/internal/Lexer.java
```diff
@@ -45,7 +46,7 @@ public Token nextToken() {
 			state.tokenStartLine = input.getLine();
 			state.text = null;
 			if ( input.LA(1)==CharStream.EOF ) {
-				return Token.EOF_TOKEN;
+				return new CommonToken(Token.EOF);
```
szarnekow, line 49: Would it make sense to create a CommonToken with correct offset information?

PR #3731, opened 2026-06-05
Simplify determination of Java version in JavaRuntimeVersion

HannesWell:


LorenzoBettini:
@cdietrich I'm not familiar with this feature.
@HannesWell, have you run any tests in that respect?

HannesWell:
> have you run any tests in that respect?

Not specifically. I just relied on the documentation stating that this returns the _feature_ version of the currently running JVM:
https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runtime.Version.html#FEATURE

which I assumed is the desired value at that place.

LorenzoBettini:
>  The sequence may be of arbitrary length but the first four elements are assigned specific meanings, as follows:

>     $FEATURE.$INTERIM.$UPDATE.$PATCH

So the result might be different from what the existing code returns, if I understand correctly.

```java
	@Test
	public void testDetermineJavaVersion() {
		assertEquals(8, JavaRuntimeVersion.determineJavaVersion("1.8.0_302"));
		assertEquals(11, JavaRuntimeVersion.determineJavaVersion("11.0.12"));
		assertEquals(13, JavaRuntimeVersion.determineJavaVersion("13.0.2"));
		assertEquals(14, JavaRuntimeVersion.determineJavaVersion("14"));
		assertEquals(15, JavaRuntimeVersion.determineJavaVersion("15.0.2"));
		assertEquals(16, JavaRuntimeVersion.determineJavaVersion("16.0.2"));
		assertEquals(17, JavaRuntimeVersion.determineJavaVersion("17-beta"));
		assertEquals(17, JavaRuntimeVersion.determineJavaVersion("17-ea"));
	}
```

we expect specific outputs just to get the major version, if I understand correctly.

HannesWell:
> So the result might be different from what the existing code returns, if I understand correctly.

Are you referring to the pre Java-9 cases, like `1.8.0_302` vs. `11.0.12`?
I assume that's obsolete since Xtext now requires Java-21 and `Runtime.version()` was only introduced in Java-10 (IIRC).
And that method is only used to determine the version of the executing Java runtime, isn't it?

LorenzoBettini:
@HannesWell I had missed the "feature" part.

I tried this

```java
package hello;

import java.lang.Runtime.Version;

public class HelloVersion {
	public static void main(String[] args) {
		Version version = Runtime.version();
		int feature = version.feature();
		System.out.println("Version: " + version);
		System.out.println("Feature: " + feature);
	}
}
```

And I get

```text
Version: 21.0.11+10
Feature: 21
```

So, we're fine.

LorenzoBettini (approved):

LorenzoBettini:
Thanks!

HannesWell:
> I had missed the "feature" part. 

Sorry, if this was not clear.

Thank you, too :)

PR #3732, opened 2026-06-05
Leverage new skipPublishing parameter of central-publishing-maven-plugin

### LINKED ISSUE #3544: Update central-publishing-maven-plugin 0.10.0
Which supports the configuration for `skipPublishing` in each module independently so that we can configure to honor a deployment skip property

HannesWell:
And update to to `central-publishing-maven-plugin` version 0.10.0 to use it.

Fixes https://github.com/eclipse-xtext/xtext/issues/3544

CC @LorenzoBettini


LorenzoBettini:
@HannesWell are you sure this is the right configuration? IIRC, in my other projects, the use of the skipping mechanism requires a bit more configuration. Setting a property outside the plugin configuration is not enough.

LorenzoBettini:
See for example https://github.com/LorenzoBettini/edelta-example-migration-library note the definition of the property in the parent, its use in the configuration of the plugin in the parent, and how the property is set in the poms of projects not to deploy.

HannesWell:
I see, but according to it's documentation `skipPublishing` parameter of the `central-publishing-maven-plugin` considers the  `skipPublishing` property by default already (similar to the `maven-deploy-plugin`, which considers `maven.deploy.skip`):

grafik

But if you prefer a more explicit configuration, I can update it as you prefer.
Or I can change it to read `maven.deploy.skip`. Then the nodeploy.parent doesn't have to change.

LorenzoBettini:
Ah! Then they further improved that. Back then, since I requested that feature, they told me to configure the plugin like I did in my other projects. That's fine. I'll review the PR on a computer, but I guess it should be fine.

LorenzoBettini:
@HannesWell I don't see what you say in the documentation of the plugin.
I seem to understand that you infer it from the hover dialog.
Can we trust that dialog?
Have you done any experiments concerning that?
Of course, I can merge this and see what it deploys as snapshot artifacts, but I was just curious.

HannesWell:
> I don't see what you say in the documentation of the plugin.

Yes the doc seems not to complete, but it's a real pity that Sonatype hasn't opensourced the plugin yet.
But in the source jar, e.g. at 
https://repo1.maven.org/maven2/org/sonatype/central/central-publishing-maven-plugin/0.10.0/central-publishing-maven-plugin-0.10.0-sources.jar

In the file `org.sonatype.central.publisher.plugin.AbstractPublisherMojo` one can find:
```
  @Parameter(property = "skipPublishing", defaultValue = "false", required = true)
  private boolean skipPublishing;
```
But I don't know when this was added.

> I seem to understand that you infer it from the hover dialog.
> Can we trust that dialog?

I think so, otherwise it would be a problem in m2e (or maybe even maven). 😅

> Have you done any experiments concerning that?

No I haven’t done any experiments with that.
 
> Of course, I can merge this and see what it deploys as snapshot artifacts, but I was just curious.

It would indeed be interesting, but I find it more consistent to read the `maven.deploy.skip` also for the skipPublisching, so I changed it accordingly. This also avoids the change in the `nodeploy.parent`.

LorenzoBettini (approved):

LorenzoBettini:
OK, thanks @HannesWell !
I merged this, let's keep an eye on what's getting published (in JIRO we can browse what's get locally published)

PR #3737, opened 2026-06-11
fix: repair template dialog sorting

cdietrich:
have no clue how to determine the default selection though


cdietrich:
cc @trancexpress @iloveeclipse

LorenzoBettini (commented):
Since the contents should now be sorted, why have you removed the corresponding test?

cdietrich:
Cause the test does not test the ui but t he w iterator the ui calls and then sorts

LorenzoBettini (approved):

PR #3753, opened 2026-07-08
fix: avoid RequestManager deadlock in LSP rename by using doSyncRead

### LINKED ISSUE #3735: Deadlock potential in LSP RequestManager (and it's usages in rename & quickfix implementations)
It's currently possible to induce a deadlock in the `RequestManager` by doing the following:

- Read Request **[R1]** is queued, then executed. `RequestManager.queue` is empty again and `RequestManager.parallel` is executing the `Runnable` of **[R1]**.
- Write Request **[W]** is queued, then executed. The implementation of `WriteRequest` awaits the completion of all current Requests (which were also marked as cancelled). During this time it blocks `RequestManager.queue`.
- **[R1]** now starts another Read Request **[R2]** and _waits for its completion_.
- **[R2]** is now queued, but has to wait for **[W]** to complete, which waits for **[R1]** to complete, which waits for **[R2]** to complete. Hence deadlock.

I think this behaviour would actually theoretically be acceptable, though I'd explicitly document, that you may not `join` the resulting futures unless you can guarantee that you're not already in a Request.

The issue is, that this exact situation can already happen with the way it's being used in the default XText implementations.

Specifically I found it for the `Rename` request and in the `ChangeConverter2`, which is used for quickfixes. Here is the code paths explained:

## ChangeConverter2

This path is probably the worse one (and the one we actually ran into with our quickfixes, since some of them take quite long to calculate.

The `ChangeConverter2` uses `ILanguageServerAccess#doRead` to access the document and then immediately awaits it:

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/ChangeConverter2.java#L80-L96

`ILanguageServerAccess#doRead` does start another Read Request:

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java#L1145-L1151

However, this code already potentially runs in a request. E.g. in the default diagnostics path:

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/editor/quickfix/DiagnosticResolution.java#L122-L144

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/codeActions/QuickFixCodeActionService.java#L80-L97

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java#L802-L828

## Rename Deadlock

The code path for a rename request looks as follows:

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java#L994-L1016

Here the requestManager is immediately called and a read request is queued. Inside of that read request, we execute the `RenameService2`:

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/rename/RenameService2.java#L89-L150

Here line 95 and 146 are the issue. They queue a new read request and then await it. So if a write request is queued in the (admitedly short) time frame of the start of the request until the rename service queues its read request, you'll get a dead lock.

(the same applies for PrepareRename)

## Suggested fix

`ILanguageServerAccess` already offers a `doSyncRead` method, which is identical to the `doRead` method, with the exception that it does NOT queue a new Read Request:

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java#L1192-L1196

vs

https://github.com/eclipse-xtext/xtext/blob/b6db004e1eff276d5ea2df982ff64623dfab033a/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/server/LanguageServerImpl.java#L1146-L1151

By using that method, we can now prevent a deadlock.

For the Rename path, this should be quite obviously fine (as it's guaranteed to run inside a read request already, so it doesn't interfere with the scheduling done via the `RequestManager`).

For the `ChangeConverter2` it's less obvious for me. In the quickfix path, it should be required, but I am not sure if there are other code paths where the asynchronous request is necessary.
I can't say I understand why this was ever done in an asynchronous manner in the first place, tbh. I haven't come up with a scenario that would require or desire that.
Maybe you still know the reasoning behind it, and whether it's fine to also change it to a synchronous method?

mvanhorn:
## Summary
LSP rename, prepareRename, and quickfix change conversion no longer queue a nested read request and block on it, which removes the `RequestManager` deadlock window: a nested read [R2] waits on a queued write [W], which awaits the in-flight outer read [R1], which waits on [R2]. The three call sites now use the existing `ILanguageServerAccess#doSyncRead`, which performs the same read without queueing.

## Why this matters
#3735 traces both in-tree paths precisely: `RenameService2.rename()` and `prepareRename()` call `doRead(uri, ...).get()` from inside the read request that `LanguageServerImpl` already queued, and `ChangeConverter2` does the same inside the code-action request. The reporter's company hits this regularly through slow quickfixes. The issue body itself proposes `doSyncRead` as the fix.

## Changes
- `RenameService2.rename()` restructures the `.exceptionally(...).get()` chain into a try/catch that preserves the existing contract: null on a root-cause `FileNotFoundException` when `shouldPrepareRename` is true, everything else rethrown via `Exceptions.sneakyThrow`.
- `ChangeConverter2` switches its `doRead(...)` + immediate join to `doSyncRead`. Every in-tree caller (the quickfix path through `DiagnosticResolution` and `QuickFixCodeActionService`) already runs inside a request and the method blocked on the future immediately, so the synchronous read is behavior-preserving apart from removing the deadlock window, answering the question the issue raised about whether the async form was ever required.

## Testing
New regression test in `Rename2Test` exercises rename while requests interleave; the existing rename, prepareRename, and code-action tests in `org.eclipse.xtext.ide.tests` stay green.

Fixes #3735

### DIFF
--- a/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/InterleavingRequestManager.java
+++ b/org.eclipse.xtext.ide.tests/src/org/eclipse/xtext/ide/tests/server/InterleavingRequestManager.java
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) 2026 TypeFox GmbH (http://www.typefox.io) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.xtext.ide.tests.server;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.eclipse.xtext.ide.server.concurrent.RequestManager;
+import org.eclipse.xtext.service.OperationCanceledManager;
+import org.eclipse.xtext.util.CancelIndicator;
+import org.eclipse.xtext.xbase.lib.Functions.Function1;
+
+import com.google.inject.Inject;
+
+public class InterleavingRequestManager extends RequestManager {
+	private final AtomicBoolean gateNextRead = new AtomicBoolean();
+	private final CountDownLatch readEntered = new CountDownLatch(1);
+	private final CountDownLatch continueRead = new CountDownLatch(1);
+
+	@Inject
+	public InterleavingRequestManager(ExecutorService parallel, OperationCanceledManager operationCanceledManager) {
+		super(parallel, operationCanceledManager);
+	}
+
+	public void gateNextRead() {
+		gateNextRead.set(true);
+	}
+
+	public boolean awaitReadEntered(long timeout, TimeUnit unit) throws InterruptedException {
+		return readEntered.await(timeout, unit);
+	}
+
+	public void continueRead() {
+		continueRead.countDown();
+	}
+
+	@Override
+	public synchronized <V> CompletableFuture<V> runRead(
+			Function1<? super CancelIndicator, ? extends V> cancellable) {
+		if (gateNextRead.compareAndSet(true, false)) {
+			return super.runRead((CancelIndicator cancelIndicator) -> {
+				readEntered.countDown();
+				try {
+					if (!continueRead.await(10, TimeUnit.SECONDS)) {
+						throw new AssertionError("Timed out waiting to continue read request.");
+					}
+				} catch (InterruptedException e) {
+					Thread.currentThread().interrupt();
+					throw new AssertionError(e);
+				}
+				return cancellable.apply(cancelIndicator);
+			});
+		}
+		return super.runRead(cancellable);
+	}
+}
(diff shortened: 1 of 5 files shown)


mvanhorn:
Pushed ed81a47d fixing the test failures: the deadlock regression test bound InterleavingRequestManager inside Rename2Test, which overrode the IRequestManager binding for every test in that class and broke the sibling rename tests. The harness and the regression test now live in a dedicated Rename2ConcurrencyTest with its own injector, so Rename2Test runs against the standard RequestManager again. Also working on getting the ECA registered for my commit email.

PR #3755, opened 2026-07-13
Fix NPE in getAdapter(URI) on non-platform resource URI

iloveeclipse:
If the URI is not a platform resource (in our case: Xtext document opened for a resource inside a jar file), getAdapter(URI.class) fails with NPE below:

```
java.lang.NullPointerException: Cannot invoke
"org.eclipse.emf.common.util.URI.isPlatformResource()" because
"resourceURI" is null
	at org.eclipse.xtext.ui.editor.model.IXtextDocument.getAdapter(IXtextDocument.java:46)
	at org.eclipse.xtext.ui.editor.model.XtextDocument.getAdapter(XtextDocument.java:669)
	at org.eclipse.xtext.ui.editor.model.XtextDocument.getAdapter(XtextDocument.java:653)
```

This patch adds trivial NP check and there will be no URI adapter for unsupported resources.


iloveeclipse:
@szarnekow : could we merge this? Shouldn't break anyone.

szarnekow (approved):

PR #3762, opened 2026-07-26
fix: bug in junit fragment if different root type is used

cdietrich:


org.eclipse.xtext.xtext.generator/src/org/eclipse/xtext/xtext/generator/junit/JUnitFragment.xtend
```diff
@@ -186,7 +186,7 @@ class JUnitFragment extends AbstractStubGeneratingFragment {
 				
 				@�test�
 				public void loadModel() throws Exception {
-					Model result = parseHelper.parse("Hello Xtext!");
+					�rootType� result = parseHelper.parse("Hello Xtext!");
```
LorenzoBettini, line 189: Just using "var" without type should be enough

cdietrich: Feel free to push to my pr or open yet another one. Won’t have pc access for some hours

PR #3767, opened 2026-07-31
org.eclipse.update.install.sources=false

LorenzoBettini:
org.eclipse.update.install.sources=false

To prevent source features from being installed in the IDE.

See also https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3448#issuecomment-4944608840

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -336,8 +336,7 @@
   </setupTask>
   <setupTask
       xsi:type="setup.p2:P2Task"
-      label="Tools"
-      profileProperties="">
+      label="Tools">
     <requirement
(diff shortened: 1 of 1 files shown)


LorenzoBettini:
@cdietrich I tested it locally and it works as intended.
IMPORTANT: it requires a a recent Oomph, or the profile property will not be honored.
For example, just rerunning from the current workspace might not work if the installed Oomph is not recent.
Starting from an updated Oomph instead works.

cdietrich:
how is the oomph of a running install updated?

LorenzoBettini:
> how is the oomph of a running install updated?

If the oomph in the installed eclipse is not updated, the option will have no effect. It is best to recreate the ide (the workspace can be the same).

cdietrich (approved):

PR #3801, opened 2026-09-15
Configure a Java 25 JRE.

merks:


merks:
@cdietrich 

I had intended to look at how to enable Java 25 for the builds.  The PR is failing the builds, but has no build impact so builds are kind of broken.  This failure is caused by not specifyin
```
[ERROR]   Missing requirement for filter properties ~= $0: org.eclipse.core.filesystem 1.12.0.v20260912-1339 requires 'osgi.ee; (&(osgi.ee=JavaSE)(version=25)), filter=(!(org.eclipse.equinox.p2.disable.require.capability.osgi.ee=true))' but it could not be found
[ERROR]   Cannot satisfy dependency: org.eclipse.platform.feature.group 4.42.0.v20260914-2300 depends on: org.eclipse.equinox.p2.iu; org.eclipse.core.filesystem [1.12.0.v20260912-1339,1.12.0.v20260912-1339]
```
The org.eclipse.core.filesystem  bundle really requires Java 25 now.

I expected to see this specified:

image

LorenzoBettini:
Shouldn't such a requirement be specified in our latest target platform file?

cdietrich:
i guess we then want to run latest always with 25, this is how we did it in the past.

cdietrich:
hmmm we actually do in jenkinsfile
H H(3-4) * * * %TARGET_PLATFORM=latest;JDK_VERSION=25

cdietrich:
but yes i had selected a java 25 for this variable already .

cdietrich:
@LorenzoBettini can you check the github workflows

merks:
I'm not sure how it's specified but the one build is running while most other builds fail and I've looked at way too many project builds today, each doing things in a slightly different way...

In any case, this PR has nothing to do with the build.  It simply configures a Java 25 JRE in the IDE as part of the setup so that one can use the latest target platform in the IDE and launch debug sessions.  So maybe best to merge it and deal with build issues separately...

LorenzoBettini:
As I said above (https://github.com/eclipse-xtext/xtext/pull/3801#issuecomment-5681618522), this line must be changed appropriately 

https://github.com/eclipse-xtext/xtext/blob/main/xtext-latest.target#L4

Tycho has become strict on these things, thus, simply using Java 25 is not enough, IIRC.

merks:
In any case, latest is using 4.42 Platform so needs to be Java 25:

image

Again a good reason to change the setup for that.

PR #3809, opened 2026-09-22
chore: Make codemining dependency non-optional to ease development

ArneDeutsch:
Make codemining dependency non-optional to ease tycho build setups. Optionality was there for backward compatibility with 9 year old eclipse oxygen which is probably not important any more (changed java versions make us incompatible anyway).


ArneDeutsch:
https://github.com/eclipse-xtext/xtext/issues/2527

tamasmiklossy:
Currently I had to add the `<plugin id="org.eclipse.xtext.ui.codemining" version="0.0.0"/>` entry to my feature.xml file to be able to use the Xtext codemining in the deployed application. I hope that this will not be necessary any more after removing the `optional` keyword.

Thanks @ArneDeutsch for your PR, LGTM, approved from my side!

